PR #256, opened 2018-04-18
Run job reconciliation framework on its own thread

tbak:
This change includes also performance improvements in the EntityHolder class.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/framework/reconciler/EntityHolder.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/framework/reconciler/EntityHolder.java
@@ -16,13 +16,12 @@
 
 package com.netflix.titus.common.framework.reconciler;
 
+import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


titus-common/src/main/java/com/netflix/titus/common/framework/reconciler/EntityHolder.java
```diff
@@ -76,34 +88,26 @@ public String getId() {
...
-                }
-        ).orElseGet(() -> Pair.of(this, Optional.empty()));
+        if (!childrenById.containsKey(id)) {
+            return Pair.of(this, Optional.empty());
+        }
+        Map<String, EntityHolder> newChildrenById = new HashMap<>(childrenById);
+        EntityHolder removedChild = newChildrenById.remove(id);
+
+        EntityHolder newRoot = new EntityHolder(this.id, this.entity, newChildrenById, this.attributes);
+        return Pair.of(newRoot, Optional.of(removedChild));
```
fabiokung: `Optional.ofNullable`

tbak: We check that the child exists, so the removedChild is always non-null.

PR #260, opened 2018-04-20
V3 engine performance improvements

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/gateway/V3GrpcTitusServiceGateway.java
```diff
@@ -233,7 +233,11 @@ public V3GrpcTitusServiceGateway(V3JobOperations jobOperations,
                     snapshot.add(SNAPSHOT_END_MARKER);
                     return snapshot;
                 }))
-                .doOnError(e -> logger.error("Unexpected error in job {} event stream", jobId, e));
+                .doOnError(e -> {
+                    if (!(e instanceof JobManagerException)) {
```
fabiokung: what `JobManagerExceptions` are expected here that we don't need to log?

tbak: Most of `JobManagerExceptions` are normal results (job/task not found, etc).
This error is not completely swallowed as it would re-surface in another component, that subscribes to this observable. I will change the implementation to bo more explicit about the intent.

PR #272, opened 2018-05-08
Log cloud simulator task status update latencies

tbak:


titus-testkit/src/main/java/com/netflix/titus/testkit/embedded/cloud/connector/remote/SimulatedRemoteMesosSchedulerDriver.java
```diff
@@ -133,7 +133,12 @@
                         .withDelay(100, 1_000, TimeUnit.MILLISECONDS)
                         .buildExponentialBackoff()
                 ).subscribe(
-                        taskStatus -> callbackHandler.statusUpdate(this, toTaskStatusUpdate(taskStatus)),
+                        taskStatus -> {
+                            long startTime = System.currentTimeMillis();
```
corindwyer: If this is not temporary code then I suggest you use `Clock` instead

PR #274, opened 2018-05-10
Federation: ensure fields required for cursors are always present

fabiokung:
When fields are being filtered in a query, ensure all that is required by the federation proxy is included, and take out what the original API query didn't specify.

### DIFF
--- a/titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobManagementService.java
+++ b/titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobManagementService.java
@@ -16,9 +16,12 @@
 
 package com.netflix.titus.federation.service;
 
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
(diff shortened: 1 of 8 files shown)


titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobManagementService.java
```diff
@@ -164,14 +174,25 @@ public Completable updateJobStatus(JobStatusUpdate request) {
                     .setPagination(emptyGrpcPagination(request.getPage()))
                     .build());
         }
+
+        Set<String> fieldsFilter = Collections.emptySet();
+        if (request.getFieldsCount() > 0) {
+            fieldsFilter = new HashSet<>(request.getFieldsList());
+            fieldsFilter.addAll(JOB_MINIMUM_FIELD_SET);
```
andrew-leung: Would missing this have been added at Gateway or Master previously or would it have failed validation?

fabiokung: master adds it too by default

PR #279, opened 2018-05-15
Rename the job attribute parameter properties

corindwyer:


titus-api/src/main/java/com/netflix/titus/api/jobmanager/JobParameterAttributes.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.api.jobmanager;
+
+/**
+ * Constant keys for Job attributes.
+ */
+public final class JobParameterAttributes {
```
tbak: Why another class?

corindwyer: I wanted to separate out the system generated things from the user supplied things

titus-api/src/main/java/com/netflix/titus/api/jobmanager/JobParameterAttributes.java
```diff
@@ -0,0 +1,31 @@
...
+    public static final String JOB_PARAMETER_ATTRIBUTES_ALLOW_CPU_BURSTING = "titusParameter.agent.allowCpuBursting";
```
fabiokung: I'd vote for `parameter.titus.*` so we run less risk of someone matching the `titus*` prefix without the dot. That also allows for higher level systems to define their own set of parameters, e.g.: `parameter.spinnaker.*` etc.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
```diff
@@ -140,14 +140,14 @@ public DefaultV3TaskInfoFactory(MasterConfiguration masterConfiguration,
         }
 
         // Configure agent job attributes
-        containerInfoBuilder.setAllowCpuBursting(Boolean.parseBoolean(attributes.get(JOB_ATTRIBUTES_ALLOW_CPU_BURSTING)));
-        containerInfoBuilder.setAllowNetworkBursting(Boolean.parseBoolean(attributes.get(JOB_ATTRIBUTES_ALLOW_NETWORK_BURSTING)));
-        containerInfoBuilder.setBatch(Boolean.parseBoolean(attributes.get(JOB_ATTRIBUTES_BATCH)));
+        containerInfoBuilder.setAllowCpuBursting(Boolean.parseBoolean(containerAttributes.get(JOB_PARAMETER_ATTRIBUTES_ALLOW_CPU_BURSTING)));
```
fabiokung: should we separate job level attributes from container level? I have a slight preference for different namespaces here, but it's not a strong opinions: `parameter.titus.container.*`.

corindwyer: Today, they are namespaced by the `agent` so I would think it would be `parameter.titus.agent` in your example

titus-api/src/main/java/com/netflix/titus/api/jobmanager/JobParameterAttributes.java
corindwyer: I am not opposed to this idea. I like `titus` being first so I can easily see that it is a titus concept but agree that it might be error prone since `titus.` is reserved.

sargun: Are there going to titusParameter.server / titusParameter.controlPlane?

corindwyer: There could be something used here to describe something to the control plane.

PR #286, opened 2018-05-18
Limit Max Environment Variable Size

tbak:


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobAssertions.java
```diff
@@ -30,12 +31,18 @@
...
+import com.netflix.titus.common.util.CollectionsExt;
 import com.netflix.titus.common.util.StringExt;
 
 /**
  */
 public class JobAssertions {
 
+    private static final Charset UTF_8 = Charset.forName("UTF-8");
+
+    public static final int MAX_ENVIRONMENT_SIZE_MB = 32;
```
sargun: MB? Megabyte?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobAssertions.java
```diff
@@ -56,6 +63,20 @@ public boolean isValidIamRole(String iamRole) {
         return !StringExt.safeTrim(iamRole).isEmpty();
     }
 
+    public boolean isEnvironmentNotTooLarge(Map<String, String> environment) {
```
sargun: Really, we should make a TODO to convert the environment variables to byte arrays, and introduce the following two requirements to them:

- No `.` in the key
- No NULL `\0` in the key nor value

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobAssertions.java
sargun: KB?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobAssertions.java
```diff
@@ -56,6 +63,20 @@ public boolean isValidIamRole(String iamRole) {
...
+        if(CollectionsExt.isNullOrEmpty(environment)) {
+            return true;
+        }
+
+        int totalSize = environment.entrySet().stream().mapToInt(entry -> {
+            int keySize = StringExt.isEmpty(entry.getKey()) ? 0 : entry.getKey().getBytes(UTF_8).length;
+            int valueSize = StringExt.isEmpty(entry.getValue()) ? 0 : entry.getValue().getBytes(UTF_8).length;
+            return keySize + valueSize + 2;
```
sargun: Might be worth adding a comment saying the 2 additional bytes are for the equal sign and the NUL terminator.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobAssertions.java
tbak: In the JIRA ticket, there is 64MB. We were talking about 32K yesterday, but I thought I misunderstood you.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobAssertions.java
tbak: We could verify that all characters map correctly to ASCII charset. On top of this, we could also check for `.` and `\0`.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobAssertions.java
tbak: Done

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobAssertions.java
sargun: right now, people use ".", which is bad, so we'll have to roam and ride.

corindwyer: Rename method/constant names to indicate this is for environment variables and not the environment.

tbak: Done

PR #288, opened 2018-05-21
Minor Improvements

corindwyer:
* Properly handle malformed json as a bad request
* Add passthrough attributes to agent

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
@@ -60,6 +60,7 @@ import static com.netflix.titus.common.util.Evaluators.applyNotNull;
 @Singleton
 public class DefaultV3TaskInfoFactory implements TaskInfoFactory<Protos.TaskInfo> {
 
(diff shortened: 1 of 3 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
```diff
@@ -154,12 +155,20 @@ public DefaultV3TaskInfoFactory(MasterConfiguration masterConfiguration,
...
+        });
+
         // Configure Environment Variables
-        Map<String, String> userProvidedEnv = container.getEnv().entrySet()
-                .stream()
-                .filter(e -> e.getValue() != null)
-                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
-        containerInfoBuilder.putAllUserProvidedEnv(userProvidedEnv);
+        container.getEnv().forEach((k, v) -> {
+            if (v != null) {
```
sargun: v != null? and v != ""?

corindwyer: This would change the current behavior so we would need to first see if anyone is depending on empty string env variables.

PR #289, opened 2018-05-25
let the federation proxy set titus.stack for jobs and tasks

fabiokung:
This makes it so jobs and tasks are always persisted with `titus.stack`
being the federated stack name. The caveat is that jobs created not
through the federation proxy (including legacy V2 API calls that don't
go through federation) will not have the attribute persisted with them.
With it, the persisted state should more accurately reflect what jobs
and tasks were created in a federated context, and what is their
federated stack name.

All jobs and tasks queried through a federation proxy continue to have
`titus.stack` automatically decorated into them, regardless if the
records from each cell have it or not.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/endpoint/common/CellDecorator.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/endpoint/common/CellDecorator.java
@@ -33,27 +33,25 @@ public class CellDecorator {
     }
 
     /**
-     * Adds {@link JobAttributes#JOB_ATTRIBUTES_CELL "titus.cell"} and {@link JobAttributes#JOB_ATTRIBUTES_STACK "titus.stack"}
-     * labels to a {@link TitusJobSpec V2 job spec}, replacing existing values if they are already present.
+     * Adds a {@link JobAttributes#JOB_ATTRIBUTES_CELL "titus.cell"} label to a {@link TitusJobSpec V2 job spec},
+     * replacing the existing value if it is already present.
      */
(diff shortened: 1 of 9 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/endpoint/common/CellDecorator.java
```diff
@@ -33,27 +33,25 @@ public CellDecorator(Supplier<String> cellNameSupplier) {
...
-     * Adds {@link JobAttributes#JOB_ATTRIBUTES_CELL "titus.cell"} and {@link JobAttributes#JOB_ATTRIBUTES_STACK "titus.stack"}
-     * attributes to a {@link JobDescriptor V3 job spec}, replacing existing values if they are already present.
+     * Adds a {@link JobAttributes#JOB_ATTRIBUTES_CELL "titus.cell"} attribute to a {@link JobDescriptor V3 job spec},
+     * replacing the existing value if it is already present.
      */
     public JobDescriptor ensureCellInfo(JobDescriptor jobDescriptor) {
         final String cellName = cellNameSupplier.get();
         return jobDescriptor.toBuilder()
                 .putAttributes(JobAttributes.JOB_ATTRIBUTES_CELL, cellName)
-                .putAttributes(JobAttributes.JOB_ATTRIBUTES_STACK, cellName)
```
andrew-leung: Do we want to check if STACK is present already or not? Seems like a Job coming directly to mainvpc may be missing a titus.stack field.

PR #292, opened 2018-05-31
 Extract common JobManagement GRPC client code + add cache implementation

tbak:


titus-common/src/main/java/com/netflix/titus/common/util/rx/RetryHandlerBuilder.java
```diff
@@ -118,6 +125,13 @@ public RetryHandlerBuilder but() {
                 .doOnNext(error -> onErrorHook.call(error))
                 .zipWith(Observable.range(0, retryCount + 1), RetryItem::new)
                 .flatMap(retryItem -> {
+                    if(retryWhenCondition != null && !retryWhenCondition.get()) {
+                        String errorMessage = String.format(
+                                "Retry condition not met for %s. Last error: %s. Returning an error to the caller",
+                                title, retryItem.cause.getMessage()
+                        );
+                        return Observable.error(new IOException(errorMessage, retryItem.cause));
```
fabiokung: minor: why `IOException`?

tbak: Retries are used mostly in connectors so `IOException` seemed reasonable.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/jobmanager/JobCache.java
```diff
@@ -0,0 +1,251 @@
...
+import com.netflix.titus.api.jobmanager.model.job.JobState;
+import com.netflix.titus.api.jobmanager.model.job.Task;
+import com.netflix.titus.api.jobmanager.model.job.TaskState;
+import com.netflix.titus.common.util.CollectionsExt;
+import com.netflix.titus.common.util.tuple.Pair;
+
+import static java.util.Collections.unmodifiableList;
+import static java.util.Collections.unmodifiableMap;
+
+public class JobCache {
```
corindwyer: As discussed. We should rename these types of classes to `Snapshot` since they are just an immutable snapshot of the state. Also, should start thinking about commonalities between the snapshots that should be followed either by setting a pattern or through code reuse:
* metrics
* data loading
* viewing the contents of a snapshot

tbak: I will rename JobCache to snapshot in another PR. Too many changes are accumulated here.
This is still work in progress, and I purposefully did not create common APIs that could be used between GRPC clients (job, agent). Building a common abstraction for `Snapshot` reusable in different places is a larger piece of work. We can try to standardize some aspects (naming, metrics), but I do not believe we can build something as generic as Guava cache.

PR #294, opened 2018-06-06
Better logging/exception handling + on abort callback

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/MesosSchedulerCallbackHandler.java
```diff
@@ -403,27 +403,31 @@ public void slaveLost(SchedulerDriver arg0, SlaveID slaveId) {
...
-            v2StatusUpdate(effectiveTaskStatus);
-        } else {
-            v3StatusUpdate(effectiveTaskStatus);
+            if (JobFunctions.isV2Task(taskId)) {
+                v2StatusUpdate(effectiveTaskStatus);
+            } else {
+                v3StatusUpdate(effectiveTaskStatus);
+            }
+        } catch (Exception e) {
+            logger.error("Unexpected error when handling the status update: {}", taskStatus, e);
```
fabiokung: should this also be re-propagated, as before?

tbak: As discussed offline, it is safer to re-throw it. I will change it accordingly.

PR #299, opened 2018-06-11
Cassandra truncate table resiliency improvements

tbak:


titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/tool/CassandraUtils.java
```diff
@@ -41,11 +42,39 @@
     public static final int SPLIT = 2;
 
     public static void truncateTable(CommandContext context, String table) {
+        for (int i = 0; i < 3; i++) {
```
corindwyer: move 3 to constant?

PR #300, opened 2018-06-11
Enforce Min Disk Size

corindwyer:
Enforce that the disk size is always greater than or equal to the min disk size configuration. If the request is changed then `diskSizeLessThanMin` is appended to the `titus.noncompliant` job attribute.

### DIFF
--- a/titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/JobManagerConfiguration.java
+++ b/titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/JobManagerConfiguration.java
@@ -31,4 +31,10 @@ public interface JobManagerConfiguration {
 
     @DefaultValue("_none_")
     String getNoncompliantClientWhiteList();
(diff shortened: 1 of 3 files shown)


titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/JobManagerConfiguration.java
```diff
@@ -31,4 +31,10 @@
 
     @DefaultValue("_none_")
     String getNoncompliantClientWhiteList();
+
+    /**
+     * @return the minimum disk size in megabytes that the disk resource dimension should be set to.
+     */
+    @DefaultValue("10000")
+    int getMinDiskSize();
```
tbak: getMinDiskSizeMB?

sargun:
megabytes or mibibytes?

corindwyer:
the same as the rest of the apis. MB

PR #302, opened 2018-06-12
[TITUS-2268] Add load balancer REST API

gabrielhartmann:
### Description of the Change

The Load Balancer API only had a gRPC interface.  This PR adds the accompanying REST API.

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/rest/LoadBalancerResource.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/rest/LoadBalancerResource.java
@@ -1,0 +1,70 @@
+package com.netflix.titus.runtime.endpoint.v3.rest;
+
+import com.netflix.titus.grpc.protogen.*;
+import com.netflix.titus.runtime.endpoint.common.rest.Responses;
+import com.netflix.titus.runtime.service.LoadBalancerService;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import javax.ws.rs.*;
+import javax.ws.rs.core.*;
+
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+@Api(tags = "Load Balancing")
+@Path("/v3/loadBalancers")
+@Singleton
+public class LoadBalancerResource {
+    private final LoadBalancerService loadBalancerService;
+
+    @Inject
+    public LoadBalancerResource(LoadBalancerService loadBalancerService) {
+        this.loadBalancerService = loadBalancerService;
+    }
+
+    @GET
+    @ApiOperation("Find the load balancer(s) with the specified ID")
+    @Path("/{jobId}")
+    public GetJobLoadBalancersResult getJobLoadBalancers(@PathParam("jobId") String jobId) {
+        return Responses.fromSingleValueObservable(loadBalancerService.getLoadBalancers(
+                JobId.newBuilder()
+                        .setId(jobId)
+                        .build()));
+    }
+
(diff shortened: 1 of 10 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/rest/LoadBalancerResource.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Inject
+    public LoadBalancerResource(LoadBalancerService loadBalancerService) {
+        this.loadBalancerService = loadBalancerService;
+    }
+
+    @GET
+    @ApiOperation("Find the load balancer with the specified ID")
+    @Path("/{jobId}")
+    public Response getJobLoadBalancers(@PathParam("jobId") String jobId) {
```
fabiokung: Should the return here be `GetLoadBalancersResult`? Is this not causing a `ClassCastException`?

corindwyer: Can there be multiple load balancers associated with a job? Also, this should be an error at runtime as fromSingleValueObservable would not return type `Response`.

titus-server-runtime/src/test/java/com/netflix/titus/runtime/endpoint/v3/rest/LoadBalancerResourceTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static com.netflix.titus.runtime.endpoint.v3.rest.RestConstants.PAGE_SIZE_QUERY_KEY;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+
+/**
+ * Tests the {@link LoadBalancerResource} class.
+ */
+public class LoadBalancerResourceTest {
```
corindwyer: Missing test for `getJobLoadBalancers`?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/rest/LoadBalancerResource.java
```diff
@@ -0,0 +1,70 @@
...
+public class LoadBalancerResource {
+    private final LoadBalancerService loadBalancerService;
+
+    @Inject
+    public LoadBalancerResource(LoadBalancerService loadBalancerService) {
+        this.loadBalancerService = loadBalancerService;
+    }
+
+    @GET
+    @ApiOperation("Find the load balancer with the specified ID")
```
corindwyer: balancer -> balancer(s)

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/rest/LoadBalancerResource.java
gabrielhartmann: Thanks @fabiokung,  as noted by @corindwyer below I forgot to write a test which would have caught that.  I've written the test and fixed the bug.

titus-server-runtime/src/test/java/com/netflix/titus/runtime/endpoint/v3/rest/LoadBalancerResourceTest.java
gabrielhartmann: Yes, I forgot to write the test.  Thanks for catching that.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/rest/LoadBalancerResource.java
gabrielhartmann: ACK

PR #304, opened 2018-06-14
federation: support for dynamic routing rules

fabiokung:
### Description of the Change

Recompute cell routing rules every time the configuration changes. This allows cell routing rules to be modified in configuration without operators having to bounce JVM processes.

Cells still can not be added or removed dynamically.


titus-common/src/main/java/com/netflix/titus/common/util/CollectionsExt.java
```diff
@@ -289,6 +289,14 @@ private CollectionsExt() {
         return result;
     }
 
+    public static <K, U, V, M extends Map<K, U>>
+    Map<K, V> mapValues(M source, Function<U, V> valueMapper, Supplier<Map<K, V>> mapSupplier) {
```
tbak: Why `M` not `Map<K, U>`?

fabiokung: you are right, fixing. I wanted it being able to return the concrete impl when needed (I was using `LinkedHashMap` in a previous version of this, but it is not needed anymore.

PR #306, opened 2018-06-15
System log service

tbak:
### Description of the Change

An API for logging critical system events (errors, exceptions, invariant violations)
with supplementary metadata information. The intent is to write those events to external
system(s) for further processing and automated analysis and reporting.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogService.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogService.java
@@ -1,0 +1,24 @@
+package com.netflix.titus.common.runtime;
+
+/**
+ * An API for logging critical system events (errors, exceptions, invariant violations) with supplementary metadata information.
+ * The events are classified by their category and priority for better searchability.
+ */
+public interface SystemLogService {
+
+    enum Priority {
+        Low,
+        Medium,
+        High
+    }
+
+    enum Category {
+        RecoverableError,
+        UnrecoverableError,
+        ConnectivityError,
+        InvariantViolation,
+        Other
+    }
+
+    boolean write(Category category, String priority, String component, String message);
+}
--- a/titus-common/src/main/java/com/netflix/titus/common/runtime/internal/LoggingSystemLogService.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/runtime/internal/LoggingSystemLogService.java
@@ -1,0 +1,22 @@
+package com.netflix.titus.common.runtime.internal;
+
+import com.netflix.titus.common.runtime.SystemLogService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoggingSystemLogService implements SystemLogService {
+
+    private static final Logger logger = LoggerFactory.getLogger(LoggingSystemLogService.class);
+
+    private static final LoggingSystemLogService INSTANCE = new LoggingSystemLogService();
+
+    @Override
+    public boolean write(Category category, String priority, String component, String message) {
+        logger.warn("[{}] [{}] {}: {}", category, component, priority, message);
+        return true;
+    }
+
+    public static LoggingSystemLogService getInstance() {
+        return INSTANCE;
+    }
+}
--- a/titus-common/src/main/java/com/netflix/titus/common/runtime/TitusRuntime.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/runtime/TitusRuntime.java
@@ -38,6 +38,11 @@ public interface TitusRuntime {
     CodeInvariants getCodeInvariants();
 
     /**
+     * Returns the configured {@link SystemLogService}.
+     */
+    SystemLogService getSystemLogService();
+
+    /**
      * Returns the configured Spectator registry.
      */
     Registry getRegistry();
--- a/titus-common/src/main/java/com/netflix/titus/common/runtime/TitusRuntimes.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/runtime/TitusRuntimes.java
@@ -19,6 +19,7 @@ package com.netflix.titus.common.runtime;
 import com.netflix.spectator.api.DefaultRegistry;
 import com.netflix.titus.common.runtime.internal.DefaultTitusRuntime;
 import com.netflix.titus.common.runtime.internal.LoggingSystemAbortListener;
+import com.netflix.titus.common.runtime.internal.LoggingSystemLogService;
 import com.netflix.titus.common.util.code.LoggingCodeInvariants;
 import com.netflix.titus.common.util.code.LoggingCodePointTracker;
 import com.netflix.titus.common.util.code.RecordingCodeInvariants;
@@ -34,6 +35,7 @@ public final class TitusRuntimes {
         return new DefaultTitusRuntime(
                 new LoggingCodePointTracker(),
                 LoggingCodeInvariants.getDefault(),
+                LoggingSystemLogService.getInstance(),
                 LoggingSystemAbortListener.getDefault(),
                 new DefaultRegistry(),
                 Clocks.system(),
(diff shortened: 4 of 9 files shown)


titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogService.java
```diff
@@ -0,0 +1,24 @@
...
+public interface SystemLogService {
+
+    enum Priority {
+        Low,
+        Medium,
+        High
+    }
+
+    enum Category {
+        RecoverableError,
```
gabrielhartmann: Some comments around your definitions of `recoverable` and `unrecoverable` would be helpful.  Do these refer to the ability to retry? `transient` vs `permanent`?

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogService.java
```diff
@@ -0,0 +1,24 @@
+package com.netflix.titus.common.runtime;
+
+/**
+ * An API for logging critical system events (errors, exceptions, invariant violations) with supplementary metadata information.
+ * The events are classified by their category and priority for better searchability.
+ */
+public interface SystemLogService {
+
+    enum Priority {
```
gabrielhartmann: What does priority mean here?  Is it different than the typical `INFO` `WARN` `ERROR` `FATAL` type levels?

tbak: I was thinking about something more actionable. For example:
* set `Critical` for a prolonged lack of connectivity to an external system
* set `Low` if the connection is dropped but we can restart quickly

However, maybe the typical levels are good enough and we should stick to them.

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogService.java
tbak: Yes. We could use those terms instead.

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogService.java
```diff
@@ -0,0 +1,24 @@
+package com.netflix.titus.common.runtime;
+
+/**
+ * An API for logging critical system events (errors, exceptions, invariant violations) with supplementary metadata information.
+ * The events are classified by their category and priority for better searchability.
+ */
+public interface SystemLogService {
```
corindwyer: This class needs to set expectations for this will be used including examples for each type and the guidelines for when you choose a different priority or level.

tbak:
@fabiokung I like this idea. We should probably have a set of predefined attributes, and an option to add custom ones. I feel we need to add the `SystemEvent` data type with a builder.

titus-common/src/main/java/com/netflix/titus/common/runtime/internal/LoggingSystemLogService.java
```diff
@@ -0,0 +1,34 @@
+package com.netflix.titus.common.runtime.internal;
+
+import com.netflix.titus.common.runtime.SystemLogEvent;
+import com.netflix.titus.common.runtime.SystemLogService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoggingSystemLogService implements SystemLogService {
```
amit-git: to keep up with existing naming convention, should it be called DefaultSystemLogService  ?

tbak: We use the term `Default` if there is no better qualifier. In this case it is specifically an implementation based on the logging library.

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
```diff
@@ -0,0 +1,243 @@
...
+ * System logger event (see {@link SystemLogService}).
+ */
+public class SystemLogEvent {
+
+    /**
+     * The criticality level of the log event.
+     */
+    public enum Priority {
+        /**
+         * Use when system operates normally, to log an information about an important system change. For example
```
corindwyer: remove `an`

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
```diff
@@ -0,0 +1,243 @@
...
+        Info,
+
+        /**
+         * Use when erroneous but non-critical things happens. For example, invariant violations that can be fixed, but
+         * indicate potential issues either in the code or the user provided data.
+         */
+        Warn,
+
+        /**
+         * Use when the system ability to make progress with its work is impaired or a major invariant violations are
```
corindwyer: system -> system's
violations are -> violation is

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
```diff
@@ -0,0 +1,243 @@
...
+         * discovered. For example, prolonged lack of connectivity with AWS service, corrupted data detected in a
+         * core system component, etc.
+         */
+        Error,
+
+        /**
+         * Use for major issues, which may result in an incorrect system behavior or forced system shutdown.
```
corindwyer: remove `an`

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
```diff
@@ -0,0 +1,243 @@
...
+     * to filter them out and detect common connectivity issues.
+     */
+    public enum Category {
+        /**
+         * A transient error.
+         */
+        Transient,
+
+        /**
+         * Permanent error. System is not able to recover from it without help from the administrator.
```
corindwyer: the -> an

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
```diff
@@ -0,0 +1,243 @@
...
+         */
+        ConnectivityError,
+
+        /**
+         * System invariant violations (for example corrupted data discovered in the core processing logic).
+         */
+        InvariantViolation,
+
+        /**
+         * Use when none of the existing categories is applicable.
```
corindwyer: is -> are

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogService.java
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+     * Write an event to an external system for further processing.
+     *
+     * @return true if the event was accepted, false otherwise. Accepting an event does not mean that it was or will be
+     * successfully delivered to an external event store. False result means that the event consumer is overloaded.
+     * {@link SystemLogService} implementation may choose in the latter case to accept higher priority events over
+     * the lower priority ones.
+     */
+    boolean write(SystemLogEvent event);
```
corindwyer: Maybe change to `submit` such that the async nature is more obvious?

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
tbak: Fixed

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
tbak: Fixed

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
tbak: Fixed

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
tbak: Fixed

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogEvent.java
tbak: Fixed

titus-common/src/main/java/com/netflix/titus/common/runtime/SystemLogService.java
tbak: Changed

PR #309, opened 2018-06-18
CodeInvariants-related improvements

dtrebbien:
### Description of the Change

Improvements to two classes in the `com.netflix.titus.common.util.code` package.


dtrebbien:
Here is the referenced usage of the unexpectedError(String, Object...) API:

https://github.com/Netflix/titus-control-plane/blob/6c76a486b7fa7f6b14d073d8ede8d89037e3dd3f/titus-server-master/src/main/java/com/netflix/titus/master/agent/service/monitor/AggregatingAgentStatusMonitor.java#L151

corindwyer:
Thanks for the contribution. Merging.

PR #311, opened 2018-06-27
update the min and max disk resource dimensions used by validators

corindwyer:
### Description of the Change

Update the min disk resource dimension to 10GB
Update the max disk resource dimension to 1TB


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobConfiguration.java
```diff
@@ -63,14 +63,14 @@
     @DefaultValue("524288")
```
sargun: You probably want to reduce this as well?

corindwyer: I can do this in another PR.

PR #312, opened 2018-06-29
allow V3 jobs to specify a custom CMD 

fabiokung:
### Description of the Change

Jobs that specify a custom `CMD` will have their entrypoint and command sent to agents not as flat
strings anymore. The new code path on agents will not force any shell parsing. It is responsibility of API callers to break commands and entrypoints into valid lists, respecting the semantics of [the docker API](https://docs.docker.com/engine/api/v1.24/#create-a-container), and [`execvp`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html).

The old behavior of flattening entrypoint strings, and relying on agent code to do shell parsing is being retired. To avoid breaking existing jobs, and allow for a smooth transition, only jobs that specify a command (previously a noop) will be sent to the new code path.

Related to Netflix/titus-executor#152 and Netflix/titus-api-definitions#52 (also TITUS-1574)

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
@@ -26,7 +26,6 @@ import javax.inject.Inject;
 import javax.inject.Singleton;
 
 import com.google.common.primitives.Ints;
-import com.netflix.archaius.api.Config;
 import com.netflix.fenzo.PreferentialNamedConsumableResourceSet;
 import com.netflix.fenzo.TaskRequest;
 import com.netflix.titus.api.jobmanager.model.job.BatchJobTask;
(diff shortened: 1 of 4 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
```diff
@@ -241,7 +249,7 @@ public DefaultV3TaskInfoFactory(MasterConfiguration masterConfiguration,
         // set scalars other than cpus, mem, disk
         final Map<String, Double> scalars = fenzoTask.getScalarRequests();
         if (scalars != null && !scalars.isEmpty()) {
-            for (Map.Entry<String, Double> entry : scalars.entrySet()) {
+            for (Map.Entry<String, Double> entry: scalars.entrySet()) {
```
corindwyer: you removed the space

PR #314, opened 2018-07-03
mark jobs with spaces in their entrypoint binaries as noncompliant

fabiokung:
These jobs are likely relying on legacy shell parsing being done by the executor, which is going away.

Internal: TITUS-1574

### DIFF
--- a/titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/ExtendedJobSanitizer.java
+++ b/titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/ExtendedJobSanitizer.java
@@ -1,15 +1,19 @@
 package com.netflix.titus.gateway.service.v3.internal;
 
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.inject.Named;
 import javax.validation.ConstraintViolation;
 
 import com.google.common.base.CharMatcher;
 import com.netflix.titus.api.jobmanager.model.job.ContainerResources;
+import com.netflix.titus.api.jobmanager.model.job.JobDescriptor;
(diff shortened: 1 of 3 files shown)


titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/ExtendedJobSanitizer.java
```diff
@@ -93,6 +99,24 @@ private boolean isInNonCompliantWhiteList(com.netflix.titus.api.jobmanager.model
         return markNonCompliant(sanitizedJobDescriptor, nonCompliant);
     }
 
+    /**
+     * Jobs with entry point binaries containing spaces are likely relying on the legacy shell parsing being done by
```
sargun: Would this break on `/bin/bash -c "cat | grep grep 'Wed May 16 20:47:08 PDT 2018'"`

sargun: Or not break, but consider it "invalid?

fabiokung: It depends how that is mapped onto `List<String> entryPoint` and `List<String> command`.

Valid:
```
entryPoint = ["/bin/bash", "-c"]
command = ["cat | grep grep 'Wed May 16 20:47:08 PDT 2018'"]
```

noncompliant:
```
entryPoint = ["/bin/bash -c "cat | grep grep 'Wed May 16 20:47:08 PDT 2018'"]
```

sargun: This may be a corner case, but `["/apps/foo\ bar"]`

fabiokung: I mention that in the comment, that spaces is _likely_ a problem, but not necessarily. The point of marking these jobs as `titus.noncompliant` is that we can reach out and help people fix the bad cases before we remove the old code path

PR #316, opened 2018-07-10
Fix the image digest validation pattern

tbak:
### Description of the Change

Implement validation rules according to spec (https://github.com/docker/distribution/blob/master/reference/reference.go)


tbak:
@fabiokung Link to reference is already there: https://github.com/Netflix/titus-control-plane/pull/316/files#diff-22f6e5a5aa6499b41ed25b522933e531R50

PR #317, opened 2018-07-17
Reduce cardinality of job/task metrics

tbak:
### Description of the Change

High cardinality, and using job and task ids as tag values causes memory leak, and high heap pressure.


titus-server-master/src/main/java/com/netflix/titus/master/mesos/MesosStateTracker.java
```diff
@@ -40,158 +35,37 @@
...
-    private final Clock clock;
     private final Registry registry;
 
     MesosStateTracker(MasterConfiguration configuration, TitusRuntime titusRuntime) {
-        this.configuration = configuration;
-        this.clock = titusRuntime.getClock();
         this.registry = titusRuntime.getRegistry();
-        PolledMeter.using(registry).withId(registry.createId(TRACKED_TASKS_METRIC_ID)).monitorSize(taskTrackers);
+
+        this.knownTasks = CacheBuilder.newBuilder()
```
corindwyer: FYI, there is a `InstrumentedCache` class that instruments a cache already. You might not care about the other metrics though.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/TaskLivenessMetrics.java
```diff
@@ -70,8 +66,9 @@
 
     private static final Logger logger = LoggerFactory.getLogger(TaskLivenessMetrics.class);
 
-    private static final String ROOT_METRIC_NAME = MetricConstants.METRIC_ROOT + "jobManager.taskLiveness.";
-    private static final String TASK_METRIC_NAME = ROOT_METRIC_NAME + "duration";
+    private static final String TASK_IN_STATE_ROOT_METRIC_NAME = MetricConstants.METRIC_ROOT + "jobManager.taskLiveness.";
+    private static final String TASK_IN_STATE_METRIC_NAME = TASK_IN_STATE_ROOT_METRIC_NAME + "duration";
```
corindwyer: Why is this named `duration`? I don't see where it is measuring a duration.

PR #325, opened 2018-07-25
Add more logging in the replicating connectors

tbak:
### Description of the Change

Additional logging in the stream replication connectors, that should help diagnose
reoccurring stability issues.

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/agent/replicator/GrpcAgentReplicatorEventStream.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/agent/replicator/GrpcAgentReplicatorEventStream.java
@@ -43,7 +43,10 @@ public class GrpcAgentReplicatorEventStream implements ReplicatorEventStream<Age
     @Override
     public Observable<ReplicatorEvent<AgentSnapshot>> connect() {
         return Observable.fromCallable(CacheUpdater::new)
-                .flatMap(cacheUpdater -> client.observeAgents().flatMap(cacheUpdater::onEvent))
+                .flatMap(cacheUpdater -> {
+                    logger.info("Connecting to the agent event stream...");
+                    return client.observeAgents().flatMap(cacheUpdater::onEvent);
+                })
                 .compose(ObservableExt.reemiter(
                         // If there are no events in the stream, we will periodically emit the last cache instance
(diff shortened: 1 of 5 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/agent/replicator/GrpcAgentReplicatorEventStream.java
```diff
@@ -56,7 +59,10 @@ public GrpcAgentReplicatorEventStream(AgentManagementClient client,
                     metrics.event(titusRuntime.getClock().wallTime() - event.getLastUpdateTime());
                 })
                 .doOnUnsubscribe(metrics::disconnected)
-                .doOnError(metrics::disconnected)
+                .doOnError(error -> {
+                    logger.warn("Connection to the agent event stream terminated with an error: {}", error.getMessage(), error);
```
fabiokung: nit: `logger.error`?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/eviction/replicator/GrpcEvictionReplicatorEventStream.java
```diff
@@ -53,7 +53,10 @@ public GrpcEvictionReplicatorEventStream(EvictionServiceClient client,
     @Override
     public Observable<ReplicatorEvent<EvictionDataSnapshot>> connect() {
```
fabiokung: all these `connect` methods seem very similar. Consider factoring them out?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/agent/replicator/GrpcAgentReplicatorEventStream.java
tbak: Connection failure is most of the time non-critical issue (could be master failover, gateway termination, etc).

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/eviction/replicator/GrpcEvictionReplicatorEventStream.java
tbak: +1
I think it is the right time to do it.

PR #337, opened 2018-08-04
removing v2 references

amit-git:
Deprecating V2 engine integration (also RxEventBus) code

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/appscale/service/DefaultAppScaleManager.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/appscale/service/DefaultAppScaleManager.java
@@ -346,7 +346,7 @@ public class DefaultAppScaleManager implements AppScaleManager {
     }
 
     private boolean isJobActive(String jobId) {
-        return v3JobOperations.getJob(jobId) != null && v3JobOperations.getJob(jobId).isPresent();
+        return v3JobOperations.getJob(jobId).isPresent();
     }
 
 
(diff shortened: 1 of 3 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/appscale/service/DefaultAppScaleManager.java
```diff
@@ -404,12 +346,7 @@ private boolean shouldRefreshScalableTargetForJob(String jobId, JobScalingConstr
     }
 
     private boolean isJobActive(String jobId) {
-        if (JobFunctions.isV2JobId(jobId)) {
-            return v2JobOperations.getJobMgr(jobId) != null && v2JobOperations.getJobMgr(jobId).isActive();
-        } else {
-            // V3
-            return v3JobOperations.getJob(jobId) != null && v3JobOperations.getJob(jobId).isPresent();
-        }
+        return v3JobOperations.getJob(jobId) != null && v3JobOperations.getJob(jobId).isPresent();
```
corindwyer: The contract is to return an `Optional` so you should not need to do a null check.

titus-server-master/src/main/java/com/netflix/titus/master/appscale/service/DefaultAppScaleManager.java
```diff
@@ -488,14 +402,6 @@ private JobScalingConstraints getJobScalingConstraints(String policyRefId, Strin
...
     private String buildAutoScalingGroup(String jobId) {
-        if (JobFunctions.isV2JobId(jobId)) {
-            if (v2JobOperations == null) {
-                return jobId;
-            }
-            V2JobDefinition jobDefinition = v2JobOperations.getJobMgr(jobId).getJobDefinition();
-            return buildAutoScalingGroupV2(jobDefinition.getParameters());
-        }
-
         if (v3JobOperations == null) {
```
corindwyer: `v3JobOperations` is injected and therefore should never be null

PR #341, opened 2018-08-18
Read Tier from AWS ASG tags

sargun:
If the AWS ASG has a special tag on it, and we recognize that as a tier, use that
when we initially see the the ASG, rather than default into flex tier.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/agent/service/cache/DataConverters.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/agent/service/cache/DataConverters.java
@@ -17,6 +17,7 @@
 package com.netflix.titus.master.agent.service.cache;
 
 import java.util.Collections;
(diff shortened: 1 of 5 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/agent/service/cache/DataConverters.java
```diff
@@ -33,11 +34,27 @@
 import com.netflix.titus.common.util.Evaluators;
 
 class DataConverters {
+    private static final String TIER_TAG = "titus:tier";
+
+    private static Tier tagsToTier(Map<String, String> tags, Tier defaultTier) {
+        final String tierTag = tags.get(TIER_TAG);
+        if (tierTag == null) {
```
fabiokung: `StringExt.isEmpty()`

PR #342, opened 2018-08-21
Bug fix for the leader election process OOS status update

tbak:
### Description of the Change

`leaveIfNotLeader` method implementation had a bug which would keep `OutOfService` instance as part of the leader election process.


titus-ext/zookeeper/src/main/java/com/netflix/titus/ext/zookeeper/supervisor/ZookeeperLeaderElector.java
```diff
@@ -106,8 +106,17 @@ public boolean join() {
 
     @Override
     public boolean leaveIfNotLeader() {
-        LeaderElectionProcess process = leaderElectionProcessRef.getAndSet(null);
-        return process != null && process.leaveIfNotLeader();
+        synchronized (leaderElectionProcessRef) {
```
fabiokung: Do we need this change? The previous version seemed good to me. Worst case when `join()` races with `leaveIfNotLeader()` can be treated as if `leaveIfNotLeader()` was called before the instance joined the leader election process.

titus-ext/zookeeper/src/main/java/com/netflix/titus/ext/zookeeper/supervisor/ZookeeperLeaderElector.java
```diff
@@ -143,16 +153,23 @@ private LeaderElectionProcess() {
...
         private boolean leaveIfNotLeader() {
-            if (!leaderFlag) {
+            if (leaderFlag) {
                 return false;
             }
 
             // Not a leader yet, so it is safe to leave the leader election process.
             close();
+
+            if (leaderActivator.isLeader()) {
```
fabiokung: isn't this still racy? The leader activation can still happen right after this check, and we'd still return true. Shouldn't this be fully serialized with `announceLeader()` instead?

I see that you added two checks on `terminateIfClosed()` to `announceLeader()`, but wouldn't it be better to just prevent activation completely?

titus-ext/zookeeper/src/main/java/com/netflix/titus/ext/zookeeper/supervisor/ZookeeperLeaderElector.java
tbak: The previous version had a bug, since it did `getAndSet(null)` even if the node was a leader. When I was fixing that bug, I chose more defensive implementation with synchronized.

titus-ext/zookeeper/src/main/java/com/netflix/titus/ext/zookeeper/supervisor/ZookeeperLeaderElector.java
tbak: Good point. I will add `synchronized` statement within the leader election callback.

titus-ext/zookeeper/src/main/java/com/netflix/titus/ext/zookeeper/supervisor/ZookeeperLeaderElector.java
fabiokung: Discussed in person. This is handling the case when the instance is already the leader, and we don't want to clean up the `leaderElectionProcessRef` in that case.

titus-ext/zookeeper/src/main/java/com/netflix/titus/ext/zookeeper/supervisor/ZookeeperLeaderElector.java
```diff
@@ -153,12 +153,14 @@ private LeaderElectionProcess() {
         }
 
         private boolean leaveIfNotLeader() {
-            if (leaderFlag) {
-                return false;
-            }
+            synchronized (this) {
```
corindwyer: minor nitpick. I prefer synchronizing on a mutex object instead of `this`

PR #349, opened 2018-08-27
es publisher wrapped in retry w/ exponential backoff

amit-git:
### Description of the Change

A full v3TasksStream needs to be wrapped with titusRuntime.persistentStream. This is to ensure any errors downstream won't terminate it.

### DIFF
--- a/titus-ext/elasticsearch/src/main/java/com/netflix/titus/ext/elasticsearch/ElasticsearchTaskDocumentPublisher.java
+++ b/titus-ext/elasticsearch/src/main/java/com/netflix/titus/ext/elasticsearch/ElasticsearchTaskDocumentPublisher.java
@@ -68,7 +68,7 @@ public class ElasticsearchTaskDocumentPublisher {
     private final Client client;
     private final Map<String, String> taskDocumentContext;
     private final TitusRuntime titusRuntime;
-    private Registry registry;
+    private final Registry registry;
     private final ObjectMapper objectMapper;
     private final SimpleDateFormat indexDateFormat;
     private final SimpleDateFormat taskDateFormat;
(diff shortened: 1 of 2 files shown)


corindwyer:
Discussed offline. Please add the other things we talked about such as metrics and error handling on the publish method.

titus-ext/elasticsearch/src/main/java/com/netflix/titus/ext/elasticsearch/ElasticsearchTaskDocumentPublisher.java
```diff
@@ -66,13 +70,20 @@
     private final ObjectMapper objectMapper;
     private final SimpleDateFormat indexDateFormat;
     private final SimpleDateFormat taskDateFormat;
+    private final AtomicInteger docsToBePublished = new AtomicInteger(0);
+    private final AtomicInteger docsPublished = new AtomicInteger(0);
+    private final AtomicInteger errorJsonConversion = new AtomicInteger(0);
+    private final AtomicInteger errorEsClient = new AtomicInteger(0);
+    private final AtomicInteger errorInPublishing = new AtomicInteger(0);
+    private final AtomicInteger isAlive = new AtomicInteger(0);
```
corindwyer: I suggest you use the `timeSinceLastPublish` strategy to where you are publishing a metric that denotes how long it has been since the last publish such that alerting on it will be easier.

titus-ext/elasticsearch/src/main/java/com/netflix/titus/ext/elasticsearch/ElasticsearchTaskDocumentPublisher.java
```diff
@@ -67,6 +68,7 @@
     private final Client client;
     private final Map<String, String> taskDocumentContext;
     private final TitusRuntime titusRuntime;
+    private Registry registry;
```
corindwyer: add `final`

PR #356, opened 2018-09-10
send fully qualified image to executor

corindwyer:
### Description of the Change

* Send fully qualified image to the executor based on the configured registry url

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
@@ -25,6 +25,7 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import com.google.common.base.Strings;
 import com.google.common.primitives.Ints;
 import com.netflix.fenzo.PreferentialNamedConsumableResourceSet;
 import com.netflix.fenzo.TaskRequest;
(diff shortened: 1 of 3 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/MesosConfiguration.java
```diff
@@ -92,4 +86,10 @@
      */
     @DefaultValue("./run")
     String getExecutorUriOverrideCommand();
+
+    /**
+     * @return the registry URL that will be merged with the image name.
+     */
+    @DefaultValue("")
+    String getRegistryUrl();
```
andrew-leung: Is this expected to also include the port suffix on the URL?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
```diff
@@ -206,6 +202,18 @@ public DefaultV3TaskInfoFactory(MasterConfiguration masterConfiguration,
         return containerInfoBuilder;
     }
 
+    private void setImage(ContainerInfo.Builder containerInfoBuilder, Image image) {
+        containerInfoBuilder.setImageName(image.getName());
+        String registryUrl = mesosConfiguration.getRegistryUrl();
+        if (!Strings.isNullOrEmpty(registryUrl)) {
```
andrew-leung: Are we supporting missing registry URLs only while we roman ride agent builds or are we expecting a long-term need for this?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/MesosConfiguration.java
corindwyer: yes, it would be the entire string that is prepended to the image name

titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
corindwyer: This logic has to exist in order to support backwards compatibility on the executor. If the contract to the executor changes, we can update this logic accordingly.

PR #357, opened 2018-09-10
Update critical tier capacity group behavior

corindwyer:
### Description of the Change

* Remove buffer from each capacity group in order to have a global shared buffer.
* Use only the current instance count instead of max for the critical tier within an instance group to determine available capacity.


titus-server-master/src/main/java/com/netflix/titus/master/service/management/internal/ResourceConsumptionEvaluator.java
```diff
@@ -115,8 +115,7 @@ private CompositeResourceConsumption buildSystemConsumption(Map<String, Map<Stri
         capacityGroupConsumptionMap.forEach((capacityGroup, appConsumptions) -> {
 
             ApplicationSLA sla = applicationSlaMap.get(capacityGroup);
-            double buffer = getBuffer(sla.getTier());
```
fabiokung: `getBuffer()` can probably be deleted now

corindwyer: +1, not going to change it now as this component will be getting an overhaul soon

PR #361, opened 2018-09-18
Add job validator to gateway

gabrielhartmann:
There are types of validation for Jobs which extend beyond the currently
available syntactic sanitization/validation approach.  Here we provide a
method for providing arbitrary validation of Jobs upon creation at the
gateway layer.  If a job fails validation the Titus master component
will never attempt to launch the job and clients will be notified of
failure.


titus-common/src/main/java/com/netflix/titus/common/model/validator/Validator.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.Set;
+
+/**
+ * A Validator determines whether an object of the parameterized type is valid.  If it finds an object to be invalid it
+ * returns a non-empty set of {@link ValidationError}s.
+ *
+ * @param <T> The type of object this Validator validates.
+ */
+public interface Validator<T> {
+    Set<ValidationError> validate(T entity);
```
tbak: This should be asynchronous API. There are two reasons for that:
* GRPC calls cannot block, as doing this would block the netty thread
* they cannot be easily parallelized/composed to speedup the execution

The `Validator` name is very generic, and already used by other framework. Can we be more specific here?

Looking at this API proposal, I see many commonalities with the existing `EntitySanitizer` framework. Does it make sense, to base both APIs on the same data model? `EntitySanitizer` uses now JavaBean model (`ConstraintViolation`), but we could use `ValidationError` provided here in both places.

titus-server-master/src/test/java/com/netflix/titus/master/integration/v3/job/JobValidatorNegativeTest.java
```diff
@@ -0,0 +1,73 @@
...
+    @Before
+    public void setUp() throws Exception {
+        instanceGroupsScenarioBuilder.synchronizeWithCloud().template(InstanceGroupScenarioTemplates.basicCloudActivation());
+        this.client = titusStackResource.getGateway().getV3BlockingGrpcClient();
+    }
+
+    @Test(timeout = 30_000)
+    public void testFailedValidationThrowsException() {
+        final com.netflix.titus.grpc.protogen.JobDescriptor jobDescriptor =
+                toGrpcJobDescriptor(batchJobDescriptors().getValue()).toBuilder().build();
```
tbak: `toBuilder().build()` is not needed here

titus-server-runtime/src/main/java/com/netflix/titus/runtime/TitusValidatorModule.java
```diff
@@ -0,0 +1,25 @@
...
+public class TitusValidatorModule extends AbstractModule {
+
+    @Override
+    protected void configure() {
+    }
+
+    @Provides
+    @Singleton
+    public Validator getJobValidator() {
+        return new PassJobValidator();
```
tbak: You can bind this directly in the `configure` method.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/jobmanager/client/GrpcJobManagementClient.java
```diff
@@ -77,6 +82,15 @@ public GrpcJobManagementClient(JobManagementServiceGrpc.JobManagementServiceStub
             return Observable.error(TitusServiceException.invalidArgument(violations));
         }
 
+        /**
+         * The validation call here is synchronous.  Any latency requirements for either the success or failure of
+         * validation should be implemented within the {@link Validator}.
+         */
+        Set<ValidationError> validationErrors = validator.validate(sanitizedCoreJobDescriptor);
```
corindwyer: By blocking here, the java thread is blocked which will block the grpc netty thread. We have used rxjava primitives to compose these pipeline style operations together without blocking the java thread.

titus-common/src/main/java/com/netflix/titus/common/model/validator/EntityValidator.java
```diff
@@ -9,5 +11,5 @@
  * @param <T> The type of object this EntityValidator validates.
  */
 public interface EntityValidator<T> {
-    Set<ValidationError> validate(T entity);
+    Flux<Set<ValidationError>> validate(T entity);
```
tbak: Contract description is missing. The expectation here is that always exactly one value is emitted. If there are no errors, this will be an empty set.

gabrielhartmann: Changed type to `Mono`

PR #366, opened 2018-09-24
Add validator to evaluate validators in parallel

gabrielhartmann:
The ParallelValidator parallelizes execution of soft/hard validators.
Failure to return within the indicated timeout generates soft/hard
ValidationErrors respectively.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
@@ -1,0 +1,80 @@
+package com.netflix.titus.api.jobmanager.model.job.validator;
+
+import java.time.Duration;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.netflix.titus.api.jobmanager.model.job.JobDescriptor;
+import com.netflix.titus.common.model.validator.EntityValidator;
+import com.netflix.titus.common.model.validator.ValidationError;
+import com.netflix.titus.common.util.tuple.Pair;
+import reactor.core.publisher.Mono;
+import reactor.util.function.Tuples;
+
+/**
+ * A ParallelValidator executes multiple {@link EntityValidator}s in parallel.
+ */
+public class ParallelValidator implements EntityValidator<JobDescriptor> {
+    private final Duration timeout;
+    private final Collection<EntityValidator<JobDescriptor>> hardValidators;
+    private final Collection<EntityValidator<JobDescriptor>> softValidators;
+
+    /**
+     * A ParallelValidator ensures that all its constituent validators complete within a specified duration.  The
+     * distinction between "hard" and "soft" validators is behavior on timeout.  When a "hard" validator fails to
+     * complete within the specified duration it produces a {@link ValidationError} with a {@link ValidationError.Type}
+     * of HARD.  Conversely "soft" validators which fail to complete within the specified duration have a
+     * {@link ValidationError.Type} of SOFT.
+     *
+     * @param timeout The duration within which each validator should complete.
+     * @param hardValidators Those validators which should generate a HARD {@link ValidationError} on timeout.
+     * @param softValidators Those validators which should generate a SOFT {@link ValidationError} on timeout.
+     */
+    public ParallelValidator(
+            Duration timeout,
+            Collection<EntityValidator<JobDescriptor>> hardValidators,
+            Collection<EntityValidator<JobDescriptor>> softValidators) {
+        this.timeout = timeout;
+        this.hardValidators = hardValidators;
+        this.softValidators = softValidators;
+    }
+
+    @Override
+    public Mono<Set<ValidationError>> validate(JobDescriptor jobDescriptor) {
+        Collection<Mono<Set<ValidationError>>> monos =
+                Stream.concat(
+                        getMonos(jobDescriptor, timeout, hardValidators, ValidationError.Type.HARD).stream(),
+                        getMonos(jobDescriptor, timeout, softValidators, ValidationError.Type.SOFT).stream())
+                        .collect(Collectors.toList());
+
+        return Mono.zip(
+                monos,
+                objects -> Arrays.stream(objects)
+                        .map(objectSet -> (Set<ValidationError>) objectSet)
+                        .flatMap(errorSet -> errorSet.stream())
+                        .collect(Collectors.toSet()))
+                .defaultIfEmpty(Collections.emptySet());
+    }
+
+    private static Collection<Mono<Set<ValidationError>>> getMonos(
+            JobDescriptor jobDescriptor,
+            Duration timeout,
+            Collection<EntityValidator<JobDescriptor>> validators,
+            ValidationError.Type errorType) {
+
+        return validators.stream()
+                .map(v -> new Pair<>(v.getClass().getSimpleName(), v.validate(jobDescriptor))) // (ValidatorClassName, Mono)
+                .map(pair -> pair.getRight().timeout(
+                        timeout,
+                        Mono.just(
+                                new HashSet<>(Arrays.asList(
+                                // Field: ValidatorClassName, Description: TimeoutMessage, Type: [SOFT|HARD]
+                                new ValidationError(pair.getLeft(), getTimeoutMsg(timeout), errorType))))))
+                .collect(Collectors.toList());
+    }
+
+    static String getTimeoutMsg(Duration timeout) {
+        return String.format("Timed out in %s ms", timeout.toMillis());
+    }
+}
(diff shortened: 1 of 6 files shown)


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
```diff
@@ -0,0 +1,78 @@
...
+                .defaultIfEmpty(Collections.emptySet());
+    }
+
+    private static Collection<Mono<Set<ValidationError>>> getMonos(
+            JobDescriptor jobDescriptor,
+            Duration timeout,
+            Collection<EntityValidator<JobDescriptor>> validators,
+            ValidationError.Type errorType) {
+
+        return validators.parallelStream()
```
andrew-leung: Perhaps more of a Java questions, will this fork a new thread for each stream? Does it draw from a thread pool that may get exhausted with a large number of validators?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
```diff
@@ -0,0 +1,78 @@
...
+            Collection<EntityValidator<JobDescriptor>> softValidators) {
+        this.timeout = timeout;
+        this.hardValidators = hardValidators;
+        this.softValidators = softValidators;
+    }
+
+    @Override
+    public Mono<Set<ValidationError>> validate(JobDescriptor jobDescriptor) {
+        Collection<Mono<Set<ValidationError>>> monos =
+                Stream.concat(
```
andrew-leung: When two parallel streams get concatenated, are they executed in parallel?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
```diff
@@ -0,0 +1,78 @@
...
+     * complete within the specified duration it produces a {@link ValidationError} with a {@link ValidationError.Type}
+     * of HARD.  Conversely "soft" validators which fail to complete within the specified duration have a
+     * {@link ValidationError.Type} of SOFT.
+     *
+     * @param timeout The duration within which each validator should complete.
+     * @param hardValidators Those validators which should generate a HARD {@link ValidationError} on timeout.
+     * @param softValidators Those validators which should generate a SOFT {@link ValidationError} on timeout.
+     */
+    public ParallelValidator(
+            Duration timeout,
```
andrew-leung: Presumably this would eventually move to a ValidatorConfiguration class in the future?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
gabrielhartmann: This is a concatenation of the monos.  This is just building a single Collection out of two collections.  The parallel execution occurs below in the `zip` command.  I manually tested and verified that `zip` does indeed execute in parallel.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
gabrielhartmann: [parallelStream doc](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#parallelStream--)

It may or not be parallel in actuality.  It is not consequential in this case and does not affect the actual parallel or sequential execution of validators.  `validators.stream()` would have worked just as well.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
```diff
@@ -0,0 +1,78 @@
...
+            Collection<EntityValidator<JobDescriptor>> hardValidators,
+            Collection<EntityValidator<JobDescriptor>> softValidators) {
+        this.timeout = timeout;
+        this.hardValidators = hardValidators;
+        this.softValidators = softValidators;
+    }
+
+    @Override
+    public Mono<Set<ValidationError>> validate(JobDescriptor jobDescriptor) {
```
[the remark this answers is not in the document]
gabrielhartmann: That would work as well.  I had a `Flux<Set<ValidationError>>` before, but was advised by @tbak to move to a `Mono`, but you're right that a `Flux<ValidationError>` would also make sense, arguably more sense.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
tbak: Why not streams()? There is no need to parallelize the collection processing in this case.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
gabrielhartmann: Yes.  I'll be figuring out the injection of a ParallelValidator with appropriate internal validators in a subsequent PR.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private static Collection<Mono<Set<ValidationError>>> getMonos(
+            JobDescriptor jobDescriptor,
+            Duration timeout,
+            Collection<EntityValidator<JobDescriptor>> validators,
+            ValidationError.Type errorType) {
+
+        return validators.parallelStream()
+                .map(v -> Tuples.of(v.getClass().getSimpleName(), v.validate(jobDescriptor))) // (ValidatorClassName, Mono)
+                .map(tuple2 -> tuple2.getT2().timeout(
```
tbak: We have own implementation of `Tuple`. The reactor one is not best, as it does not provide any operators.

titus-api/src/test/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidatorTest.java
```diff
@@ -0,0 +1,232 @@
...
+        EntityValidator pass = new PassJobValidator();
+        EntityValidator never = new NeverJobValidator();
+        EntityValidator validator = new ParallelValidator(TIMEOUT, Arrays.asList(pass, never), Collections.emptySet());
+        Mono<Set<ValidationError>> mono = validator.validate(MOCK_JOB);
+
+        StepVerifier.create(mono)
+                .expectNextMatches(errors -> errors.size() == 1)
+                .verifyComplete();
+
+        Set<ValidationError> errors = mono.block();
```
tbak: Why this check is not above in the expectNextMatches?

titus-api/src/test/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidatorTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
```
tbak: We use `assertj` in new junit classes.

titus-api/src/test/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidatorTest.java
gabrielhartmann: Why not below?  Ordering is unimportant here.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/ParallelValidator.java
gabrielhartmann: I'll change it, but why not `parallelStream()`?

gabrielhartmann:
@tbak It might be worth adding a preference for the [`assertj`](http://joel-costigliola.github.io/assertj/) testing library in [CONTRIBUTING.md](https://github.com/Netflix/titus-control-plane/blob/master/CONTRIBUTING.md).  A preference for a particular implementation of `Tuple` or [`Pair`](https://github.com/Netflix/titus-control-plane/blob/20536f85995c79ec2eec52996332b2ab8ed9ae58/titus-common/src/main/java/com/netflix/titus/common/util/tuple/Pair.java) might also be worth adding.

PR #372, opened 2018-09-26
Initial version of the container health API

tbak:
### Description of the Change

Container health API.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/containerhealth/model/ContainerHealthState.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/containerhealth/model/ContainerHealthState.java
@@ -1,0 +1,29 @@
+package com.netflix.titus.api.containerhealth.model;
+
+public enum ContainerHealthState {
+    /**
+     * The application process running in the container is healthy, and ready to handle user requests.
+     */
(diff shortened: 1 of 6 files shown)


titus-api/src/main/java/com/netflix/titus/api/containerhealth/model/ContainerHealthState.java
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * The health status of the application running in the container is not known.
+     */
+    Unknown,
+
+    /**
+     * The application or container are terminating.
+     */
+    Terminating,
```
corindwyer: I don't think we should have a `Terminating` state.

titus-api/src/main/java/com/netflix/titus/api/containerhealth/model/ContainerHealthState.java
```diff
@@ -0,0 +1,29 @@
+package com.netflix.titus.api.containerhealth.model;
+
+public enum ContainerHealthState {
```
corindwyer: Does it also make sense to have a `Starting` or similar state?

PR #373, opened 2018-09-27
add experimental agent attributes and remove AutoScaleRule and OverrideStatus

corindwyer:
### Description of the Change

* add experimental agent attributes
* remove AutoScaleRule and OverrideStatus


titus-common/src/main/java/com/netflix/titus/common/util/CollectionsExt.java
```diff
@@ -461,6 +461,23 @@ private CollectionsExt() {
         return true;
     }
 
+    /**
+     * @return true when the <tt>map</tt> contains any of the keys
+     */
+    @SafeVarargs
+    public static <T> boolean containsAnyKeys(Map<T, ?> map, T... keys) {
+        if (map == null) {
```
tbak: Or empty?
Same with the `keys` parameter.

titus-server-master/src/main/java/com/netflix/titus/master/agent/service/DefaultAgentManagementService.java
```diff
@@ -182,6 +172,16 @@ public Completable updateInstanceGroupAttributes(String instanceGroupId, Map<Str
                 }).toCompletable();
     }
 
+    @Override
+    public Completable updateAgentInstanceAttributes(String instanceId, Map<String, String> attributes) {
```
tbak: This is a full override. I think we should also provide merge/remove operations.

titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
```diff
@@ -340,8 +342,13 @@ Completable doAgentScaling() {
                                                                      long elapsed) {
         return instancesForActiveInstanceGroups.entrySet().stream()
                 .flatMap(e -> e.getValue().stream().filter(i -> {
-                    InstanceOverrideStatus overrideStatus = i.getOverrideStatus();
-                    return overrideStatus.getState() == InstanceOverrideState.Removable && hasTimeElapsed(overrideStatus.getTimestamp(), finish, elapsed);
+                    String removableTimestampValue = i.getAttributes().get(AgentAttributes.REMOVABLE);
```
tbak: A help function in the `AgentAttributes` class? This fetch and check logic will be executed in many other places.

titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
```diff
@@ -419,20 +426,19 @@ Completable doAgentScaling() {
...
             int agentCountToRemoveInInstanceGroup = Ints.min(remainingAgentsToRemove, agentCountEligibleToRemoveInInstanceGroup, agentsEligibleToRemoveInInstanceGroup.size());
             for (int i = 0; i < agentCountToRemoveInInstanceGroup; i++) {
                 AgentInstance agentInstance = agentsEligibleToRemoveInInstanceGroup.get(i);
-                InstanceOverrideStatus removableOverrideStatus = InstanceOverrideStatus.newBuilder()
-                        .withState(InstanceOverrideState.Removable)
-                        .withTimestamp(clock.wallTime())
-                        .withDetail("ClusterAgentAutoScaler setting to Removable to scale down the instance")
-                        .build();
-                Completable completable = agentManagementService.updateInstanceOverride(agentInstance.getId(), removableOverrideStatus);
+                Map<String, String> updatedAttributes = CollectionsExt.merge(
```
tbak: This is an example, where doing full attributes override at the GRPC layer is dangerous, and prone to race conditions.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/fitness/AgentManagementFitnessCalculator.java
```diff
@@ -88,6 +95,22 @@ public double calculateFitness(TaskRequest taskRequest, VirtualMachineCurrentSta
             }
         }
 
-        return NOT_ACTIVE_INSTANCE_GROUP_SCORE;
+        String instanceId = getAttributeValueOrEmptyString(targetVM, schedulerConfiguration.getInstanceAttributeName());
```
tbak: This section should be in the `if` statement above.

titus-server-master/src/main/java/com/netflix/titus/master/service/management/internal/SimpleCapacityGuaranteeStrategy.java
```diff
@@ -129,7 +129,7 @@ public CapacityAllocations compute(CapacityRequirements capacityRequirements) {
         Map<AgentInstanceGroup, Integer> result = new HashMap<>();
         for (Map.Entry<AgentInstanceGroup, Integer> entry : instanceAllocations.entrySet()) {
             AgentInstanceGroup instanceGroup = entry.getKey();
-            int minSize = instanceGroup.getAutoScaleRule().getMin();
+            int minSize = instanceGroup.getMin();
```
tbak: Is this valid? We want to enforce some min of min constraint. This was previously set in the auto scale rule. Now it should be 0.

titus-server-master/src/main/java/com/netflix/titus/master/agent/service/DefaultAgentManagementService.java
corindwyer: Yes, I believe we should remove this API and move to the style of add/remove attributes to be safer for multiple concurrent updates.

titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
corindwyer: I agree. Will fix in a new PR.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/fitness/AgentManagementFitnessCalculator.java
corindwyer: Yes, this code will never execute. I am going to remove this for now as we need to figure out what the behavior should be.

titus-server-master/src/main/java/com/netflix/titus/master/service/management/internal/SimpleCapacityGuaranteeStrategy.java
corindwyer: Will remove this as we don't need this behavior here anymore.

PR #376, opened 2018-09-29
Server side filters for JobManagement event streams

fabiokung:
### Description of the Change

Allow gRPC event streams (`JobManagement`) to be filtered, with the same criteria that can be used with `FindJobs` and `FindTasks`.

This depends on Netflix/titus-api-definitions#59. I will modify this PR once it gets merged with the dependencies updated.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/service/V3JobOperations.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/service/V3JobOperations.java
@@ -31,9 +31,8 @@ import com.netflix.titus.common.util.tuple.Pair;
 import rx.Completable;
 import rx.Observable;
 
-/**
- *
- */
+import static com.netflix.titus.common.util.FunctionExt.alwaysTrue;
+
 public interface V3JobOperations {
 
     String COMPONENT = "jobManagement";
@@ -88,7 +87,12 @@ public interface V3JobOperations {
      */
     Completable recordTaskPlacement(String taskId, Function<Task, Task> changeFunction);
 
-    Observable<JobManagerEvent<?>> observeJobs();
+    default Observable<JobManagerEvent<?>> observeJobs() {
+        return observeJobs(alwaysTrue(), alwaysTrue());
+    }
+
+    Observable<JobManagerEvent<?>> observeJobs(Predicate<Pair<Job<?>, List<Task>>> jobsPredicate,
+                                               Predicate<Pair<Job<?>, Task>> tasksPredicate);
 
     Observable<JobManagerEvent<?>> observeJob(String jobId);
 }
--- a/titus-common/src/main/java/com/netflix/titus/common/util/FunctionExt.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/FunctionExt.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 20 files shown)


titus-api/src/main/java/com/netflix/titus/api/jobmanager/service/V3JobOperations.java
```diff
@@ -88,7 +87,12 @@
      */
     Completable recordTaskPlacement(String taskId, Function<Task, Task> changeFunction);
 
-    Observable<JobManagerEvent<?>> observeJobs();
+    default Observable<JobManagerEvent<?>> observeJobs() {
+        return observeJobs(alwaysTrue(), alwaysTrue());
+    }
+
+    Observable<JobManagerEvent<?>> observeJobs(Predicate<Pair<Job<?>, List<Task>>> jobsPredicate,
```
tbak: For the `jobsPredicate` shouldn't we have only the `Job<?>` parameter?

titus-common/src/main/java/com/netflix/titus/common/util/FunctionExt.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.titus.common.util;
+
+import java.util.function.Predicate;
+
+public final class FunctionExt {
+    public static <T> Predicate<T> alwaysTrue() {
+        return ignored -> true;
```
tbak: Make it a constant to avoid object construction on each invocation.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/jobmanager/replicator/GrpcJobReplicatorEventStream.java
```diff
@@ -43,7 +44,8 @@ public GrpcJobReplicatorEventStream(JobManagementClient client,
         return Flux.defer(() -> {
             CacheUpdater cacheUpdater = new CacheUpdater();
             logger.info("Connecting to the job event stream...");
-            return ReactorExt.toFlux(client.observeJobs()).flatMap(cacheUpdater::onEvent);
+            return ReactorExt.toFlux(client.observeJobs(ObserveJobsQuery.newBuilder().build()))
```
tbak: `getDefaultInstance`?

titus-common/src/main/java/com/netflix/titus/common/util/FunctionExt.java
fabiokung: can't because of generic types. Unless I force the cast?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/service/V3JobOperations.java
fabiokung: discussed in person, this is making it more consistent with the `findJobs` API, where you can filter jobs (job events in this case) with criteria on their associated tasks.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/jobmanager/replicator/GrpcJobReplicatorEventStream.java
fabiokung: ah, nice, TIL. Will fix.

titus-common/src/main/java/com/netflix/titus/common/util/FunctionExt.java
tbak: Casting should be ok here.

fabiokung: OK, will do.

PR #379, opened 2018-10-01
Add application name to the metatron validation metrics tags

gabrielhartmann:
In order to track with applications are violating validation rules its helpful to tag their metrics with the application name.

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/jobmanager/client/GrpcJobManagementClient.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/jobmanager/client/GrpcJobManagementClient.java
@@ -104,7 +104,7 @@ public class GrpcJobManagementClient implements JobManagementClient {
         return ReactorExt.toObservable(validationErrors)
                 .flatMap(errors -> {
                     // Report metrics on all errors
-                    reportErrorMetrics(errors);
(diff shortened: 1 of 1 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/jobmanager/client/GrpcJobManagementClient.java
```diff
@@ -213,12 +213,13 @@ public Completable killTask(TaskKillRequest taskKillRequest) {
...
     }
 
-    private void reportErrorMetrics(Set<ValidationError> errors) {
+    private void reportErrorMetrics(Set<ValidationError> errors, JobDescriptor jobDescriptor) {
         errors.forEach(error ->
                 registry.counter(
                         error.getField(),
                         "type", error.getType().name(),
-                        "description", error.getDescription())
+                        "description", error.getDescription(),
```
tbak: Should this be bounded? I am not sure if there are limits on metric values in Atlas.
Is it guaranteed to be always non-null?

gabrielhartmann: Atlas bounds it.  It's a required field in all constructors.  A user of `ValidationError` would have to explicitly set it to `null`.

PR #381, opened 2018-10-02
allow commands to be modified by subclasses

fabiokung:
Allow more specialized CLIs to override builtin commands, and add new ones.


corindwyer:
Why do you want the ability to override commands in the CLI. Do you expect there to be multiple CLIs with different available commands?

fabiokung:
the use case was more about customizing flags and behavior on more specialized versions of the CLI. E.g. internally we use a custom mTLS auth layer that may require customizations that the builtin commands don't allow.

PR #383, opened 2018-10-02
Remove V2JobStatusMonitor and use LifecycleAgentStatusMonitor as the default instead

tbak:
### Description of the Change

Further V2 engine cleanup. `V2JobStatusMonitor` was the default `AgentStatusMonitor` in OSS, and
had to be replaced with something else. `LifecycleAgentStatusMonitor` is now the default agent 
health provider.


titus-server-master/src/main/java/com/netflix/titus/master/agent/service/monitor/LifecycleAgentStatusMonitor.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private Optional<AgentStatus> findStatus(String agentInstanceId) {
+        return agentManagementService.findAgentInstance(agentInstanceId).map(this::statusOf);
+    }
+
+    private AgentStatus statusOf(AgentInstance agent) {
+        InstanceLifecycleState state = agent.getLifecycleStatus().getState();
+        if (state == InstanceLifecycleState.Started) {
+            return AgentStatus.healthy(SOURCE_ID, agent, "Agent in Started state", System.currentTimeMillis());
```
corindwyer: use TItusRuntime clock?

titus-server-master/src/main/java/com/netflix/titus/master/agent/service/monitor/LifecycleAgentStatusMonitor.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return AgentStatus.unhealthy(SOURCE_ID, agent, "Agent not in Started state: currentState=" + state, System.currentTimeMillis());
```
corindwyer: same as above

titus-server-master/src/main/java/com/netflix/titus/master/agent/service/monitor/LifecycleAgentStatusMonitor.java
tbak: Fixed

titus-server-master/src/main/java/com/netflix/titus/master/agent/service/monitor/LifecycleAgentStatusMonitor.java
tbak: Fixed

PR #388, opened 2018-10-05
add job disruption budget core models and converters

corindwyer:
### Description of the Change

Add job disruption core models and converters between grpc and core.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/JobDescriptor.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/JobDescriptor.java
@@ -16,11 +16,16 @@
 
 package com.netflix.titus.api.jobmanager.model.job;
 
+import java.util.Collections;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.Function;
 import javax.validation.Valid;
 import javax.validation.constraints.Size;
 
+import com.netflix.titus.api.jobmanager.model.job.disruptionbudget.DisruptionBudget;
+import com.netflix.titus.api.jobmanager.model.job.disruptionbudget.SelfManagedDisruptionBudgetPolicy;
+import com.netflix.titus.api.jobmanager.model.job.disruptionbudget.UnlimitedDisruptionBudgetRate;
 import com.netflix.titus.api.jobmanager.model.job.ext.BatchJobExt;
 import com.netflix.titus.api.jobmanager.model.job.ext.ServiceJobExt;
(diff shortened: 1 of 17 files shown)


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/JobDescriptor.java
```diff
@@ -36,6 +41,13 @@
 @ClassInvariant(expr = "@asserts.notExceedsComputeResources(capacityGroup, container)", mode = VerifierMode.Strict)
 public class JobDescriptor<E extends JobDescriptor.JobDescriptorExt> {
 
+    private static final DisruptionBudget DEFAULT_DISRUPTION_BUDGET = DisruptionBudget.newBuilder()
+            .withDisruptionBudgetPolicy(SelfManagedDisruptionBudgetPolicy.newBuilder().build())
```
tbak: This could be a constant.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/disruptionbudget/UnlimitedDisruptionBudgetRate.java
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.api.jobmanager.model.job.disruptionbudget;
+
+public class UnlimitedDisruptionBudgetRate extends DisruptionBudgetRate {
```
tbak: Singleton?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/JobDescriptor.java
corindwyer: I agree. We should probably have a constants file for budgets with specific meanings.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/disruptionbudget/UnlimitedDisruptionBudgetRate.java
corindwyer: I did not make these singleton to be consistent with the other builders. I would like to see it this way in case we add fields in the future.

PR #401, opened 2018-10-17
 Add metrics/logging and REST endpoint for LocalScheduler

tbak:
### Description of the Change

Add metrics/logging and REST endpoint for LocalScheduler.


titus-common/src/main/java/com/netflix/titus/common/framework/scheduler/model/TransactionId.java
```diff
@@ -0,0 +1,70 @@
+package com.netflix.titus.common.framework.scheduler.model;
+
+import java.util.Objects;
+
+public class TransactionId {
+
+    private static final TransactionId INITIAL = new TransactionId(1, 0, 1);
+
+    private final int major;
```
corindwyer: I think the current object model is confusing. Talked offline for other approaches.

titus-common/src/main/java/com/netflix/titus/common/framework/scheduler/internal/LocalSchedulerTransactionLogger.java
```diff
@@ -0,0 +1,119 @@
...
+        String summary;
+        switch (schedulingState) {
+            case Waiting:
+                summary = "Waiting...";
+                break;
+            case Running:
+                summary = "Running...";
+                break;
+            case Cancelling:
+                summary = "Schedule canceled by a user";
```
corindwyer: canceled -> cancelled

titus-common/src/main/java/com/netflix/titus/common/framework/scheduler/model/TransactionId.java
tbak: Changed

titus-common/src/main/java/com/netflix/titus/common/framework/scheduler/internal/LocalSchedulerTransactionLogger.java
tbak: Fixed

PR #404, opened 2018-10-19
Task migration unit tests

tbak:
### Description of the Change

This commit includes also a major refactoring of the data generation framework
for job, agent and eviction components.


titus-api/src/main/java/com/netflix/titus/api/agent/model/AgentFunctions.java
```diff
@@ -47,4 +48,18 @@ private AgentFunctions() {
...
+
+    public static Function<AgentInstanceGroup, AgentInstanceGroup> withId(String newId) {
+        return ig -> ig.toBuilder().withId(newId).build();
+    }
+
+    public static Function<AgentInstanceGroup, AgentInstanceGroup> inState(InstanceGroupLifecycleState state) {
+        return ig -> ig.toBuilder().withLifecycleStatus(
+                InstanceGroupLifecycleStatus.newBuilder()
+                        .withState(state)
+                        .withTimestamp(System.currentTimeMillis())
```
corindwyer: Shouldn't this be based on the clock abstraction?

titus-api/src/main/java/com/netflix/titus/api/agent/model/AgentFunctions.java
```diff
@@ -47,4 +48,18 @@ private AgentFunctions() {
...
+                        .withDetail("No reason given")
```
corindwyer: Shouldn't this force a reason in the contract as there is generally always a reason to doing the transition.

titus-api/src/main/java/com/netflix/titus/api/agent/model/AgentFunctions.java
tbak: Fixed

titus-api/src/main/java/com/netflix/titus/api/agent/model/AgentFunctions.java
tbak: I created `AgentTestFunctions` and moved this method there.

PR #411, opened 2018-10-25
moving policyName, resourceId in a util class

amit-git:
### Minor Refactoring

refactoring AWSAppAutoScalingClient in order to isolate logic for building entities such as policy name, resource Id. Also removes service namespace and scalable dimension into a publicly accessible location.

### DIFF
--- a/titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/appscale/AWSAppAutoScalingClient.java
+++ b/titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/appscale/AWSAppAutoScalingClient.java
@@ -65,6 +65,10 @@ import static com.netflix.titus.ext.aws.appscale.AWSAppAutoScalingUtil.buildScal
 @Singleton
 public class AWSAppAutoScalingClient implements AppAutoScalingClient {
     private static Logger logger = LoggerFactory.getLogger(AWSAppAutoScalingClient.class);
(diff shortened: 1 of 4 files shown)


titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/appscale/AWSAppScalingConfig.java
```diff
@@ -19,6 +19,9 @@
 import com.netflix.archaius.api.annotations.PropertyName;
 
 public interface AWSAppScalingConfig {
+    String SERVICE_NAMESPACE = "custom-resource";
+    // AWS requires this field be set to this specific value for all application-autoscaling calls.
+    String SCALABLE_DIMENSION = "custom-resource:ResourceType:Property";
```
corindwyer: If these are constants then I think you should just move them to the client and make them `public static final`.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/appscale/AWSAppAutoScalingClient.java
```diff
@@ -65,6 +65,10 @@
 @Singleton
 public class AWSAppAutoScalingClient implements AppAutoScalingClient {
     private static Logger logger = LoggerFactory.getLogger(AWSAppAutoScalingClient.class);
```
corindwyer: these 3 statics should also be final

PR #412, opened 2018-10-25
Move the eviction service to spring-reactor

tbak:
### Description of the Change

Part of RxJava -> Spring Reactor migration.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/rx/ReactorExt.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/rx/ReactorExt.java
@@ -163,6 +163,45 @@ public final class ReactorExt {
     }
 
     /**
+     * In case subscriber does not provide exception handler, the error is propagated back to source, and the stream is broken.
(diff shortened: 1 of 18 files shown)


titus-common/src/main/java/com/netflix/titus/common/util/rx/ReactorExt.java
```diff
@@ -162,6 +162,45 @@ public static void safeDispose(Disposable... disposables) {
...
+        return Flux.create(sink -> {
+            Disposable disposable = unprotectedStream.subscribe(
+                    event -> {
+                        try {
+                            sink.next(event);
+                        } catch (Exception e) {
+                            try {
+                                sink.error(e);
+                            } catch (Exception e2) {
+                                logger.warn("Subscriber threw an exception from onNext handler", e);
```
corindwyer: Shouldn't this line be in the try before the `sink.error` or before the try in general?

tbak: Fixed.

PR #432, opened 2018-11-16
TitusFederation does not forward the extension headers

tbak:


titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobManagementClient.java
```diff
@@ -413,6 +427,10 @@ private static JobChangeNotification buildJobSnapshotEndMarker() {
         return JobChangeNotification.newBuilder().setSnapshotEnd(marker).build();
     }
 
+    private JobManagementServiceStub wrap(Optional<CallMetadata> context, JobManagementServiceStub client) {
```
fabiokung: should we combine both `wrap` methods to reduce the chance of someone forgetting to wrap stubs twice?

fabiokung: that would also allow us to move wrapping stubs into other helper methods that are already calling the other `wrap(JobManagementServiceStub)`

tbak: I agree that the code today is complex, and we have many wrappers here. Doing proper cleanup will take more time however, so I would prefer to postpone it.

PR #433, opened 2018-11-19
Persist disruption budget; guard it with a feature flag

tbak:


titus-common/src/main/java/com/netflix/titus/common/util/feature/FeatureGuardWhiteListConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+ * Feature activation configuration. Each instantiation of this configuration should start with a feature specific prefix.
+ */
+public interface FeatureGuardWhiteListConfiguration {
+
+    /**
+     * If set to false, the feature is disabled. If set to true, the feature activation is determined by evaluating
+     * the remaining predicates.
+     */
+    @DefaultValue("true")
+    boolean isTurnedOn();
```
corindwyer: isFeatureEnabled?

tbak: Renamed in another PR.

PR #436, opened 2018-11-20
Add feature flag for move tasks api

amitaekbote:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
```diff
@@ -390,6 +390,9 @@ public Completable recordTaskPlacement(String taskId, Function<Task, Task> chang
 
     @Override
     public Observable<Void> moveServiceTask(String sourceJobId, String targetJobId, String taskId) {
+        if(!jobManagerConfiguration.enableMoveTaskApi()){
```
tbak: To follow the JavaBean/Archaius naming convention, it should be `isMoveTaskApiEnabled`.

PR #441, opened 2018-11-28
add property to force leader election

corindwyer:
### Description of the Change

Add property to force leader election on a titus master instance.


titus-server-master/src/main/java/com/netflix/titus/master/supervisor/service/leader/LeaderElectionOrchestrator.java
```diff
@@ -58,15 +59,18 @@
     private final TitusRuntime titusRuntime;
     private final Scheduler scheduler;
 
-    private final Subscription localMasterUpdateSubscription;
-    private final SpectatorExt.FsmMetrics<MasterState> stateFsmMetrics;
+    private Subscription localMasterUpdateSubscription;
```
fabiokung: @corindwyer minor nit here: probably a good idea to make these `volatile`, AFAICT there is nothing guaranteeing visibility between the (guice) thread that instantiates this singleton, and the rxjava threads that are going to run the subscription and execute the callbacks/actions.

PR #452, opened 2018-12-20
handle LoadBalancers (TargetGroups on AWS) removed

fabiokung:
When a LoadBalancer is removed outside of Titus, the reconciliation loop will (on each iteration):

1. Detect the load balancer gone, mark all its job associations as orphan.
2. Update orphan associations as Dissociated on the store.
3. Reap (delete) dissociated records when no more operations are pending.


titus-api/src/main/java/com/netflix/titus/api/connector/cloud/LoadBalancer.java
```diff
@@ -0,0 +1,58 @@
...
+    public State getState() {
+        return state;
+    }
+
+    public Set<String> getRegisteredIps() {
+        return registeredIps;
+    }
+
+    @Override
+    public String toString() {
```
tbak: hash/equals for completeness?

titus-api/src/main/java/com/netflix/titus/api/connector/cloud/LoadBalancerConnector.java
```diff
@@ -37,5 +37,5 @@
      */
     Completable isValid(String loadBalancerId);
 
-    Single<Set<String>> getRegisteredIps(String loadBalancerId);
+    Single<LoadBalancer> getLoadBalancer(String loadBalancerId);
```
tbak: Whenever we modify or create a new method with RX result, we should use Spring-Reactor, not RxJava.
If we do not, we will have to do it anyway in a few months from now.

tbak: I have created a little script to estimate RxJava vs Spring-Reactor usage in our code base: https://github.com/Netflix/titus-control-plane/blob/master/rx.sh

fabiokung:
@gabrielhartmann @corindwyer @andrew-leung PTAL

PR #465, opened 2019-01-29
JobManagement GRPC mutating requests authorization (initial)

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
```diff
@@ -436,16 +441,29 @@ public void killJob(JobId request, StreamObserver<Empty> responseObserver) {
...
-                    e -> safeOnError(logger, e, responseObserver),
-                    () -> {
-                        responseObserver.onNext(Empty.getDefaultInstance());
-                        responseObserver.onCompleted();
-                    }
-            );
+            ReactorExt.toObservable(authorizationService.authorize(callMetadata, jobTaskPair.getLeft()))
+                    .flatMap(authorizationResult -> {
+                        if (!authorizationResult.isAuthorized()) {
+                            return Observable.error(new IllegalArgumentException(authorizationResult.getReason()));
```
corindwyer: Do we need a different exception here? How do you expect to map these types of error to `Forbidden` and `PermissionDenied`?

tbak: Good point. I will fix that.

PR #466, opened 2019-01-29
Signal that a task has been moved in TaskUpdateEvents

fabiokung:
Task events from `JobOperations` will now have a flag indicating when a task moved from another job. This allows subscribers to update their indexes and act accordingly.

This PR also adds handling of tasks being moved to the `LoadBalancerEngine`, so IPs get properly deregistered/registered as tasks move to a different Job.

The information about tasks moving is not being added to the public gRPC APIs yet. That will come in another PR, along with changes to some components already using the gRPC event stream. I added some `TODOs` where changes will be required.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/event/TaskUpdateEvent.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/event/TaskUpdateEvent.java
@@ -25,8 +25,6 @@ import com.netflix.titus.api.jobmanager.model.job.Task;
 public class TaskUpdateEvent extends JobManagerEvent<Task> {
 
     private final Job<?> currentJob;
-    private final Task currentTask;
-    private final Optional<Task> previousTask;
     private final boolean movedFromAnotherJob;
 
     private TaskUpdateEvent(Job<?> currentJob, Task currentTask, Optional<Task> previousTask) {
@@ -36,8 +34,6 @@ public class TaskUpdateEvent extends JobManagerEvent<Task> {
     private TaskUpdateEvent(Job<?> currentJob, Task currentTask, Optional<Task> previousTask, boolean moved) {
         super(currentTask, previousTask);
         this.currentJob = currentJob;
-        this.currentTask = currentTask;
-        this.previousTask = previousTask;
         this.movedFromAnotherJob = moved;
     }
 
@@ -46,11 +42,11 @@ public class TaskUpdateEvent extends JobManagerEvent<Task> {
     }
 
     public Task getCurrentTask() {
-        return currentTask;
+        return getCurrent();
     }
 
     public Optional<Task> getPreviousTask() {
-        return previousTask;
+        return getPrevious();
     }
 
     public boolean isMovedFromAnotherJob() {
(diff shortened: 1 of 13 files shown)


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/event/TaskUpdateEvent.java
```diff
@@ -46,43 +53,51 @@ public Task getCurrentTask() {
...
     }
 
     @Override
     public String toString() {
         return "TaskUpdateEvent{" +
                 "currentJob=" + currentJob +
-                ", current=" + getCurrent() +
-                ", previous=" + getPrevious() +
-                "}";
+                ", currentTask=" + currentTask +
```
corindwyer: The old `toString` used to call the super classes methods to print out information.

titus-server-master/src/main/java/com/netflix/titus/master/loadbalancer/service/LoadBalancerEngine.java
```diff
@@ -163,33 +168,78 @@ public void shutdown() {
...
+            }
+            Set<JobLoadBalancer> sourceJobLoadBalancers = store.getAssociatedLoadBalancersSetForJob(sourceJobId);
+
+            // operations below will dedup load balancers present in both source and target Sets, so we avoid
+            // unnecessary churn by deregistering and registering the same targets on the same loadBalancerId
+
+            // register on load balancers associated with the target job
+            if (TaskHelpers.isStartedWithIp(task)) {
+                // targetJobLoadBalancers - sourceJobLoadBalancers
+                Set<JobLoadBalancer> toRegister = targetJobLoadBalancers.stream()
```
tbak: The same filtering logic is used in two places. This could be extracted to a method and reused.

fabiokung: I had it extracted before, but the two minor differences (the desired state, and the logging line) made it so extracting it ended in more lines of code than have it separate. I will give it another shot.

fabiokung: nvm, I was thinking on a different part of the PR. I will extract this.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/jobmanager/replicator/GrpcJobReplicatorEventStream.java
```diff
@@ -149,12 +149,13 @@ public GrpcJobReplicatorEventStream(JobManagementClient client,
                     coreEvent = toJobCoreEvent(job);
                     break;
                 case TASKUPDATE:
+                    //TODO(fabio): handle task moved when it is added to the gRPC API definition
                     com.netflix.titus.grpc.protogen.Task task = event.getTaskUpdate().getTask();
-                    Job<?> taskJob = lastSnapshot.getJobs().stream().filter(j -> j.getId().equals(task.getJobId())).findFirst().orElse(null);
-                    if (taskJob != null) {
-                        Task coreTask = V3GrpcModelConverters.toCoreTask(taskJob, task);
+                    Optional<Job<?>> taskJob = lastSnapshot.findJob(task.getJobId());
```
corindwyer: can you rename this to something like `taskJobOpt` and use `taskJob` as a local variable if it is present?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/event/TaskUpdateEvent.java
fabiokung: I don't think so. Maybe you are reffering to `equals()` and `hashcode()`? The super class doesn't even have a `toString`, and our current template prints out the class type, which would be awkward when calling from a subclass.

fabiokung: I will remove the duplicated fields and use the superclass for all of it.

PR #476, opened 2019-02-07
Require that jobs are compatible when moving tasks

fabiokung:
Require that jobs are compatible for tasks to be moved across them.

### Description of the Change

Jobs are compatible when their descriptors are identical, ignoring the following values:

* Owner
* Application name
* Job group info (stack, details, sequence)
* Disruption budget
* Any attributes **not** prefixed with `titus.` or `titusParameter.`
* Any container attributes **not** prefixed with `titus.` or `titusParameter.`
* All information specific to Service jobs (extensions): capacity, retry policy, etc

Validation errors will include a detailed message, with a diff explaining what incompatibilities were found. I used protobuf representations of `JobDescriptor` to compute the differences. I started by handcrafting my own `Differencer`, but soon realized there are too many edge cases with different protobuf field types and unknown fields.

There is [a reference implementation for a `ProtobufDifferencer` in C++](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.util.message_differencer), and the closest port to Java I could find is in google/truth: [ProtoTruthMessageDifferencer](https://github.com/google/truth/blob/master/extensions/proto/src/main/java/com/google/common/truth/extensions/proto/ProtoTruthMessageDifferencer.java). It doesn't allow for much message customization (or access to the internal raw diff results), but it seems to be good enough. It is only used on the non-happy path, when validations fail.


fabiokung:
I will update the documentation on `titus-api-definitions` with the rules here after this gets merged.

titus-server-master/src/test/java/com/netflix/titus/master/jobmanager/service/integration/MoveTaskTest.java
```diff
@@ -88,10 +90,34 @@ public void testMoveWithInvalidTargetJob() {
...
+                        .withName("other/image")
+                        .build()
+                )
+        );
+        String targetJobId = startNewJob(incompatible).getJobId();
+
+        try {
+            sourceJobBuilder.advance()
+                    .moveTask(0, 0, sourceJobId, targetJobId)
+                    .expectJobEvent(job -> assertThat(JobFunctions.getJobDesiredSize(job)).isEqualTo(0));
```
tbak: Is the last step needed? Exception should be thrown from the `moveTask` step?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/JobCompatibility.java
```diff
@@ -0,0 +1,112 @@
...
+        JobDescriptor<ServiceJobExt> jobFromNormalized = unsetIgnoredFieldsForCompatibility(from);
+        JobDescriptor<ServiceJobExt> jobToNormalized = unsetIgnoredFieldsForCompatibility(to);
+        boolean identical = jobFromNormalized.equals(jobToNormalized);
+        return new JobCompatibility(jobFromNormalized, jobToNormalized, identical);
+    }
+
+    private static JobDescriptor<ServiceJobExt> unsetIgnoredFieldsForCompatibility(JobDescriptor<ServiceJobExt> descriptor) {
+        HashMap<String, String> onlyTitusAttributes = new HashMap<>(descriptor.getAttributes());
+        onlyTitusAttributes.entrySet().removeIf(entry ->
+                !entry.getKey().startsWith("titus.") && !entry.getKey().startsWith("titusParameter.")
```
corindwyer: constants for `titus.` and `titusParameter.`

amit-git: should we ignore JobGroupInfo (stack/detail/seq) as well ?

titus-server-master/src/test/java/com/netflix/titus/master/jobmanager/service/integration/MoveTaskTest.java
fabiokung: True, copy-and-pasta. I will fix.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/JobCompatibility.java
fabiokung: it is already being ignored on the job descriptor directly. These are not attributes.

PR #477, opened 2019-02-08
Add initial support for validating Job IAM roles on submission.

andrew-leung:
This PR adds initial support for validating a Job's IAM role on submission. There are additional changes in the internal repo as well.

The changes here include an AWS connector that is disabled by default, with a noop connector favored by default. The validator is configured to fail open (i.e., a soft validator) in this change.

The validator checks that an IAM role exists and that a provided IAM profile is able to assume into that role. It does not currently do cross-account assume verification and skips validating roles that would require the Gateway to have assume access into other accounts.


titus-api/src/main/java/com/netflix/titus/api/iam/model/IamRole.java
```diff
@@ -0,0 +1,39 @@
...
+ *  * limitations under the License.
+ *
+ */
+
+package com.netflix.titus.api.iam.model;
+
+/**
+ * Interface for an IAM Role Policy.
+ */
+public interface IamRole {
```
tbak: This class bundles data type (`IamRole`), with behavior (`canAssume`). I think canAssume should be part of `IamConnector`, or we should expose here a getter that provides all roles the given one can assume into.

We should also make the data type `IamRole` a regular JavaBean (no interface).

titus-api/src/main/java/com/netflix/titus/api/iam/service/IamConnectorException.java
```diff
@@ -0,0 +1,44 @@
...
+    public IamConnectorException(ErrorCode errorCode, String message) {
+        this(errorCode, message, new RuntimeException(message));
+    }
+
+    private IamConnectorException(ErrorCode errorCode, String message, Throwable cause) {
+        super(message, cause);
+        this.errorCode = errorCode;
+    }
+
+    public ErrorCode getErrorCode() { return errorCode; }
```
tbak: Nit pick. We use error factory methods for better exception formatting (see `JavaManagerException`).

titus-common/src/main/java/com/netflix/titus/common/network/client/RxRestClientConfiguration.java
```diff
@@ -31,4 +35,8 @@
 
     @DefaultValue("200")
     long getRetryDelayMs();
+
+    default Set<HttpResponseStatus> getNoRetryStatuses() {
```
tbak: Is it possible to configure this via Archaius? Maybe it should be `Set<String>`, which would be mapped to `HttpResponseStatus` internally?

corindwyer: What is the purpose of this method? Wouldn't retry behavior be dependent on the specific API contract that is being called?

corindwyer:
This PR is taking an IAM Role name that is provided on the request and attempting to validate the policy. The downside to this approach that we need to continue accepting multiple flavors of role names, inspecting the policy, and guessing if we will be able to assume based on what is in the policy. I suggest changing the behavior to the following:
1. if an IAM role is not provided then apply the default IAM ARN.
2. if an IAM role is provided and does not start with the ARN prefix, prepend the ARN prefix.
3. validate that the ARN is in the correct string format.
4. check to in local memory cache is the ARN has previously been validated and continue if it has.
5. if it does not exist in the local cache, make a sts assume role call with the arn and put the result into local cache.

By moving to this approach, we don't have to infer what the sts assume role api logic is and we can start to move towards enforcing that an ARN is sent into the Titus API.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamRole.java
```diff
@@ -0,0 +1,151 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Describes an AWS IAM Role Policy.
+ */
+public class AwsIamRole implements IamRole {
+    private static final Logger logger = LoggerFactory.getLogger(AwsIamRole.class);
+
+    private static final String assumeAction = "sts:AssumeRole";
```
fabiokung: minor nit on naming conventions for constants `ALL_UPPERCASE`

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsAssumePrincipalDeserializer.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public AwsAssumePrincipal deserialize(JsonParser jp, DeserializationContext ctxt)
+            throws IOException {
+        List<String> principals = new ArrayList<>();
+
+        JsonNode node = jp.getCodec().readTree(jp);
+        // Find the specific key
+        if (node.hasNonNull(AWS_KEY)) {
```
fabiokung: should deserialization throw an exception if the format is not as expected? The current code silently ignores unexpected formats.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamRole.java
```diff
@@ -0,0 +1,151 @@
...
+        return canAssume(this.assumePolicy, assumeResourceName);
+    }
+
+    public static Boolean canAssume(String assumePolicy, String assumeResourceName) {
+        try {
+            // AWS provides us a URL encoded policy doc
+            String assumePolicyJson = URLDecoder.decode(assumePolicy, StandardCharsets.UTF_8.toString());
+
+            ObjectMapper objectMapper = ObjectMappers.defaultMapper();
+            AwsAssumePolicy awsAssumePolicy = objectMapper.readValue(assumePolicyJson, AwsAssumePolicy.class);
```
fabiokung: see my other comment on the `ObjectMapper` itself, this will never throw an exception and will silently swallow errors with unexpected and malformed JSONs.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamRole.java
```diff
@@ -0,0 +1,151 @@
...
+
+            // Check if there is a policy statement that allows assumeRole for the provided resource
+            for (AwsAssumeStatement statement : awsAssumePolicy.getStatements()) {
+                if (statement.getAction().equals(assumeAction) &&
+                        statement.getEffect().equals(assumeEffect) &&
+                        null != statement.getPrincipal().getPrincipals()) {
```
fabiokung: do the `null` check first in the chain? Preferably use `CollectionsExt.isNotEmpty()`

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamRole.java
```diff
@@ -0,0 +1,151 @@
...
+                    for (String principalName : statement.getPrincipal().getPrincipals()) {
+                        // Note this checks the resource name matches but does not check the account specifics.
```
fabiokung: minor nit: extract this logic to another method

titus-server-gateway/src/test/java/com/netflix/titus/gateway/service/v3/internal/JobSecurityValidatorTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.junit.Test;
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class JobSecurityValidatorTest {
+    private static final String validIamRoleName = "myValidIamRole";
```
fabiokung: same nit on `CONSTANT_NAME_CONVENTIONS`

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobIamValidator.java
```diff
@@ -0,0 +1,133 @@
...
+
+        String iamRoleName = jobDescriptor.getContainer().getSecurityProfile().getIamRole();
+
+        // Skip validation if no IAM was provided because a valid default will be used.
+        if (iamRoleName.isEmpty()) {
+            registry.counter(validationSkippedId.withTag(skippedMetricReasonTag, "noneProvided")).increment();
+            return Mono.just(Collections.emptySet());
+        }
+
+        // Skip any IAM that is not in "friendly" format. A non-friendly format is
```
fabiokung: is this expected to come in the future, or is the plan to never validate these?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobSecurityValidatorConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.netflix.titus.runtime.endpoint.validator;
+
+import com.netflix.archaius.api.annotations.Configuration;
+import com.netflix.archaius.api.annotations.DefaultValue;
+
+@Configuration(prefix = "titus.validate.job.security")
+public interface JobSecurityValidatorConfiguration {
+    @DefaultValue("true")
+    boolean iamValidatorEnabled();
```
fabiokung: doesn't archaius expect JavaBeans conventions for accessors here? (i.e.: `isIamValidatorEnabled()`)

fabiokung:
@corindwyer's proposal is sound, but the local validation cache will probably not be very effective on gateways, since they are shared-nothing instances and ephemeral. If the cache hit is too low, we could consider moving the validation into master, or some other caching method.

fabiokung:
Another problem with trying to `STS:AssumeRole` is that gateways and masters don't share the same IAM credentials with agents (which will do the actual `STS:AssumeRole` call).

corindwyer:
My proposal assumed that the credentials that the assume calls are being made with are the same as the agent. For this particular validator, it would first assume into the agent role and use that role to make additional calls. It would then refresh the credentials periodically as the STS token should expire after an hour or so. Also, the local cache should be fine as it is really for the high volume job submission use cases which are typically using a very small set of IAM Roles.

titus-api/src/main/java/com/netflix/titus/api/iam/model/IamRole.java
andrew-leung: Discussed offline. We'll move into the connector. Though moving away from subtyped IamRoles means another theoretical cloud connector's `IamRole` won't be explicit and deserializing the assume policy document in the connector may barf. Not ideal, but keeps the `IamRole` simpler.

titus-api/src/main/java/com/netflix/titus/api/iam/service/IamConnectorException.java
andrew-leung: Yeah, will change.

titus-common/src/main/java/com/netflix/titus/common/network/client/RxRestClientConfiguration.java
andrew-leung: The purpose is to tell the Retryable HTTP client which codes to not retry on. Otherwise it retries on all (e.g., it makes 3 GET attempts when the first response is a 404). When a specific connector creates its HTTP client it specifies whatever values make sense based on its contract, which is what is done if nftitus.

I'll update to a List<String> for Archaius, though the default method is the recommended way to define the collection value (no DefaultValue for collections).

andrew-leung:
I agree with at some point fully building the ARN because we'll want to validate those once most IAMs are cross-account.

My main reason for not doing the AssumeRole in the validation is given the power of the Agent's role (e.g., can assume into many other roles), it seems preferable to not also allow the Gateway role to assume into the Agent's role (then the Gateway can assume many app's roles). Even for cross-account IAM validation we can scope the Gateway's role for narrow access into the target account.

The couple other things I liked about using Edda. We don't need to write/manage a cache (e.g., stuff like not worrying about caching stale AssumeState for an hour when Edda refreshes ever 5 minutes) so this seemed quicker. Also the most common IAM misconfiguration is a bunch of Jobs being launched with a role that Titus can't assume, which we can validate from Edda without making any AWS calls.

There will need to be a V2 of this when we reorganize accounts anyways. Happy to discuss looking at using AssumeRole at that point. Unless there's major disagreement with not doing Assume now, I'd like to move forward with this approach (and address other PR comments).

titus-api/src/main/java/com/netflix/titus/api/connector/cloud/noop/NoopIamRole.java
```diff
@@ -0,0 +1,54 @@
...
+package com.netflix.titus.api.connector.cloud.noop;
+
+import com.netflix.titus.api.iam.model.IamRole;
+
+/**
+ * Describes a generic IAM Role Policy. This class is intended for testing.
+ */
+public class NoopIamRole implements IamRole {
+
+    private final static String noopString = "";
```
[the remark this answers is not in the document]
andrew-leung: Good catch. Copy/pasta.

titus-api/src/main/java/com/netflix/titus/api/iam/model/IamRole.java
```diff
@@ -0,0 +1,39 @@
...
+
+    String getResourceName();
+
+    String getAssumePolicy();
+
+    /**
+     * Returns true of the provided resource name can assume into
+     * this role based on its assume policy.
+     */
+    Boolean canAssume(String assumeResourceName);
```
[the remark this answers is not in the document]
andrew-leung: Yeah, will change. Wasn't intentional.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamRole.java
andrew-leung: Will change.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsAssumePrincipalDeserializer.java
andrew-leung: An IOException is thrown if basic JSON parsing fails, but if the `AWS_KEY` we're looking for is missing it materializes as a failed validation to the user as the required keys won't be found. Maybe we should log something?

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamRole.java
andrew-leung: Not sure if my previous comment helps here...

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamRole.java
andrew-leung: Yeah, returning an empty collection is a good idea.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamRole.java
andrew-leung: Yeah, good idea.

titus-server-gateway/src/test/java/com/netflix/titus/gateway/service/v3/internal/JobSecurityValidatorTest.java
andrew-leung: Will update.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobIamValidator.java
andrew-leung: Yes, expected to come in the future.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobSecurityValidatorConfiguration.java
andrew-leung: Not sure, but happy to change it.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AwsModule.java
```diff
@@ -35,6 +36,7 @@ protected void configure() {
         bind(AmazonAutoScalingAsync.class).toProvider(AmazonAutoScalingAsyncProvider.class);
         bind(AmazonElasticLoadBalancingAsync.class).toProvider(AmazonElasticLoadBalancingAsyncProvider.class);
         bind(AmazonAutoScaling.class).toProvider(AmazonAutoScalingProvider.class);
+        bind(AmazonIdentityManagementAsync.class).toProvider(AmazonIamAsyncProvider.class);
```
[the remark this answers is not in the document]
andrew-leung: Ah, good catch. Should be singleton.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamConnector.java
```diff
@@ -0,0 +1,134 @@
...
+                            String.format("Unable to query IAM %s: %s", iamRoleName, throwable.getMessage()));
+                });
+    }
+
+    private Mono<GetRoleResult> getAwsIamRole(String iamRoleName) {
+        GetRoleRequest request = new GetRoleRequest().withRoleName(iamRoleName);
+        return toMono(request, iamClient::getRoleAsync);
+    }
+
+    private <REQUEST extends AmazonWebServiceRequest, RESPONSE> Mono<RESPONSE> toMono(
```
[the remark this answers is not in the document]
andrew-leung: Yeah, probably good to start putting common AWS Reactor helpers somewhere.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsAssumePrincipal.java
```diff
@@ -39,6 +40,9 @@ public AwsAssumePrincipal(@JsonProperty("Principal") List<String> principals) {
     }
 
     public List<String> getPrincipals() {
+        if (null == principals) {
```
tbak: As this object is immutable, this check can be done in the constructor.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamUtil.java
```diff
@@ -0,0 +1,75 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(AwsIamUtil.class);
+
+    private static final String ASSUME_ACTION = "sts:AssumeRole";
+    private static final String ASSUME_EFFECT = "Allow";
+
+    /**
+     * Helper method to deserialize an IAM Role's assume policy and verify if
+     * another role can assume into it.
+     */
+    public static boolean canAssume(IamRole iamRole, String assumeResourceName) {
```
tbak: Some unit tests would be great here. They would also server as a documentation for the IAM data parsed/processed here.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsAssumePrincipal.java
andrew-leung: Good point.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamUtil.java
andrew-leung: Will add a test with some documentation of the policy format.

PR #478, opened 2019-02-11
 Extend the CallMetadata model with more caller data

tbak:
It is a major rewrite of the `CallMetadata` model (see corresponding changes in the GRPC IDL here https://github.com/Netflix/titus-api-definitions/pull/82).

The extended model includes a `Caller` data type, with a caller id, caller type and a context (key/value map). The new data model allows us to capture complete information about both the caller and the intermediary relying the request, as well as the invoked service endpoints.


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/metadata/CallMetadataHeaders.java
```diff
@@ -16,23 +16,98 @@
...
+     */
+    public static final String DIRECT_CALLER_CONTEXT_INSTANCE_ID = "titus.caller.instanceId";
+
+    /**
+     * Name of the service that is invoked.
+     */
+    public static final String DIRECT_CALLER_CONTEXT_SERVICE_NAME = "titus.service.name";
+
+    /**
+     * Nome of the method that is invoked.
```
amitaekbote: s/Nome/Name

tbak: Changed

PR #482, opened 2019-02-13
Run all integration tests on CI

fabiokung:
Not all tests were being executed on Travis CI.

@tbak this is why CI didn't catch the transitive dependencies issue a few commits ago.


fabiokung:
@tbak better now, the `ZookeeperMasterMonitorTest` faillures look legit and real problems

fabiokung:
Now part of #483.

PR #485, opened 2019-02-14
retry builder fix for reactor based unlimited retries

amit-git:
### Retry builder logic fix for unlimited retries for reactor stream

Reactor based retry builder code uses FluxInterval to generate timer events. It throws IllegalArgumentException when period < 0. The fix is meant to address this issue


titus-common/src/main/java/com/netflix/titus/common/util/rx/RetryHandlerBuilder.java
```diff
@@ -150,7 +151,7 @@ public RetryHandlerBuilder but() {
                         );
                         return Observable.error(new IOException(errorMessage, retryItem.cause));
                     }
-                    long expDelay = Math.min(maxDelay, (1 << retryItem.retry) * retryDelayMs);
```
tbak: `Math.min(31, retry)`?

PR #495, opened 2019-02-28
Use different GRPC Channel per client connector

tbak:
So we are not limited to connecting always to the same server type. For example we may connect to TitusFederation to get jobs/tasks, but when fetching agent data we have to call TitusGateway directly as the agent management API is not exposed via TitusFederation.


fabiokung:
probably worth noting that this will only work for single-cell stacks

titus-testkit/src/main/java/com/netflix/titus/testkit/perf/load/LoadGeneratorMain.java
```diff
@@ -0,0 +1,124 @@
...
+    @Named(GATEWAY_CHANNEL)
+    public Channel getGatewayChannel(ApplicationArguments arguments) {
+        List<String> gateway = arguments.getOptionValues("gateway");
+
+        String gatewayAddress = CollectionsExt.isNullOrEmpty(gateway) ? "localhost:8091" : gateway.get(0);
+
+        return NettyChannelBuilder
+                .forTarget(gatewayAddress)
+                .usePlaintext(true)
+                .maxHeaderListSize(65536)
```
corindwyer: constant? this value is copied and pasted a lot

tbak: Good idea. Most of this code could be extracted to a common function.

PR #504, opened 2019-03-04
Add call metadata to events

amitaekbote:
Add call metadata information to events


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
```diff
@@ -608,15 +608,16 @@ public Completable recordTaskPlacement(String taskId, Function<Task, Task> chang
         Task changed = modelUpdateEvent.getChangedEntityHolder().getEntity();
         if (!modelUpdateEvent.getPreviousEntityHolder().isPresent()) {
             return tasksPredicate.test(Pair.of(job, changed))
-                    ? Optional.of(toNewTaskUpdateEvent(job, changed))
+                    ? Optional.of(toNewTaskUpdateEvent(job, changed, modelUpdateEvent.getCallMetadata()))
                     : Optional.empty();
         }
         Task previous = modelUpdateEvent.getPreviousEntityHolder().get().getEntity();
+        CallMetadata callMetadata = (CallMetadata)modelUpdateEvent.getChangedEntityHolder().getAttributes().get(JobManagerConstants.JOB_MANAGER_ATTRIBUTE_CALLMETADATA);
```
tbak: Why the `CallMetadata` value is not taken directly from the modelUpdateEvent like above?

tbak:
After the failover all job and task records will miss the `CallMetadata` attributes. Is there an assumption that this attribute is always present, or it is optional and if missing `undefined` value is used?

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
amitaekbote: I must have changed that before I used modelUpdateEvent, thanks for catching that.

amitaekbote:
I currently assume its always present, I will check in the event getCallMetadata for sanity

PR #508, opened 2019-03-08
add owner passthrough attribute to agent

corindwyer:
### Description of the Change

* Add owner passthrough attribute to agent.
* Fix cluster operations attribute names


gabrielhartmann:
Thanks @corindwyer.

PR #509, opened 2019-03-11
 Add rate per interval disruption budget policy

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/eviction/service/quota/job/UnhealthyTasksLimitTracker.java
```diff
@@ -87,23 +87,8 @@ public EvictionQuota getQuota(Reference reference) {
...
-            return Pair.of(
-                    0,
-                    String.format("too few containers started: desired=%s, started=%s", JobFunctions.getJobDesiredSize(job), started)
-            );
-        }
-
-        // We have enough tasks started. Check now how many are healthy.
         int healthy = 0;
-        Map<String, String> unhealthyTasks = new HashMap<>();
+        Map<String, String> notStartedAndHealthyTasks = new HashMap<>();
```
corindwyer: is this not started and unhealthy tasks or not started and healthy tasks?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/disruptionbudget/RatePerIntervalDisruptionBudgetRate.java
```diff
@@ -0,0 +1,103 @@
...
+
+package com.netflix.titus.api.jobmanager.model.job.disruptionbudget;
+
+import java.time.Duration;
+import java.util.Objects;
+import javax.validation.constraints.Min;
+
+public class RatePerIntervalDisruptionBudgetRate extends DisruptionBudgetRate {
+
+    @Min(1)
```
corindwyer: I think there should be maxes on this.

titus-server-master/src/main/java/com/netflix/titus/master/eviction/service/quota/job/UnhealthyTasksLimitTracker.java
tbak: The meaning is not (started and healthy). I agree the naming is confusing. I will make it more explicit.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/disruptionbudget/RatePerIntervalDisruptionBudgetRate.java
tbak: Max makes sense for interval. For `limitPerInterval` it is ok to have it as high as user wants it.
I will set `intervalMs` max to 1 hour.

PR #510, opened 2019-03-12
Add job type (service or batch) field to passthrough attributes

gabrielhartmann:
### Description of the Change

The cpu prediction model requires the job type (service or batch) be available on the Agent.


titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
```diff
@@ -159,6 +161,7 @@ public DefaultV3TaskInfoFactory(MasterConfiguration masterConfiguration,
 
         // add owner email
         containerInfoBuilder.putPassthroughAttributes(OWNER_EMAIL_ATTRIBUTE, jobDescriptor.getOwner().getTeamEmail());
+        containerInfoBuilder.putPassthroughAttributes(JOB_TYPE_ATTRIBUTE, isServiceJob(jobDescriptor) ? "service" : "batch");
```
gabrielhartmann: I can make these values constants if you like.  Do you have a place for stuff like that?  In this file?

tbak: If you want to do that, I would create yet another helper function in `JobFunctions` which would return job type or alternatively create a `JobType` enum type.

PR #516, opened 2019-03-14
removing titus-ext/elasticsearch module

amit-git:
### Removal of elasticsearch submodule

This change removes a submodule titus-ext/elasticsearch as it contains an elastic search 2.4 publisher with native/transport ES client. The new module titus-supplementary-components/tasks-publisher contains a HTTP based ES 5.+ publisher. It is designed to be run as a separate spring boot service outside of Titus master process. The code change here also contains some refactoring of tasks-publisher module where a bunch of elastic search integration related classes are moved into a separate sub-package.


amit-git:
redoing this one since a lot of code has changed around Tasks publisher.

PR #522, opened 2019-03-18
Add RatePercentagePerInterval rate policy to the job disruption budget

tbak:


tbak:
@fabiokung The only real difference in the derived classes is how to extract an `intervalMs` and an `evictionRate`. There wouldn't be much shared code between `JobPercentagePerHourRelocationRateController` and `JobPercentagePerIntervalRateController`.

PR #527, opened 2019-03-25
Add reason tag to failed IAM validations

andrew-leung:
Adds a reason tag to failed IAM validation counter. The cardinality of the value should be similar to the cardinality of the `iamRoleName` itself since each should fail for a specific reason.

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobIamValidator.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobIamValidator.java
@@ -42,7 +42,7 @@ public class JobIamValidator implements EntityValidator<JobDescriptor> {
     private static final Logger logger = LoggerFactory.getLogger(JobImageValidator.class);
     private static final String METRICS_ROOT = "titus.validation.iam";
     private static final String failedMetricIamTag = "iamrole";
-    private static final String skippedMetricReasonTag = "reason";
+    private static final String metricReasonTag = "reason";
 
     private final JobSecurityValidatorConfiguration configuration;
     private final IamConnector iamConnector;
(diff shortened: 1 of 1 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobIamValidator.java
```diff
@@ -87,7 +87,10 @@ public JobIamValidator(JobSecurityValidatorConfiguration configuration, IamConne
                 // populate the set with a specific error.
                 .thenReturn(Collections.<ValidationError>emptySet())
                 .onErrorResume(throwable -> {
-                    registry.counter(validationFailureId.withTag(failedMetricIamTag, iamRoleName)).increment();
+                    registry.counter(validationFailureId
+                            .withTag(failedMetricIamTag, iamRoleName)
+                            .withTag(metricReasonTag, throwable.getMessage()))
```
tbak: Is it safe to store exception message as a metric tag value?

corindwyer: I prefer to put the Exception class name here generally if the message contains data to reduce the cardinality.

andrew-leung: The Exception class name here will always be an `IamConnectorException`. That class controls the type/size of the message, which is why I used it. If we were really have concerns about it, that class has a `ErrorType` that we could use, though it's less detailed.

tbak: `ErrorType` sounds much better. Details can be written in the log file.

andrew-leung: Will do.

PR #528, opened 2019-03-25
More metrics/logging in the Titus client connectors

tbak:
+ bug fixes

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/replicator/ReplicatedSnapshot.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/replicator/ReplicatedSnapshot.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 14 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/replicator/ReplicatedSnapshot.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.runtime.connector.common.replicator;
+
+public abstract class ReplicatedSnapshot {
+
+    public abstract String toSignatureString();
```
corindwyer: Can you add docs explaining why this is needed?

tbak: I have renamed this to `toSummaryString`, and I have added a short comment explaining what is the purpose of this method.

PR #532, opened 2019-03-29
add a timeout to image sanitization/validation

fabiokung:
Without an explicit timeout, the sanitization chain was failing with the default timeout from `AggregatingValidator` (2s), which does not provide a useful message. It fails with something like:

> Did not observe any item or terminal signal within 2000ms in 'flatMap' (and no fallback has been configured)

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobImageValidatorConfiguration.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobImageValidatorConfiguration.java
@@ -28,6 +28,6 @@ public interface JobImageValidatorConfiguration {
      * Since Image validations are on the job accept path the timeout value is aggressive.
      * This must be smaller than {@link TitusValidatorConfiguration#getTimeoutMs()}.
      */
-    @DefaultValue("1000")
+    @DefaultValue("1200")
     long getJobImageValidationTimeoutMs();
 }
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobImageValidatorConfiguration.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobImageValidatorConfiguration.java
@@ -23,4 +23,11 @@ import com.netflix.archaius.api.annotations.DefaultValue;
 public interface JobImageValidatorConfiguration {
     @DefaultValue("true")
     boolean isEnabled();
+
+    /**
+     * Since Image validations are on the job accept path the timeout value is aggressive.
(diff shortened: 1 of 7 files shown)


andrew-leung:
Ah, looks like this got lost in the shuffle going from RxNetty for WebClient. The registry connector used to configure the timeout on the RxNetty client (https://github.com/Netflix/titus-control-plane/commit/dcda444fec1f5ee7eab33f66452a6c3c68925d2b#diff-402c2e1f0eb455ee458f267014e0b350L73). The registry connector then has a large safety net timeout, though looking at it, it's really too large right now to be useful. It probably makes sense to also have a timeout at the validator layer.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobImageValidatorConfiguration.java
```diff
@@ -23,4 +23,11 @@
 public interface JobImageValidatorConfiguration {
     @DefaultValue("true")
     boolean isEnabled();
+
+    /**
+     * Since Image validations are on the job accept path the timeout value is aggressive.
+     * This must be smaller than {@link TitusValidatorConfiguration#getTimeoutMs()}.
+     */
+    @DefaultValue("1000")
```
andrew-leung: It might also be worth adjusting the retry count in `TitusRegistryClientConfiguration` because this would timeout while the `DefaultDockerRegistryClient` is still still retrying. Maybe 2 500ms retries at the registry client and a 1500ms timeout here?

fabiokung: That could potentially only leave 500ms for the other sanitizers (Iam) to run. How about 2x 400ms? Is there anything experimental about the 500ms timeout?

andrew-leung: We've been using the 500ms for sometime now, so the 400ms would be new. I think we lost latency metrics though going Rx->Spring :-( FWIW, the IAM sanitization should be fast as the lookup should effectively always be cached since validation run just prior to sanitization adds it to the cache. But I'm OK using either 400 or 500ms.

fabiokung: alright, I'll change this to 1500ms and tweak the `TitusRegistryClientConfiguration`

fabiokung:
@andrew-leung I limited the `RegistryClient` to 2 retries, with a 500ms timeout and 5ms delay in between, and aligned the Sanitizer to 1200ms total timeout, which should be sufficient buffer.

PR #533, opened 2019-03-30
AlarmConfiguration#autoscalingGroupName is not being used

fabiokung:
Remove dead code.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/appscale/store/mixin/AlarmConfigurationMixIn.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/appscale/store/mixin/AlarmConfigurationMixIn.java
@@ -31,7 +31,6 @@ public abstract class AlarmConfigurationMixIn {
             @JsonProperty("region") String region,
             @JsonProperty("actionsEnabled") Optional<Boolean> actionsEnabled,
             @JsonProperty("comparisonOperator") ComparisonOperator comparisonOperator,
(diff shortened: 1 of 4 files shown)


titus-api/src/main/java/com/netflix/titus/api/appscale/store/mixin/AlarmConfigurationMixIn.java
```diff
@@ -31,7 +31,6 @@
             @JsonProperty("region") String region,
             @JsonProperty("actionsEnabled") Optional<Boolean> actionsEnabled,
             @JsonProperty("comparisonOperator") ComparisonOperator comparisonOperator,
-            @JsonProperty("autoScalingGroupName") String autoScalingGroupName,
```
tbak: Removing this attribute may prevent loading of the existing records from the database. We need to provide a hint to  Jackson deserializer to ignore this value if present.

corindwyer: I believe it is the `@JsonIgnoreProperties(ignoreUnknown = true)` annotation.

fabiokung: good catch, I thought this was always the default

fabiokung: Done. PTAL

PR #535, opened 2019-04-02
Add latency metric for regitry WebClient

andrew-leung:
- Moves WebClient Metrics to its own metrics-specific class.
- Records latency docker registry client requests.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/network/client/internal/WebClientMetric.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/network/client/internal/WebClientMetric.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


titus-common/src/main/java/com/netflix/titus/common/network/client/internal/WebClientMetric.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public void registerLatency(HttpMethod method, long startTimeMs) {
+        registry.timer(latencyId
+                .withTag(WEB_CLIENT_METHOD_TAG, method.name()))
+                .record(System.currentTimeMillis() - startTimeMs, TimeUnit.MILLISECONDS);
+    }
+
+    // Returns a function to increment on success
+    public BiConsumer<? super HttpClientResponse, ? super Connection> getIncrementOnSuccess() {
```
fabiokung: minor nit: the API here looks a bit odd with these methods being factories for functions that do the actual action. I'd recommend making these do the increment directly, and just wrapping them with a lambda in `TitusWebClientAddOns.addMetricCallbacks`

andrew-leung: Will do.

PR #536, opened 2019-04-02
Store caller id as a job attribute when the job is created

tbak:


titus-api/src/main/java/com/netflix/titus/api/jobmanager/JobAttributes.java
```diff
@@ -27,6 +27,11 @@
 
     // Job Descriptor Attributes
 
+    /**
+     * Job creator.
+     */
+    public static final String JOB_ATTRIBUTES_CALLER_ID = TITUS_ATTRIBUTE_PREFIX + "caller.id";
```
fabiokung: should the attributed by called `createdBy`?

fabiokung: expanding: `caller.id` could mean any operation, not necessarily who submitted the job

tbak: You are right. I will change this.

PR #541, opened 2019-04-05
Add common metrics to validators and cleanup

andrew-leung:
This PR makes the following changes:

- Adds a common validator metrics class and integrates it with each validator.
- Moves the "hard" and "soft" attribute to be a property of the validators themselves rather than the aggregating validator.
- Adds validator timeouts where missing.
- Makes IamConnectorMetrics more general so internal connector implementations can use them.


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobImageValidatorConfiguration.java
```diff
@@ -21,6 +21,9 @@
 
 @Configuration(prefix = "titus.validate.job.image")
 public interface JobImageValidatorConfiguration {
+    @DefaultValue("SOFT")
```
tbak: This interface looks similar for all validators (has `getErrorType` and `isEnabled`). These operational configuration could be captured via common interface.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/validator/FailJobValidator.java
```diff
@@ -46,4 +56,9 @@
     public Mono<JobDescriptor> sanitize(JobDescriptor entity) {
         return Mono.error(TitusServiceException.invalidArgument(ERR_DESCRIPTION));
     }
+
+    @Override
+    public ValidationError.Type getErrorType() {
```
tbak: This implementation is identical in all cases, and could be moved to an abstract class. This will also make it easier to add new methods in the future if needed.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/validator/JobImageValidator.java
```diff
@@ -144,4 +143,30 @@ private boolean hasDigest(Image image) {
...
+        logger.error("Exception while checking image digest", throwable);
+        if (throwable instanceof TitusRegistryException) {
+            TitusRegistryException tre = (TitusRegistryException) throwable;
+            // Use a more specific error tag if available
+            validatorMetrics.incrementValidationError(
+                    imageName,
+                    tre.getErrorCode().name()
+            );
+            // We are ignoring most image validation errors. We will propagate
+            // more errors as we going feature confidence.
```
corindwyer: typo?

andrew-leung: Indeed.

PR #531, opened 2019-03-28
Job manager client connector layer re-write

tbak:

PR #542, opened 2019-04-09
testkit/perf: use the new JobManagement connector

fabiokung:
Follow-up of the refactoring made in #531, which makes it easier to use `LoadGeneratorComponent` in SpringBoot applications.


titus-testkit/src/main/java/com/netflix/titus/testkit/perf/load/plan/generator/ConstantLoadJobExecutableGenerator.java
```diff
@@ -16,42 +16,35 @@
...
     private final Executable executable;
-    private final int numberOfJobs;
-
-    private volatile Subscriber<? super Executable> scenarioSubscriber;
+    private final FluxProcessor<Executable, Executable> executionPlans;
 
     public ConstantLoadJobExecutableGenerator(String owner, JobDescriptor<?> jobSpec, ExecutionPlan plan, int numberOfJobs) {
         this.executable = new Executable(owner, jobSpec, plan);
-        this.numberOfJobs = numberOfJobs;
+        this.executionPlans = UnicastProcessor.<Executable>create().serialize();
```
tbak: Will this work for multiple subscriptions?

fabiokung: the original code was trying to force a single subscription

PR #544, opened 2019-04-11
basic circle ci config to run all tests

fabiokung:
Travis is being a bit too slow when running integration tests. This is an experiment in using CircleCI to see if it does any better.

I left tests running in both Travis and Circle so we can compare after a while.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,7 @@ jobs:
     executor: java
     resource_class: large
     steps:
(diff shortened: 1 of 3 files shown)


fabiokung:
5min on Circle (https://circleci.com/gh/Netflix/titus-control-plane/13) vs 20min on Travis (https://travis-ci.org/Netflix/titus-control-plane/jobs/519037609) 🤔

PR #552, opened 2019-04-22
Support scaling phased out instance groups

corindwyer:
### Description of the Change
Scale phased out instance groups and prefer Active for scale ups and PhasedOut for scale downs.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
@@ -103,8 +103,8 @@ public class ClusterAgentAutoScaler {
     private static final long SCALE_DOWN_TOKEN_BUCKET_REFILL_AMOUNT = 2;
     private static final long SCALE_DOWN_TOKEN_BUCKET_REFILL_INTERVAL_MS = 1_000;
 
-    private static final Comparator<Map.Entry<AgentInstanceGroup, List<AgentInstance>>> IDLE_INSTANCE_COMPARATOR = Comparator.comparing(e -> e.getKey().getLifecycleStatus().getState(), Comparator.reverseOrder());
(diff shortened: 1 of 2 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
```diff
@@ -360,13 +368,15 @@ Completable doAgentScaling() {
...
                                                         long elapsed) {
+        // use reverse order of instance group state such that PhasedOut instances are scaled down first.
         return instancesForActiveInstanceGroups.entrySet().stream()
                 .filter(e -> {
                     AgentInstanceGroup instanceGroup = e.getKey();
                     return instanceGroup.getTier() == tier &&
                             instanceGroup.getInstanceType().equals(primaryInstanceType) &&
                             !instanceGroup.getAttributes().containsKey(ClusterOperationsAttributes.NOT_REMOVABLE);
                 })
+                .sorted(IDLE_INSTANCE_COMPARATOR)
```
fabiokung: if my read of `createSetRemovableOverrideStatusesCompletable()` is correct, it doesn't seem this order will influence which instance groups get picked to be scaled down first. The order for scaling down seems to be also controlled by `getActiveInstanceGroups()`.

fabiokung: more specifically, here: https://github.com/corindwyer/titus-control-plane/blob/scale-phased-out-agents/titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java#L434

which gets called with the list of instance groups being `activeInstanceGroupsForTier`, here: https://github.com/corindwyer/titus-control-plane/blob/scale-phased-out-agents/titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java#L302

fabiokung:
some unit tests for the behavior here would be nice

PR #554, opened 2019-04-22
Enable task relocation for containers in  Eureka/OUT_OF_SERVICE state

tbak:
If a container is in OUT_OF_SERVICE state in Eureka, its real state is lost. This forces us two chose between two strategies:
* assume containers are down, and report them as unhealthy
* assume containers are up, and report them as healthy
The former strategy was assumed before this change. This resulted in large pools of resources to be excluded from the relocation process. We switch to the latter strategy, accepting that this may increase the risk for service availability during rollbacks.

### DIFF
--- a/titus-ext/eureka/src/main/java/com/netflix/titus/ext/eureka/containerhealth/EurekaContainerHealthService.java
+++ b/titus-ext/eureka/src/main/java/com/netflix/titus/ext/eureka/containerhealth/EurekaContainerHealthService.java
@@ -37,14 +37,18 @@ import com.netflix.titus.api.containerhealth.model.ContainerHealthStatus;
 import com.netflix.titus.api.containerhealth.model.event.ContainerHealthEvent;
 import com.netflix.titus.api.containerhealth.model.event.ContainerHealthUpdateEvent;
 import com.netflix.titus.api.containerhealth.service.ContainerHealthService;
+import com.netflix.titus.api.jobmanager.model.job.Job;
+import com.netflix.titus.api.jobmanager.model.job.JobFunctions;
 import com.netflix.titus.api.jobmanager.model.job.Task;
 import com.netflix.titus.api.jobmanager.model.job.TaskState;
 import com.netflix.titus.api.jobmanager.model.job.event.JobManagerEvent;
+import com.netflix.titus.api.jobmanager.model.job.event.JobUpdateEvent;
 import com.netflix.titus.api.jobmanager.model.job.event.TaskUpdateEvent;
 import com.netflix.titus.api.jobmanager.service.ReadOnlyJobOperations;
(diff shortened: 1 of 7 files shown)


titus-ext/eureka/src/main/java/com/netflix/titus/ext/eureka/containerhealth/EurekaContainerHealthService.java
```diff
@@ -115,23 +119,28 @@ private ContainerHealthStatus buildHealthStatus(Task task, ContainerHealthState
...
+    private ContainerHealthState takeStateOf(Job<?> job, Task task) {
         // If it is finished, ignore Eureka status
         if (task.getStatus().getState() == TaskState.Finished) {
             return ContainerHealthState.Terminated;
         }
 
+        List<InstanceInfo> instances = eurekaClient.getInstancesById(task.getId());
+
+        // If a job is disabled, the real Eureka state is hidden. If the container is not registered with Eureka in
+        // the disabled job, we also do not put any constrains here. In both cases we report it is healthy.
```
corindwyer: constrains -> constraints

titus-ext/eureka/src/main/java/com/netflix/titus/ext/eureka/containerhealth/EurekaContainerHealthService.java
```diff
@@ -194,13 +216,32 @@ private String takeStateReasonOf(Task task) {
...
     }
 
-    private Optional<ContainerHealthEvent> handleTaskStateUpdate(Task task, ConcurrentMap<String, ContainerHealthEvent> state) {
+    private Flux<ContainerHealthEvent> handleJobEnabledStatusUpdate(Job current, Job previous, ConcurrentMap<String, ContainerHealthEvent> state) {
+
+        if (!JobFunctions.isServiceJob(current)) {
+            return Flux.empty();
+        }
+
+        // We need to examine in if a job 'enabled' status was changed.
```
corindwyer: `We need to examine in if a job 'enabled' status was changed.` -> `Examine if a job's 'enabled' status was changed.`

titus-ext/eureka/src/main/java/com/netflix/titus/ext/eureka/containerhealth/EurekaContainerHealthService.java
tbak: Fixed

titus-ext/eureka/src/main/java/com/netflix/titus/ext/eureka/containerhealth/EurekaContainerHealthService.java
tbak: Fixed

PR #559, opened 2019-05-02
using caffeine cache for jobs in Titus client

amit-git:
### Fixing memory leak in tasks publisher

Memory is leaking in tasks publisher because of the fact that a map used to cache jobs metadata is unbounded. This change is going to use Caffeine cache with a max size of 40k that'll put an upper bound on heap growth. It loads the cache using async calls to Titus API. This change doesn't affect the way task events generator is using the client.


titus-supplementary-component/tasks-publisher/src/main/java/com/netflix/titus/supplementary/taskspublisher/TitusClientImpl.java
```diff
@@ -182,4 +169,25 @@ private void configureMetrics(Registry registry) {
...
+    private void buildCacheForJobs() {
+        jobs = Caffeine.newBuilder()
+                .maximumSize(MAX_CACHE_SIZE)
+                .buildAsync(new AsyncCacheLoader<String, Job>() {
+                    @Nonnull
+                    @Override
+                    public CompletableFuture<Job> asyncLoad(@Nonnull String jobId, @Nonnull Executor executor) {
+                        CompletableFuture<Job> jobResult = new CompletableFuture<>();
+                        ListenableFuture<Job> jobFuture = jobManagementServiceFutureStub.findJob(JobId.newBuilder().setId(jobId).build());
+                        jobFuture.addListener(() -> {
```
fabiokung: probably need to do more here, e.g.: https://github.com/spotify/futures-extra/blob/master/src/main/java/com/spotify/futures/ListenableToCompletableFutureWrapper.java

titus-supplementary-component/tasks-publisher/src/main/java/com/netflix/titus/supplementary/taskspublisher/TitusClientImpl.java
```diff
@@ -42,19 +49,27 @@
 public class TitusClientImpl implements TitusClient {
     private static final Logger logger = LoggerFactory.getLogger(TitusClientImpl.class);
     private static final String CLIENT_ID = "tasksPublisher";
-    private final JobManagementServiceGrpc.JobManagementServiceStub jobManagementService;
+    public static final int MAX_CACHE_SIZE = 40000;
+    private final JobManagementServiceStub jobManagementService;
+    private JobManagementServiceFutureStub jobManagementServiceFutureStub;
```
fabiokung: any reason all these attributes can't be `final`? Not a big deal, but it would make it easier to maintain this code in the future.

titus-supplementary-component/tasks-publisher/src/main/java/com/netflix/titus/supplementary/taskspublisher/TitusClientImpl.java
corindwyer: I like to use `Futures.addCallback` in guava for this.

titus-supplementary-component/tasks-publisher/src/main/java/com/netflix/titus/supplementary/taskspublisher/TitusClientImpl.java
amit-git: Done

PR #562, opened 2019-05-08
Add metrics for remaining relocations in the TaskRelocation service

tbak:


titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/util/RelocationPredicates.java
```diff
@@ -123,6 +138,10 @@ public static boolean isRelocationRequired(AgentInstance agentInstance) {
                 .equalsIgnoreCase("true");
     }
 
+    public static boolean isRelocationRequired(AgentInstanceGroup instanceGroup) {
+        return instanceGroup.getLifecycleStatus().getState() == InstanceGroupLifecycleState.Removable;
```
corindwyer: shouldn't this also check for the `RELOCATION_REQUIRED` attribute?

tbak: Is it supported on instance group? We make this check on instance.

PR #563, opened 2019-05-08
removing elasticsearch module from titus-ext

amit-git:
### Removing titus-ext module for elasticsearch publishing

The code being removed is deprecated after a separate Titus supplementary module 'tasks-publisher' was created to support Titus tasks publishing mechanism. As a result, the TaskDocument class is being moved into tasks-publisher module.


titus-supplementary-component/tasks-publisher/src/main/java/com/netflix/titus/supplementary/taskspublisher/es/DefaultEsWebClientFactory.java
```diff
@@ -56,7 +54,6 @@ private HttpClient buildHttpClient() {
             TcpClient tcpClientWithConnectionTimeout = tcpClient.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, 1000);
             return tcpClientWithConnectionTimeout.doOnConnected(connection -> {
                 connection.addHandlerLast(new ReadTimeoutHandler(READ_TIMEOUT_SECONDS));
-                connection.addHandlerLast(new WriteTimeoutHandler(WRITE_TIMEOUT_SECONDS));
```
corindwyer: can you add a comment to look at this into the future? I am concerned about the behavior when not timing out client side requests.

PR #567, opened 2019-05-14
Add experimental constraints

corindwyer:
### Description of the Change

Add experimental constraints to prefer placement based on:
* availability zone
* machine id
* machine type

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/scheduler/SchedulerConfiguration.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/scheduler/SchedulerConfiguration.java
@@ -18,6 +18,7 @@ package com.netflix.titus.master.scheduler;
 
 import com.netflix.archaius.api.annotations.Configuration;
 import com.netflix.archaius.api.annotations.DefaultValue;
+import com.netflix.titus.api.agent.model.AgentInstance;
 import com.netflix.titus.master.scheduler.fitness.networkinterface.TitusNetworkInterfaceFitnessEvaluator;
 
 @Configuration(prefix = "titus.scheduler")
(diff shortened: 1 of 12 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/scheduler/SchedulerConfiguration.java
```diff
@@ -114,4 +115,16 @@
      */
     @DefaultValue("true")
     boolean isCriticalTierJobSpreadingEnabled();
+
+    /**
+     * Availability zone attribute name to look at on the {@link AgentInstance}.
+     */
+    @DefaultValue("availabilityZone")
+    String getAvailabilityZoneAttributeName();
```
tbak: We should use standard names internally and map if needed at the connector layer.
This introduces unnecessary complexity.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/MachineTypeConstraint.java
```diff
@@ -0,0 +1,67 @@
...
+        this.machineType = machineType;
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public Result evaluate(TaskRequest taskRequest, VirtualMachineCurrentState targetVM, TaskTrackerState taskTrackerState) {
```
tbak: There is a lot of common code in these three constraints.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/SchedulerConfiguration.java
corindwyer: This is based on how the current model works. I agree that we should change the approach but I do not want to do it in this PR.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/MachineTypeConstraint.java
corindwyer: I prefer to keep these classes orthogonal at the time because they are experimental. When they are no longer experimental, inheritance or composition could be used here to reduce the common code.

PR #568, opened 2019-05-14
Fix memory leak in data replicator Rx channel

tbak:


titus-common/src/test/java/com/netflix/titus/common/util/rx/ReactorReEmitterOperatorTest.java
```diff
@@ -17,69 +17,75 @@
...
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 public class ReactorReEmitterOperatorTest {
 
-    private static final long INTERVAL_MS = 1_000;
+    private static final Duration EMIT_INTERVAL = Duration.ofMillis(1_000);
 
-    private final DirectProcessor<String> subject = DirectProcessor.create();
+    private boolean cancelled;
```
fabiokung: given we are using `StepVerifier` and `Schedulers.parallel()`, it would probably be safer to make this `volatile` to avoid flakiness.

tbak: `StepVerifier` swaps `Schedulers.parallel()` with test scheduler, which makes it effectively run on the main thread.

PR #570, opened 2019-05-20
Add support for configuring shared memory size

andrew-leung:
Adds the ability to configure the size of /dev/shm shared memory.

If no shm value is provided, a default value is used.
If the value provided is larger than the memory, the job is rejected.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ContainerResources.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ContainerResources.java
@@ -17,11 +17,13 @@
 package com.netflix.titus.api.jobmanager.model.job;
 
 import java.util.List;
+import java.util.Objects;
 import javax.validation.Valid;
 import javax.validation.constraints.Min;
 
 import com.netflix.titus.api.jobmanager.model.job.sanitizer.EfsMountsSanitizer;
 import com.netflix.titus.api.model.EfsMount;
+import com.netflix.titus.common.model.sanitizer.ClassInvariant;
 import com.netflix.titus.common.model.sanitizer.FieldInvariant;
 import com.netflix.titus.common.model.sanitizer.FieldSanitizer;
 import com.netflix.titus.common.util.CollectionsExt;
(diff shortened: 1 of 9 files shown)


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ContainerResources.java
```diff
@@ -103,43 +122,20 @@ public boolean equals(Object o) {
...
-        }
-        return efsMounts != null ? efsMounts.equals(that.efsMounts) : that.efsMounts == null;
+        return Double.compare(that.cpu, cpu) == 0 &&
+                gpu == that.gpu &&
+                memoryMB == that.memoryMB &&
+                diskMB == that.diskMB &&
+                networkMbps == that.networkMbps &&
+                allocateIP == that.allocateIP &&
+                shmMB == that.shmMB &&
+                efsMounts.equals(that.efsMounts);
```
rgulewich: This leaves out the `efsMounts != null` guard in the previous code - not sure if that's OK here or not.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ContainerResources.java
```diff
@@ -171,7 +168,8 @@ public static Builder newBuilder(ContainerResources containerResources) {
                 .withDiskMB(containerResources.getDiskMB())
                 .withNetworkMbps(containerResources.getNetworkMbps())
                 .withAllocateIP(containerResources.isAllocateIP())
-                .withEfsMounts(containerResources.getEfsMounts());
+                .withEfsMounts(containerResources.getEfsMounts())
+                .witShmMB(containerResources.getShmMB());
```
rgulewich: Should this be `.with`, instead of `.wit` (an extra "h")?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ContainerResources.java
```diff
@@ -221,12 +220,24 @@ public Builder withAllocateIP(boolean allocateIP) {
             return this;
         }
 
+        public Builder witShmMB(int shmMB) {
```
rgulewich: See comment above.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ContainerResources.java
andrew-leung: This was just the intelliJ regenerate to hash/equals. Maybe I chose the wrong format. Will check.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ContainerResources.java
andrew-leung: Good catch!

PR #571, opened 2019-05-28
Move TierSlaUpdated and AvailableCapacityService to its own thread

amitaekbote:


titus-server-master/src/main/java/com/netflix/titus/master/scheduler/DefaultTierSlaUpdater.java
```diff
@@ -53,7 +54,7 @@
     public DefaultTierSlaUpdater(SchedulerConfiguration schedulerConfiguration,
                                  ApplicationSlaManagementService applicationSlaManagementService,
                                  AvailableCapacityService availableCapacityService) {
-        this(schedulerConfiguration, applicationSlaManagementService, availableCapacityService, Schedulers.io());
+        this(schedulerConfiguration, applicationSlaManagementService, availableCapacityService, SchedulerExt.createSingleThreadScheduler("tier-sla-updater"));
```
tbak: We need shutdown method here to destroy the scheduler.

titus-server-master/src/main/java/com/netflix/titus/master/service/management/internal/DefaultAvailableCapacityService.java
```diff
@@ -67,7 +68,7 @@ public DefaultAvailableCapacityService(CapacityManagementConfiguration configura
                                            ServerInfoResolver serverInfoResolver,
                                            AgentManagementService agentManagementService,
                                            Registry registry) {
-        this(configuration, serverInfoResolver, agentManagementService, registry, Schedulers.io());
+        this(configuration, serverInfoResolver, agentManagementService, registry, SchedulerExt.createSingleThreadScheduler("available-capacity-service"));
```
tbak: We need shutdown method here to destroy the scheduler.

PR #574, opened 2019-05-30
Add task activity to job activity api 

amitaekbote:


titus-grpc-api/src/main/proto/netflix/titus/job_activity_history.proto
```diff
@@ -29,20 +29,42 @@ message JobActivityRecord {
...
 }
 
+message TaskActivityQueryResult {
+    repeated TaskActivityRecord records = 1;
+}
+
 service JobActivityService {
 
-    /// Get all known TitusMaster instances.
+    /// Get all known active job records.
```
tbak: What about point queries (by jobId or taskId)?

PR #575, opened 2019-06-03
Extend the 'MasterInstance' supervisor model

tbak:


titus-common/src/main/java/com/netflix/titus/common/util/StringExt.java
```diff
@@ -447,6 +447,21 @@ public static String doubleQuotes(String text) {
         return "\"" + text + "\"";
     }
 
+    public static String removeQuotes(String text) {
```
corindwyer: Comment on what it does

tbak: Done

PR #577, opened 2019-06-04
Minor refactor to support IAM Role authorization

andrew-leung:
This is change supports changes to nftitus. There are 2 main changes:
1. Adds common authorization metrics
2. Moves all feature predicate checking to the sub-class implementing `protected abstract Mono<AuthorizationStatus> authorize()`. Note that because this moves the predicate check it depends on the companion PR to properly scope which apps to check.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/spectator/AuthorizationMetrics.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/spectator/AuthorizationMetrics.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.common.util.spectator;
+
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
+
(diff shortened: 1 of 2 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/JobAuthorizationService.java
```diff
@@ -75,10 +68,6 @@ protected JobAuthorizationService(Predicate<String> callerIdPredicate) {
     }
 
     private Mono<AuthorizationStatus> authorizeCaller(JobDescriptor<?> jobDescriptor, String originalCallerId) {
-        if (!callerIdPredicate.test(originalCallerId)) {
```
tbak: Does it mean, we do not support user-level security domains?

titus-common/src/main/java/com/netflix/titus/common/util/spectator/AuthorizationMetrics.java
```diff
@@ -0,0 +1,53 @@
...
+    private static final String AUTHORIZATION_RESOURCE_TAG = "resource";
+    private static final String AUTHORIZATION_DOMAIN_TAG = "domain";
+    private static final String AUTHORIZATION_ERROR_TAG = "error";
+
+    private final Registry registry;
+    private final Id authorizationId;
+
+    public AuthorizationMetrics(String authorizationName, Registry registry) {
+        this.registry = registry;
+        authorizationId = registry.createId(AUTHORIZATION_METRICS_ROOT + authorizationName);
```
tbak: What is authorization name?

andrew-leung: That would be the Authorization implementation class name (e.g., GandalfJobAuth).

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/JobAuthorizationService.java
andrew-leung: We do but the check is made in the implementing class.

PR #585, opened 2019-06-20
track Jobs with sanitization (open) failures, and relax moveTask validation for them

fabiokung:
When a job is cloned and either one of the copies had sanitization failures, they can have different values for the non-sanitized field(s), but are still compatible.

This PR also ignores all user-supplied `titus.sanitization.*` attributes for new jobs, since these are reserved and automatically filled by Titus.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/CollectionsExt.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/CollectionsExt.java
@@ -223,6 +223,12 @@ public final class CollectionsExt {
         return result;
     }
 
+    public static <K, V> Map<K, V> copyAndRemoveByKey(Map<K, V> original, Predicate<K> keysPredicate) {
+        return original.entrySet().stream()
+                .filter(entry -> !keysPredicate.test(entry.getKey()))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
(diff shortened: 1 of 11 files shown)


titus-common/src/main/java/com/netflix/titus/common/util/CollectionsExt.java
```diff
@@ -223,6 +223,12 @@ private CollectionsExt() {
         return result;
     }
 
+    public static <K, V> Map<K, V> copyAndRemoveByKey(Map<K, V> original, Predicate<K> keysPredicate) {
+        return original.entrySet().stream()
```
tbak: Stream implementation has poor performance, which may matter in some cases. Direct for loop would do much better here.

fabiokung: I agree there's a performance hit, but have you seen it being the actual bottleneck in practice for the control plane? We already use the stream api on the similar `copyAndRemoveByValue`, and since this is only being used during job submissions, I'd say avoid the early optimization for now.

tbak: Yes, we have seen this in Fenzo. As a general rule we avoided streams API in all functions called by Fenzo.

PR #586, opened 2019-06-21
Update kill message with caller id

amitaekbote:
### Description of the Change

Update the kill message to include the caller id.


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
```diff
@@ -434,7 +435,8 @@ public Completable recordTaskPlacement(String taskId, Function<Task, Task> chang
                         }
                         reasonCode = TaskStatus.REASON_SCALED_DOWN;
                     }
-                    String reason = String.format("%s (shrink=%s)", Evaluators.getOrDefault(callMetadata.getCallReason(), "<no_reason>"), shrink);
+                    String reason = String.format("%s %s(shrink=%s)", Evaluators.getOrDefault(callMetadata.getCallReason(), "<no_reason>"),
```
corindwyer: I think the callerId should be first instead of second.

amitaekbote: done

PR #594, opened 2019-07-04
Make distinction clear between validation for POJOs and API admission control

fabiokung:
Rename classes and packages to make it clear where each type of validation should be used.

I will separate validation from sanitization onto different interfaces for admission control in a separate PR.


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/PassJobValidator.java
```diff
@@ -14,27 +14,26 @@
...
  * should be overridden.
  */
 @Singleton
-public class PassJobValidator implements EntityValidator<JobDescriptor> {
+public class PassJobValidator implements AdmissionValidator<JobDescriptor> {
     private final ValidationError.Type errorType;
 
     public PassJobValidator() {
-        this(ValidationError.Type.HARD);
+        this(ValidationError.Type.SOFT);
```
tbak: Does it matter if it is hard or soft here?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/AdmissionValidator.java
```diff
@@ -14,19 +14,20 @@
...
+package com.netflix.titus.runtime.endpoint.admission;
 
 import java.util.Set;
 
+import com.netflix.titus.common.model.sanitizer.ValidationError;
 import reactor.core.publisher.Mono;
 
 /**
- * A EntityValidator determines whether an object of the parameterized type is valid.  If it finds an object to be invalid it
+ * A AdmissionValidator determines whether an object of the parameterized type is valid.  If it finds an object to be invalid it
```
corindwyer: A -> An

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/PassJobValidator.java
fabiokung: I don't think so since this particular implementation never fails validations.

PR #595, opened 2019-07-08
admission: split sanitization and validation into different interfaces

fabiokung:
This prepares the ground for more advanced sanitization logic, and dedicated sanitizers that are not validators (and vice versa).

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/AdmissionSanitizer.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/AdmissionSanitizer.java
@@ -22,8 +22,8 @@ public interface AdmissionSanitizer<T> {
     /**
      * Async method to clean and add missing data elements to an entity
      *
-     * @return Returns a Mono/Single that emits an a cleaned up version. If no changes were made,
-     * the emitted item will share reference with the source parameter.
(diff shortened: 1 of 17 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/AdmissionSanitizer.java
```diff
@@ -0,0 +1,29 @@
...
+
+package com.netflix.titus.runtime.endpoint.admission;
+
+import reactor.core.publisher.Mono;
+
+public interface AdmissionSanitizer<T> {
+    /**
+     * Async method to clean and add missing data elements to an entity
+     *
+     * @return Returns a Mono/Single that emits an a cleaned up version. If no changes were made,
```
corindwyer: delete `an`

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/AdmissionSanitizer.java
```diff
@@ -0,0 +1,29 @@
...
+     * the emitted item will share reference with the source parameter.
```
corindwyer: `share reference` -> `share a reference`

PR #601, opened 2019-07-15
only log a stacktrace when debug is enabled

fabiokung:
... when an error is expected and won't cause any harm.

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/replicator/AbstractReplicatorEventStream.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/replicator/AbstractReplicatorEventStream.java
@@ -54,11 +54,8 @@ public abstract class AbstractReplicatorEventStream<SNAPSHOT, TRIGGER> implement
                 })
                 .doOnCancel(metrics::disconnected)
                 .doOnError(error -> {
-                    if (logger.isDebugEnabled()) {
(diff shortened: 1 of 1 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/replicator/AbstractReplicatorEventStream.java
```diff
@@ -54,7 +54,11 @@ protected AbstractReplicatorEventStream(DataReplicatorMetrics metrics, TitusRunt
                 })
                 .doOnCancel(metrics::disconnected)
                 .doOnError(error -> {
-                    logger.warn("[{}] Connection to the event stream terminated with an error: {}", getClass().getSimpleName(), error.getMessage(), error);
+                    if (logger.isDebugEnabled()) {
+                        logger.debug("[{}] Connection to the event stream terminated with an error: {}", getClass().getSimpleName(), error.getMessage(), error);
+                    } else {
+                        logger.warn("[{}] Connection to the event stream terminated with an error: {}", getClass().getSimpleName(), error.getMessage());
```
tbak: I prefer to always log at `warn` level, and have an additional line for debug. Otherwise `grep WARN ... ` would miss this log line if debug level is set.

fabiokung: sure, will change

PR #602, opened 2019-07-16
 Implement Reactor API/GRPC server side binding

tbak:


titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/ServerStreamingMethodHandler.java
```diff
@@ -0,0 +1,67 @@
...
+                        logger.warn("Subscriber threw error in onError handler", ex);
+                    }
+                },
+                () -> {
+                    try {
+                        responseObserver.onCompleted();
+                    } catch (Exception ex) {
+                        logger.warn("Subscriber threw error in onCompleted handler. Retrying with onError", ex);
+                        try {
+                            responseObserver.onError(ex);
```
fabiokung: I think this breaks the `StreamObserver` contract:

```
  /**
   * Receives a notification of successful stream completion.
   *
   * <p>May only be called once and if called it must be the last method called. In particular if an
   * exception is thrown by an implementation of {@code onCompleted} no further calls to any method
   * are allowed.
   */
  void onCompleted();
```

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/ServerStreamingMethodHandler.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public void invoke(REQ request, StreamObserver<RESP> responseObserver) {
+        super.invoke(request, responseObserver);
+    }
+
+    @Override
+    Disposable handleResult(Publisher<RESP> result, StreamObserver<RESP> responseObserver) {
+        return ((Flux<RESP>) result).subscribe(
+                responseObserver::onNext,
```
fabiokung: we need to handle exceptions here and terminate the stream calling `onError()`. From the `StreamObserver` contract:

```
  /**
   * Receives a value from the stream.
   *
   * <p>Can be called many times but is never called after {@link #onError(Throwable)} or {@link
   * #onCompleted()} are called.
   *
   * <p>Unary calls must invoke onNext at most once.  Clients may invoke onNext at most once for
   * server streaming calls, but may receive many onNext callbacks.  Servers may invoke onNext at
   * most once for client streaming calls, but may receive many onNext callbacks.
   *
   * <p>If an exception is thrown by an implementation the caller is expected to terminate the
   * stream by calling {@link #onError(Throwable)} with the caught exception prior to
   * propagating it.
   *
   * @param value the value passed to the stream
   */
  void onNext(V value);
```

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/ServerStreamingMethodHandler.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public void invoke(REQ request, StreamObserver<RESP> responseObserver) {
+        super.invoke(request, responseObserver);
+    }
+
+    @Override
+    Disposable handleResult(Publisher<RESP> result, StreamObserver<RESP> responseObserver) {
+        return ((Flux<RESP>) result).subscribe(
```
fabiokung: `Flux.from(result)` is better than the explicit cast here

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/UnaryMethodHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                    },
+                    () -> {
+                        try {
+                            // Void must be mapped to GRPC/Empty value.
+                            responseObserver.onNext((RESP) Empty.getDefaultInstance());
+                            responseObserver.onCompleted();
+                        } catch (Exception ex) {
+                            logger.warn("Subscriber threw error in onCompleted handler. Retrying with onError", ex);
+                            try {
+                                responseObserver.onError(ex);
```
fabiokung: same comment about breaking the `StreamObserver` contract here

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/UnaryMethodHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                            } catch (Exception ignore) {
+                            }
+                        }
+                    }
+            );
+        } else {
+            disposable = monoResult.subscribe(
+                    responseObserver::onNext,
```
fabiokung: similar comments from `ServerStreamingMethodHandler`

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/GrpcToReactorServerFactoryComponent.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.netflix.titus.runtime.connector.common.reactor;
+
+import com.netflix.archaius.api.annotations.Configuration;
+import com.netflix.titus.runtime.connector.common.reactor.server.DefaultGrpcToReactorServerFactory;
+import com.netflix.titus.runtime.endpoint.metadata.CallMetadataResolver;
+import org.springframework.context.annotation.Bean;
+
+@Configuration
```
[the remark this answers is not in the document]
tbak: Good catch.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/ServerStreamingMethodHandler.java
tbak: I will keep the warn, but remove the subsequent onError.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/ServerStreamingMethodHandler.java
tbak: Fixed

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/ServerStreamingMethodHandler.java
tbak: Nice. Changed

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/UnaryMethodHandler.java
tbak: Fixed

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/UnaryMethodHandler.java
tbak: Fixed

titus-server-runtime/src/main/java/com/netflix/titus/runtime/connector/common/reactor/server/ServerStreamingMethodHandler.java
```diff
@@ -42,24 +42,31 @@ public void invoke(REQ request, StreamObserver<RESP> responseObserver) {
...
-        return ((Flux<RESP>) result).subscribe(
-                responseObserver::onNext,
+        return Flux.from(result).subscribe(
+                value -> {
+                    try {
+                        responseObserver.onNext(value);
+                    } catch (Exception e) {
+                        logger.warn("Subscriber threw error in onNext handler", e);
+                        try {
+                            responseObserver.onError(e);
```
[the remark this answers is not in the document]
tbak: This should be handled by `ServerCallStreamObserver.setOnCancelHandler`. But I agree that the current implementation is not very elegant, and if the callback does not happen, we may continue emitting values to a closed channel.

PR #609, opened 2019-07-18
only log the full stacktrace at DEBUG level

fabiokung:
... since some exceptions are expected and OK (e.g.: image not found)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobImageSanitizer.java
```diff
@@ -95,7 +95,8 @@ private boolean isDisabled() {
 
     // Determines if this Exception should produce a sanitization failure
     private boolean isValidationOK(Throwable throwable, Image image) {
```
andrew-leung: Not part of your changes, but should probably be renamed to `isValidationExceptionOK()`. The method name is somewhat confusing.

amitaekbote: +1

fabiokung: Fixed.

PR #610, opened 2019-07-24
Add task context for tasks waiting for in use IP allocations

andrew-leung:
This PR adds a task context field to tasks that are not placed during scheduling because their assigned IP allocation is already in use by another task. This PR consists of the following changes:

1. Adds a new task placement failure type indicating that the task is waiting for an in use IP allocation.
2. When tasks are returned their task context is updated to reflect if the last placement failed because of an in use IP allocation.
3. Group task placement failures by task id to allow point queries when looking up failed allocations for specific tasks.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/IpAllocationConstraint.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/IpAllocationConstraint.java
@@ -41,12 +41,13 @@ import static com.netflix.titus.api.jobmanager.TaskAttributes.TASK_ATTRIBUTES_IP
 public class IpAllocationConstraint implements SystemConstraint {
 
     public static final String NAME = "IpAllocationConstraint";
+    private static final String IP_ALLOCATION_ALREADY_IN_USE_REASON_PREFIX = "Assigned IP allocation used by another task:";
+
 
     private static final Result VALID = new Result(true, null);
     private static final Result MACHINE_DOES_NOT_EXIST = new Result(false, "The machine does not exist");
     private static final Result IP_ALLOCATION_NOT_IN_ZONE = new Result(false, "Assigned IP allocation not in instance's zone");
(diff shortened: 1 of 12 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/IpAllocationConstraint.java
```diff
@@ -98,4 +99,18 @@ public Result evaluate(TaskRequest taskRequest, VirtualMachineCurrentState targe
...
+        return reason != null && reason.startsWith(IP_ALLOCATION_ALREADY_IN_USE_REASON_PREFIX);
+    }
+
+    public static Optional<String> getTaskIdFromIpAllocationInUseReason(String reason) {
+        return reason.startsWith(IP_ALLOCATION_ALREADY_IN_USE_REASON_PREFIX)
+                ? Optional.of(reason.substring(IP_ALLOCATION_ALREADY_IN_USE_REASON_PREFIX.length()))
+                : Optional.empty();
+    }
+
+    private static Result getIpAllocationInUseResult(String taskId) {
```
corindwyer: How often do you expect this result to be generated? A major issue we had in the past was creating too many `Result` objects leading to slowdowns caused by GC which is why we moved to static results for every constraint even though we lost visibility.

andrew-leung: Not often. I don't expect wide spread use of this feature to begin with and then this catches usages where either the user did something wrong or is part of a rolling deploy (e.g., redeploying ZK).

I understand the GC issue, though also don't have an obviously better place to stash the ID of the blocking task.

andrew-leung:
@fabiokung , thanks for the input. I agree there are scenarios where the context won't be accurate, though given that it's advisory information, thought that would be acceptable. Most of the existing task placement failure was handled with this mechanism, which is why I chose it. Happy to discuss changes to persist after the scheduling loop (either as part of this feature or as we refactor the placement service).

PR #615, opened 2019-07-31
Fix Terminated Accepted Tasks

corindwyer:
### Description of the Change
* Remove task from fenzo if the task state is not still in Accepted when attempting to launch.
* Change k8s GC to delete all pending pods with a deletion timestamp.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
@@ -608,26 +608,21 @@ public class KubeApiServerIntegrator implements VirtualMachineMasterService {
         }
         logger.debug("Finished orphaned pod GC without valid nodes");
 
-        Set<String> currentPodNames = currentPods.stream().map(p -> p.getMetadata().getName()).collect(Collectors.toSet());
-        Set<String> currentTaskIds = currentTasks.keySet();
-        List<V1Pod> podsStuckBeingDeleted = currentPods.stream()
+        List<V1Pod> pendingPodsWithDeletionTimestamp = currentPods.stream()
                 .filter(pod -> {
-                    String podName = pod.getMetadata().getName();
-                    DateTime deletionTime = pod.getMetadata().getDeletionTimestamp();
-                    return deletionTime != null &&
-                            clock.isPast(deletionTime.getMillis() + POD_GC_DELETION_TTL) &&
-                            !currentTaskIds.contains(podName);
+                    DateTime deletionTimestamp = pod.getMetadata().getDeletionTimestamp();
+                    return pod.getStatus().getPhase().equalsIgnoreCase(PENDING) && deletionTimestamp != null;
                 })
                 .collect(Collectors.toList());
 
-        // GC orphaned pods that are stuck being deleted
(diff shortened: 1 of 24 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -608,26 +608,21 @@ private void reconcileNodesAndPods() {
...
+                    return pod.getStatus().getPhase().equalsIgnoreCase(PENDING) && deletionTimestamp != null;
                 })
                 .collect(Collectors.toList());
 
-        // GC orphaned pods that are stuck being deleted
-        logger.debug("Attempting to GC {} orphaned pods not in Titus", podsStuckBeingDeleted.size());
-        for (V1Pod pod : podsStuckBeingDeleted) {
+        // GC pods in accepted with a deletion timestamp
+        logger.debug("Attempting to GC {} accepted pods with deletion timestamp", pendingPodsWithDeletionTimestamp.size());
+        for (V1Pod pod : pendingPodsWithDeletionTimestamp) {
```
amit-git: minor : could do the forEach instead

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementRecorder.java
```diff
@@ -155,8 +156,8 @@
             ).onErrorResumeNext(error -> {
                 Throwable recordTaskError = (Throwable) error; // Type inference issue
                 if (JobManagerException.hasErrorCode(recordTaskError, JobManagerException.ErrorCode.UnexpectedTaskState)) {
-                    // TODO More checking here. If it is actually running, we should kill it
-                    logger.info("Not launching task, as it is no longer in Accepted state (probably killed): {}", v3Task.getId());
+                    logger.info("Not launching task: {} and removing from fenzo as it is no longer in Accepted state (probably killed)", v3Task.getId());
```
fabiokung: minor: log the current task state here to aid in debugging

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementRecorder.java
```diff
@@ -155,8 +156,8 @@
...
+                    removeTask(assignmentResult, fenzoTask);
```
fabiokung: I understand it's not necessarily part of your changes, but I just noticed we may have the same problem in all cases below where we call `killBrokenV3Task()`. Should we remove the task from Fenzo in all cases we do `killBrokenV3Task()` as well?

fabiokung: I guess it will be eventually removed from Fenzo by line 130 (`removing after not finding jobMgr for task`) sometime in the future when the task gets archived, or by the next scheduling iteration when we try to place a Task that is in `KillInitiated|Finished` states, but it would be probably better to not leak resources for an extra scheduling iteration and waste some cycles trying to place them

corindwyer: I agree that this is problematic, but I am going to leave it the same for now as it will eventually get cleaned up. We will need to revisit this in the future.

PR #617, opened 2019-07-31
allow ENV to differ when moving tasks between jobs

fabiokung:
This allows jobs to have a "shadow job" counterpart with slightly different configuration (ENV), e.g. disabling some behavior, that will only be enabled when a task is moved to another job.

(internal ref: TITUS-2997)

### DIFF
--- a/titus-api/src/test/java/com/netflix/titus/api/jobmanager/model/job/JobCompatibilityTest.java
+++ b/titus-api/src/test/java/com/netflix/titus/api/jobmanager/model/job/JobCompatibilityTest.java
@@ -26,6 +26,7 @@ import com.netflix.titus.common.util.CollectionsExt;
 import com.netflix.titus.testkit.model.job.JobDescriptorGenerator;
 import org.junit.Test;
 
+import static com.netflix.titus.common.util.CollectionsExt.copyAndAdd;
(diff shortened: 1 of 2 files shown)


titus-api/src/test/java/com/netflix/titus/api/jobmanager/model/job/JobCompatibilityTest.java
```diff
@@ -102,17 +103,29 @@ public void testDisruptionBudgetIsIgnored() {
...
 
+    @Test
+    public void testEnvIsIgnored() {
+        JobDescriptor<ServiceJobExt> reference = JobDescriptorGenerator.oneTaskServiceJobDescriptor();
+        JobDescriptor<ServiceJobExt> other = reference.toBuilder()
+                .withContainer(reference.getContainer().but(container ->
+                        container.toBuilder().withEnv(copyAndAdd(container.getEnv(), "SOME", "OVERRIDE"))
+                ))
+                .build();
+        JobCompatibility compatibility1 = JobCompatibility.of(reference, other);
```
andrew-leung: This test only has a simple `JobCompatibility `. Should the variable be just `compatibility`?

PR #619, opened 2019-08-05
Add API validations for IP allocations

andrew-leung:
This PR adds two validations for IP allocations:

- Check that provided IP allocations signatures are Base64 encoded.
- Check that jobs cannot be created or updated with a max or size greater than their number of IP allocations.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
@@ -358,8 +358,8 @@ public class DefaultJobManagementServiceGrpc extends JobManagementServiceGrpc.Jo
     @Override
     public void updateJobCapacity(JobCapacityUpdate request, StreamObserver<Empty> responseObserver) {
         execute(callMetadataResolver, responseObserver, callMetadata -> {
-            verifyServiceJob(request.getJobId());
             CapacityAttributes capacityAttributes = V3GrpcModelConverters.toCoreCapacityAttributes(request.getCapacity());
+            verifyServiceJob(request.getJobId(), capacityAttributes);
 
             authorizeJobUpdate(callMetadata, request.getJobId())
                     .concatWith(jobOperations.updateJobCapacityAttributesReactor(request.getJobId(), capacityAttributes, callMetadata))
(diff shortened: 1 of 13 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
```diff
@@ -816,18 +815,22 @@ private JobChangeNotification toJobChangeNotification(com.netflix.titus.api.jobm
...
 
-    private com.netflix.titus.api.jobmanager.model.job.Job<ServiceJobExt> verifyServiceJob(String jobId) {
+    private com.netflix.titus.api.jobmanager.model.job.Job<ServiceJobExt> verifyServiceJob(String jobId, CapacityAttributes capacityAttributes) {
         return jobOperations
                 .getJob(jobId)
                 .map(j -> {
                     if (!JobFunctions.isServiceJob(j)) {
                         throw JobManagerException.notServiceJob(j.getId());
+                    } else if (j.getJobDescriptor().getContainer().getContainerResources().getSignedIpAddressAllocations().size() > 0 &&
+                            capacityAttributes.getMax().orElse(0) > j.getJobDescriptor().getContainer().getContainerResources().getSignedIpAddressAllocations().size()) {
```
fabiokung: the boolean logic seems to be flipped here, `max > ipAllocations` is what we expect, and shouldn't generate an error, no?

I think this check needs to be performed only when `max` or `desired` are being changed, users are not required to always specify a `max` for every `updateCapacity` call.

fabiokung: It seems tests didn't catch the wrong logic here, it may be worth adding a few test cases with partial capacity updates to cover this.

andrew-leung: I _think_ the logic here is correct (maybe I've been looking at it too long). If this is a service job, check if there are any Job IP Allocations (we don't want to make the further checks if the job does not have allocations), and if it does then call `.getMax().orElse(0)`.

In cases where the legacy job capacity API is used, all 3 capacity values (min, max, desired) are set and we'll get the `max` value being set and check that it is not larger than the number of IP allocations.

In cases where the new job capacity API is used and max is set we'll get max and make sure it's not larger than the number of IP allocations.

In cases where the new job capacity API is used and max is not being set we'll get the fallback of `0`, which is not going to be larger than the number of IP allocations (since we checked it is `> 0` above) so it does not throw. In this last case, we then assume the previously set max is valid and the other values being updated (min and/or desired) are checked elsewhere.

Let me know if there are details I'm missing here.

fabiokung: nvm, I read this incorrectly.

PR #625, opened 2019-08-12
Initial version of K8S based cluster membership connector

tbak:
This is still work in progress. I am sharing this code early so we can discuss the correctness of this approach before the full implementation is ready. Service/connector layers are functionally complete (but still miss logs/metrics/tests). GRPC layer implementation is trivial, so I will do at the end.


titus-api/src/main/java/com/netflix/titus/api/clustermembership/connector/mixin/ClusterMembershipRevisionMixIn.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.titus.api.clustermembership.connector.mixin;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.netflix.titus.api.clustermembership.model.ClusterMember;
+
+public abstract class ClusterMembershipRevisionMixIn {
```
fabiokung: add `@JsonIgnoreProperties(ignoreUnknown = true)` preemptively to allow backwards compatible changes later

tbak: I will remove this mixin (no longer needed).

PR #626, opened 2019-08-13
Junit tests for SimpleReconciliationEngine

tbak:


titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/internal/ChangeActionHolder.java
```diff
@@ -67,4 +78,11 @@ boolean isCancelled() {
     MonoSink<DATA> getSubscriberSink() {
         return subscriberSink;
     }
+
+    void addCancelCallback(Disposable cancelCallback) {
+        cancelCallbacks.add(cancelCallback);
+        if (cancelledRef.get()) {
+            ReactorExt.safeDispose(cancelCallback);
```
fabiokung: can these be safely executed more than once?

fabiokung: rephrasing: maybe add some documentation that these callbacks may be called multiple times and should be idempotent

tbak: Idempotens is the API contract for `Disposable#dispose`:
```java
	/**
	 * Cancel or dispose the underlying task or resource.
	 * <p>
	 * Implementations are required to make this method idempotent.
	 */
	void dispose();
```
I will add a comment there.

PR #628, opened 2019-08-15
Update K8S Integration

corindwyer:
### Description of the Change
* Update reason classifications to change the code based on a pattern.
* Change GC logic for tasks that don't exist in api server.
* Use pod message instead of reason for Titus reasonMessage field.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
@@ -28,6 +28,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Singleton;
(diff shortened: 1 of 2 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -518,29 +538,44 @@ private void podUpdated(V1Pod pod) {
...
         } else if (phase.equalsIgnoreCase(SUCCEEDED)) {
-            publishContainerEvent(podName, Finished, REASON_NORMAL, reason, Optional.empty());
+            publishContainerEvent(podName, Finished, REASON_NORMAL, reasonMessage, Optional.empty());
         } else if (phase.equalsIgnoreCase(FAILED)) {
-            publishContainerEvent(podName, Finished, REASON_FAILED, reason, Optional.empty());
+            publishContainerEvent(podName, Finished, REASON_FAILED, reasonMessage, Optional.empty());
         }
     }
 
     private void publishContainerEvent(String taskId, TaskState taskState, String reasonCode, String reasonMessage,
```
fabiokung: is there anything useful that can be in the old `reasonCode` we may be potentially discarding here?

corindwyer: It is mapped to the pod phase so it is not useful at this time unless we add more codes.

PR #632, opened 2019-08-20
Schedule batch tasks on opportunistic CPUs

fabiokung:
This PR adds support for oversubscribing agent machines, and scheduling tasks onto opportunistic (underutilized) CPUs.

By default, a noop implementation of `OpportunisticCpuAvailabilityProvider` is used. A following PR will add support for fetching opportunistic CPUs from k8s clusters (kube api-server), when Titus is configured to run tasks on k8s.

Opportunistic CPUs are only enabled for batch jobs that include a prediction of their tasks runtime. IOW, a batch job needs to opt into opportunistic scheduling by having an attribute that hints at its runtime duration. The Titus scheduler will then match that job with opportunistic CPUs that will be available during that window.

## Key aspects of this implementation

* Before each scheduling loop iteration, we compute all available opportunistic CPUs, and all current usage. That information is cached for a single scheduling iteration.
* Tasks requesting opportunistic CPUs can only be scheduled on machines that have opportunistic CPUs available.
* We first try to schedule batch jobs with runtime predictions by allocating only opportunistic CPUs (all or none);
* if it fails (say no opportunistic CPUs available on any compatible machine), we then decrement the number of requested opportunistic CPUs for the task, which increments the number of regular CPUs requested, and try again;
* ultimately, if there are no compatible machines with any opportunistic CPUs available, we fall back to scheduling regularly with all "normal" CPUs.
* If the above fails again, we loop back to scheduling with all opportunistic CPUs, in case more are available now, and the cycle repeats. The tradeoff is that opportunistic batch jobs can take multiple scheduling iteration cycles to be placed on a machine, but we always try to maximize opportunistic usage of CPUs.
* Tasks will have metadata in their `taskContext` about opportunistic CPU allocation when they were scheduled opportunistically.
* Other resource dimensions (memory, network, disk, ENIs, etc) are consumed and allocated normally.
* Existing resource consumption metrics for CPU will only account for regular CPU usage, opportunistic CPUs will count as `zero` CPU consumption towards metrics and capacity groups.

## Future improvements

* Fitness function that binpacks opportunistic CPUs, and minimizes fragmentation
* Opportunistic CPU consumption metrics
* kube implementation of `OpportunisticCpuAvailabilityProvider`, compatible with how kubelets will emit their underutilized (opportunistic) CPU resources.

cc @jigish @gabrielhartmann @rostyboost


titus-server-master/src/main/java/com/netflix/titus/master/scheduler/opportunistic/OpportunisticCpuAvailability.java
```diff
@@ -0,0 +1,49 @@
...
+package com.netflix.titus.master.scheduler.opportunistic;
+
+import java.time.Instant;
+
+import com.netflix.titus.common.util.time.Clock;
+
+public class OpportunisticCpuAvailability {
+    private final String allocationId;
+    private final Instant expiresAt;
+    private final int count;
```
rostyboost: Could we make this a `float` instead? so it matches https://github.com/Netflix-Skunkworks/titus-isolate/pull/198/files#diff-88d117d04068fc260a43286f0d058b59R7

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/OpportunisticCpuConstraint.java
```diff
@@ -0,0 +1,106 @@
...
+            titusRuntime.getCodeInvariants().inconsistent("No machine id attribute filled by Fenzo");
+            return NO_OPPORTUNISTIC_CPUS;
+        }
+
+        Optional<OpportunisticCpuAvailability> availabilityOpt = opportunisticCpuCache.findAvailableOpportunisticCpus(agentId);
+        if (!availabilityOpt.isPresent()) {
+            return NO_OPPORTUNISTIC_CPUS;
+        }
+
+        Duration taskRuntime = runtimePrediction.get();
```
rostyboost: if `getJobRuntimePrediction` in `JobManagerUtil` were to return the whole quantile predictions from `runtimePrediction.available`, we could implement more advanced policies (like the one I presented last week, using multiple predicted quantiles)

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/opportunistic/OpportunisticCpuAvailability.java
fabiokung: I would really, *really*, prefer to avoid it and avoid headaches with floating point based math everywhere

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/OpportunisticCpuConstraint.java
fabiokung: The support for custom policies is in some non-OSS extensions that we have today, they do exactly what you mention. We may move them into the OSS codebase in the future if it works well.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementRecorder.java
```diff
@@ -207,6 +220,23 @@ private void removeTask(TaskAssignmentResult assignmentResult, TitusQueuableTask
...
     }
 
+    private Map<String, String> buildOpportunisticResourcesContext(VirtualMachineLease lease, TitusQueuableTask fenzoTask) {
+        int count = fenzoTask.getOpportunisticCpus();
+        if (count < 1) {
+            return Collections.emptyMap();
+        }
+        Optional<OpportunisticCpuAvailability> availability = opportunisticCpuCache.findAvailableOpportunisticCpus(lease.getVMID());
+        if (!availability.isPresent()) {
+            titusRuntime.getCodeInvariants().inconsistent("Task %s was scheduled on opportunistic CPUs that are not available",
```
gabrielhartmann: I don't understand this message.  What does "unavailable" mean?  They exist but are from the wrong time period?

fabiokung: This should never happen. If it happens, likely I made a mistake. Code invariants are markers we use to point to inconsistencies that need to be fixed, and the code is likely wrong somewhere.

"Not available" here means the opportunistic resource was available right before during the scheduling loop, and we can't find it here when recording the placement decision anymore. It should never happen since we freeze the state of the world (resources and machines available) during the scheduling iteration.

gabrielhartmann: ok, thanks.  makes sense

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/opportunistic/OpportunisticCpuAvailability.java
```diff
@@ -0,0 +1,49 @@
...
+
+package com.netflix.titus.master.scheduler.opportunistic;
+
+import java.time.Instant;
+
+import com.netflix.titus.common.util.time.Clock;
+
+public class OpportunisticCpuAvailability {
+    private final String allocationId;
+    private final Instant expiresAt;
```
tbak: `long` for better performance?

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/OpportunisticCpuAllocation.java
```diff
@@ -0,0 +1,117 @@
...
+    public static OpportunisticCpuAllocationBuilder newBuilder() {
+        return new OpportunisticCpuAllocationBuilder();
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(taskId, agentId, allocationId, cpuCount);
+    }
+
+    public static final class OpportunisticCpuAllocationBuilder {
```
tbak: Just `Builder` to align with the rest of the code?

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/OpportunisticCpuAllocation.java
```diff
@@ -0,0 +1,117 @@
...
+
+import java.util.Objects;
+
+public class OpportunisticCpuAllocation {
+    private final String taskId;
+    private final String agentId;
+    private final String allocationId;
+    private final int cpuCount;
+
+    public OpportunisticCpuAllocation(String taskId, String agentId, String allocationId, int cpuCount) {
```
tbak: private as we have the builder?

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/OpportunisticCpuAllocation.java
```diff
@@ -0,0 +1,117 @@
...
+            this.allocationId = allocationId;
+            return this;
+        }
+
+        public OpportunisticCpuAllocationBuilder withCpuCount(int cpuCount) {
+            this.cpuCount = cpuCount;
+            return this;
+        }
+
+        public OpportunisticCpuAllocation build() {
```
tbak: Check arguments to eliminate null values.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/TaskCache.java
```diff
@@ -116,28 +134,69 @@ private void buildTaskCacheInfo(List<Pair<Job, List<Task>>> jobsAndTasks) {
...
             }
-            titusRuntime.getCodeInvariants().inconsistent("Unable to find zone for IP allocation ID {} in job allocations {}",
+            codeInvariants().inconsistent("Unable to find zone for IP allocation ID {} in job allocations {}",
                     ipAllocationId, jobDescriptor.getContainer().getContainerResources().getSignedIpAddressAllocations());
             return Optional.empty();
         }
+
+        private Optional<OpportunisticCpuAllocation> getOpportunisticCpuAllocation(Task task) {
+            Optional<String> allocationIdOpt = getOpportunisticCpuAllocationId(task);
+            Optional<OpportunisticCpuAllocation> allocationOpt = allocationIdOpt
```
tbak: Optimize?

titus-server-master/src/test/java/com/netflix/titus/master/integration/v3/OpportunisticCpuSchedulingTest.java
```diff
@@ -0,0 +1,196 @@
...
+                .allTasks(taskScenarioBuilder -> taskScenarioBuilder
+                        .expectTaskOnAgent()
+                        .expectTaskContext(TASK_ATTRIBUTES_AGENT_ID, opportunisticCpus.agentId)
+                        .expectTaskContext(TASK_ATTRIBUTES_OPPORTUNISTIC_CPU_ALLOCATION, opportunisticCpus.allocationId)
+                        .expectTaskContext(TASK_ATTRIBUTES_OPPORTUNISTIC_CPU_COUNT, "2")
+                )
+        );
+    }
+
+    private OpportunisticCpus addOpportunisticCpusToFlex(int count, Instant expiresAt) {
```
tbak: This could be added to `InstanceGroupsScenarioBuilder`

titus-api/src/main/java/com/netflix/titus/api/jobmanager/JobAttributes.java
```diff
@@ -53,6 +54,11 @@
      */
     public static final String JOB_ATTRIBUTES_SANITIZATION_SKIPPED_IMAGE = JOB_ATTRIBUTE_SANITIZATION_PREFIX + "skipped.image";
 
+    /**
+     * Predicted runtime for a particular job in seconds, used in opportunistic CPU scheduling
+     */
+    public static final String JOB_ATTRIBUTES_RUNTIME_PREDICTION_VALUE = PREDICTION_ATTRIBUTE_PREFIX + "value";
```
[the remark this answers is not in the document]
fabiokung: there is already some code and analysis pipeline using this name, so I'm a bit hesitant in changing it, but I can if you think it is a big deal. If you look at the full name, including the prefix, it maybe would be a bit too redundant?

```
titus.runtimePrediction.value
titus.runtimePrediction.predictedRuntimeSec
```

titus-api/src/main/java/com/netflix/titus/api/jobmanager/TaskAttributes.java
```diff
@@ -45,6 +45,8 @@
     public static final String TASK_ATTRIBUTES_TIER = "task.tier";
     public static final String TASK_ATTRIBUTES_IP_ALLOCATION_ID = "task.ipAllocationId";
     public static final String TASK_ATTRIBUTES_IN_USE_IP_ALLOCATION = "task.ipAllocationAlreadyInUseByTask";
+    public static final String TASK_ATTRIBUTES_OPPORTUNISTIC_CPU_ALLOCATION = "task.opportunisticCpuAllocation";
```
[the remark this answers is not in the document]
fabiokung: one is an uuid, the other is a count of opportunistic CPUs (int). I will add a comment.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobReconciliationFrameworkFactory.java
```diff
@@ -304,11 +305,16 @@ private TaskFenzoCheck addTaskToFenzo(ReconciliationEngine<JobManagerReconcilerE
         try {
             Pair<Tier, String> tierAssignment = JobManagerUtil.getTierAssignment(job, capacityGroupService);
             String host = task.getTaskContext().get(TaskAttributes.TASK_ATTRIBUTES_AGENT_HOST);
+            int opportunisticCpus = Integer.parseInt(task.getTaskContext().getOrDefault(
```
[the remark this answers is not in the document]
fabiokung: good catch! (no pun intended)

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/common/V3QueueableTask.java
```diff
@@ -237,4 +299,13 @@ private void setupCustomNamedResources(Job<?> job) {
     private static String getConcatenatedSecurityGroups(List<String> securityGroups) {
         return String.join(":", securityGroups);
     }
+
+    private static int initialOpportunisticCpuCount(JobDescriptor jobDescriptor) {
+        return initialOpportunisticCpuCount(jobDescriptor.getContainer().getContainerResources().getCpu());
+    }
+
+    private static int initialOpportunisticCpuCount(double cpusRequested) {
+        return (int) Math.floor(cpusRequested);
```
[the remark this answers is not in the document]
fabiokung: if we ever have fractional CPU asks, I'm making an assumption that only the floor integer part of it will be allocated opportunistically, and the fractional part would come from regular CPUs

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/OpportunisticCpuCache.java
```diff
@@ -0,0 +1,52 @@
...
+    public OpportunisticCpuCache(OpportunisticCpuAvailabilityProvider provider, TitusRuntime titusRuntime) {
+        this.provider = provider;
+        this.titusRuntime = titusRuntime;
+    }
+
+    public void prepare() {
+        snapshot = provider.getOpportunisticCpus();
+    }
+
+    public Optional<OpportunisticCpuAvailability> findAvailableOpportunisticCpus(String agentId) {
```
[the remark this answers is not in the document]
fabiokung: this is not using the `Stream` API, it's flatMapping an `Optional`

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/OpportunisticCpuAllocation.java
fabiokung: will do, I used a plugin to autogenerate this

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/OpportunisticCpuAllocation.java
fabiokung: I considered, but the one-shot constructor seemed a good fit in `TaskCacheEventListener` where we didn't need to build the POJO incrementally

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/OpportunisticCpuAllocation.java
fabiokung: I'm not opposed, but tried to be consistent, we don't do this in any other `Builder` I could find

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/TaskCache.java
fabiokung: did you mean `inline`? I'm using the `allocationIdOpt` variable again below

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/OpportunisticCpuConstraint.java
```diff
@@ -0,0 +1,106 @@
...
+        Preconditions.checkArgument(taskRequest instanceof V3QueueableTask, "opportunistic CPUs can only be used with V3QueueableTask");
+        V3QueueableTask request = (V3QueueableTask) taskRequest;
+
+        if (!request.isCpuOpportunistic()) {
+            return VALID;
+        }
+
+        Optional<Duration> runtimePrediction = request.getRuntimePrediction();
+        if (!runtimePrediction.isPresent()) {
+            titusRuntime.getCodeInvariants().inconsistent("Task requested opportunistic CPUs without a runtime prediction");
```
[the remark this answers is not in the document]
fabiokung: Fair enough, will do.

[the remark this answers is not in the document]
fabiokung: Fixed. @corindwyer PTAL

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/OpportunisticCpuConstraint.java
```diff
@@ -0,0 +1,106 @@
...
+            return NO_OPPORTUNISTIC_CPUS;
+        }
+
+        Optional<OpportunisticCpuAvailability> availabilityOpt = opportunisticCpuCache.findAvailableOpportunisticCpus(agentId);
+        if (!availabilityOpt.isPresent()) {
+            return NO_OPPORTUNISTIC_CPUS;
+        }
+
+        Duration taskRuntime = runtimePrediction.get();
+        Instant now = Instant.ofEpochMilli(titusRuntime.getClock().wallTime());
```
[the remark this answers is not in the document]
fabiokung: agree, but opted for readability here, since it's a code path that's only triggered for a few tasks that have duration predictions, and a few agents that have opportunistic CPUs

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/TaskCacheEventListener.java
```diff
@@ -17,34 +17,76 @@
...
-    public TaskCacheEventListener(TaskCache taskCache) {
+    public TaskCacheEventListener(TaskCache taskCache, OpportunisticCpuCache opportunisticCpuCache, TitusRuntime titusRuntime) {
         this.taskCache = taskCache;
+        this.opportunisticCpuCache = opportunisticCpuCache;
+        this.titusRuntime = titusRuntime;
     }
 
     @Override
     public void onScheduleStart() {
+        taskCache.prepare();
```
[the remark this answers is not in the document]
fabiokung: You are right that `opportunisticCpuCache` is always `current_allocatoin + unassigned_allocation`. You are right I'm not discounting used opportunistic CPUs in `OpportunisticCpuCache` as we schedule tasks (that will always have the total advertised per agents), but I do that accounting in the scheduling constraint directly (`OpportunisticCpuConstraint`).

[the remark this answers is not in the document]
fabiokung: in other words, yes, we don't keep the final number of free opportunistic CPUs in any of the caches, we always compute it on the fly.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/opportunistic/OpportunisticCpuAvailability.java
fabiokung: I really prefer the readability and integration with the rest of the codebase, since it is not on the super hot path of the scheduling loop (code that uses it is only triggered when both the task and the agent match opportunistic requirements), would you be OK if we try it like this first, and optimize later as necessary?

titus-server-master/src/test/java/com/netflix/titus/master/integration/v3/OpportunisticCpuSchedulingTest.java
fabiokung: Done.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/opportunistic/OpportunisticCpuAvailability.java
fabiokung: We can revisit if we ever add fractional CPUs (which can also be done with ints, e.g.: `1.5 CPU == 150, 2.3 CPU == 230, etc`

titus-api/src/main/java/com/netflix/titus/api/jobmanager/JobAttributes.java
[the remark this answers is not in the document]
tbak: This name is part of Task external representation, and `value` does not give any hint what it is or what is the unit. Specially the latter may be constant source of confusion, as most of time intervals are defined as ms.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/common/V3QueueableTask.java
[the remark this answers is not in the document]
tbak: This sounds like an architectural decision, with some good rationale behind. Could you write a short comment here why is that?

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/OpportunisticCpuCache.java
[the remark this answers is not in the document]
tbak: It is about creating temporal wrapper objects like `Optional`. We used to have performance issues caused by that.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/TaskCache.java
tbak: No, it is again about not using functional API, as it will produce several temporal objects.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/JobAttributes.java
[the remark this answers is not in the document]
fabiokung: fair enough. Given it's not widely deployed yet I will just change it.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/resourcecache/OpportunisticCpuCache.java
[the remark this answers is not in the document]
fabiokung: Got it. As discussed, I will add a feature toggle to disable the entire code path in case there is a performance issue.

fabiokung:
@tbak @corindwyer some more fixes based on reviews, and I added a feature flag to enable/disable opportunistic scheduling dynamically. It only controls the behavior inside each scheduling loop iteration, new pre-scheduling and post-scheduling hooks will always be executed regardless of the flag.

Given we have more concerns about code running in the scheduling loop for all tasks, I think this was a reasonable compromise. Thoughts?

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/common/V3QueueableTask.java
```diff
@@ -226,19 +235,27 @@ public void opportunisticSchedulingFailed() {
         if (!runtimePrediction.isPresent()) {
             return; // noop, opportunisticCpuCount is always 0
         }
-        opportunisticCpuCount.updateAndGet(current -> current >= 1 ? current - 1 : initialOpportunisticCpuCount(cpus));
+        if (!opportunisticEnabledSupplier.get()) {
+            opportunisticCpuEnabled = false;
```
tbak: I do not see a place where it is reset. Unless we do not want to do it for a particular task?

fabiokung: good catch

fabiokung: Fixed

PR #636, opened 2019-08-22
Fix out of order cancel execution in DefaultLocalScheduler

tbak:


titus-common/src/main/java/com/netflix/titus/common/framework/scheduler/internal/DefaultLocalScheduler.java
```diff
@@ -258,7 +258,11 @@ public boolean isClosed() {
 
                 @Override
                 public void cancel() {
-                    cancelInternal();
+                    // Run the cancel request via event loop so it is properly serialized with other concurrent updates.
+                    DefaultLocalScheduler.this.cancel(scheduleId).subscribe(
+                            next -> logger.debug("Action cancelled ba a user: {}", executor.getAction().getExecutionId()),
```
corindwyer: typo

PR #637, opened 2019-08-22
Fix race condition in the DefaultSimpleReconciliationEngineTest.testReconcilerAction test

tbak:


titus-common/src/test/java/com/netflix/titus/common/framework/simplereconciler/internal/DefaultSimpleReconciliationEngineTest.java
```diff
@@ -214,6 +214,9 @@ private void newReconciler(Function<String, List<Mono<Function<String, String>>>
         this.changesSubscriber = new TitusRxSubscriber<>();
         reconciliationEngine.changes().subscribe(changesSubscriber);
 
-        assertThat(changesSubscriber.takeNext()).isIn(INITIAL, RECONCILED);
+        String event = changesSubscriber.takeNext();
+        if (!event.equals(INITIAL) && !event.startsWith(RECONCILED)) {
```
fabiokung: do you care about the order of these events?

tbak: No. It is possible that `INITIAL` is emitted before subscription happens, so we have to check for `RECONCILED` as well.

PR #638, opened 2019-08-23
Publish jobs in the taskpublisher supplementary component

tbak:


titus-supplementary-component/tasks-publisher/src/main/java/com/netflix/titus/supplementary/taskspublisher/config/EsPublisherConfiguration.java
```diff
@@ -42,6 +42,9 @@
     @Value("${titus.es.taskDocumentEsIndexName}")
     private String taskDocumentEsIndexName;
 
+    @Value("${titus.es.enabled}")
```
amit-git: minor : titus.es.publish.enabled  would be more explicit

PR #647, opened 2019-09-04
Move common code to titus-common-XXX submodules

tbak:
In this PR I have moved the new cluster membership component, with its common dependencies. This is only the first step. I expect the full refactoring to take multiple iterations.


tbak:
@andrew-leung @amitaekbote @amit-git ^^^

PR #648, opened 2019-09-04
Synchronize concurrent writes to Flux DirectProcessor

tbak:


titus-common-runtime/src/test/java/com/netflix/titus/runtime/clustermembership/service/ClusterMembershipConnectorStub.java
```diff
@@ -142,4 +142,10 @@ void becomeLeader() {
                 )
                 .build();
     }
+
+    private void emitEvent(ClusterMembershipEvent event) {
+        synchronized (eventProcessor) {
```
fabiokung: it might be better to use a `DirectProcessor#sink()` to send events to the processor, which serializes all `onNext` events.

tbak: Good idea. I did not know about that one.

PR #649, opened 2019-09-04
Buffer reconciliation framework events

corindwyer:
### Description of the Change
* Buffer reconciliation framework events and move API requests to IO thread
* Add metric for total active jobs

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
@@ -153,7 +153,18 @@ public class DefaultV3JobOperations implements V3JobOperations {
         this.transactionLoggerSubscription = JobTransactionLogger.logEvents(reconciliationFramework);
 
         // Remove finished jobs from the reconciliation framework.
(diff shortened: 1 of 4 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
```diff
@@ -152,21 +153,22 @@ public void enterActiveMode() {
...
-                event -> {
-                    if (event instanceof JobModelUpdateReconcilerEvent) {
-                        JobModelUpdateReconcilerEvent jobUpdateEvent = (JobModelUpdateReconcilerEvent) event;
-                        handleJobCompletedEvent(jobUpdateEvent.getChangedEntityHolder());
-                    }
-                },
-                e -> logger.error("Event stream terminated with an error", e),
-                () -> logger.info("Event stream completed")
-        );
+        this.reconcilerEventSubscription = titusRuntime.persistentStream(reconciliationFramework.events().onBackpressureBuffer())
```
tbak: This creates unbounded buffer. I think we should do on back-pressure drop, and do full reconciliation at subscription time.

corindwyer: changed

amitaekbote:
thanks for fixing this!

PR #650, opened 2019-09-04
track usage of opportunistic CPUs

fabiokung:
.... as a consumption metric.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/model/v1/V1OpportunisticResource.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/model/v1/V1OpportunisticResource.java
@@ -137,4 +137,11 @@ public class V1OpportunisticResource {
         }
         return spec.getCapacity().getCpu();
     }
+
+    public String getName() {
+        if (metadata == null || metadata.getName() == null) {
+            return "";
+        }
+        return metadata.getName();
+    }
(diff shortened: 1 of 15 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/model/v1/V1OpportunisticResource.java
```diff
@@ -137,4 +137,11 @@ public int getCpus() {
         }
         return spec.getCapacity().getCpu();
     }
+
+    public String getName() {
+        if (metadata == null || metadata.getName() == null) {
```
tbak: Is this a valid state when any of these is null? Empty string seems misleading.

corindwyer: You could use an unknown constant here like the other places you used one, but I am assuming this is not supposed to happen as the metadata name should be the unique key of the resource. Is that a correct assumption?

fabiokung: that is correct, the name should never be null. Ideally I'd want this POJO to be immutable and require non-nulls, but GSON (used by kubernetes-client-java) doesn't support it, so I'm just being extra defensive here.

PR #654, opened 2019-09-06
Add taints and tolerations

corindwyer:
### Description of the Change
* Add agent instance and instance group taints attribute and job tolerations attribute.
* Set K8s restart policy to `Never`.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/AgentManagementConstraint.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/AgentManagementConstraint.java
@@ -25,7 +25,6 @@ import javax.inject.Singleton;
 import com.netflix.fenzo.TaskRequest;
 import com.netflix.fenzo.TaskTrackerState;
 import com.netflix.fenzo.VirtualMachineCurrentState;
-import com.netflix.fenzo.queues.QueuableTask;
 import com.netflix.titus.api.agent.model.AgentInstance;
 import com.netflix.titus.api.agent.model.AgentInstanceGroup;
 import com.netflix.titus.api.agent.model.InstanceGroupLifecycleState;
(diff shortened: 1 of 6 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/AgentManagementConstraint.java
```diff
@@ -214,4 +210,37 @@ private Result evaluateAgentInstanceAttributes(AgentInstance agentInstance) {
...
+    private boolean taskCanTolerateTaints(V3QueueableTask taskRequest, AgentInstanceGroup instanceGroup, AgentInstance instance) {
+        Set<String> taints = getTaints(instanceGroup, instance);
+        Set<String> tolerations = getTolerations(taskRequest);
+        if (taints.isEmpty() && tolerations.isEmpty()) {
+            return true;
+        }
+
+        for (String taint : taints) {
+            if (tolerations.contains(taint)) {
+                return true;
```
tbak: We should tolerate all taints.

corindwyer: This was a mistake. Fixed

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/AgentManagementConstraint.java
```diff
@@ -219,12 +219,12 @@ private boolean taskCanTolerateTaints(V3QueueableTask taskRequest, AgentInstance
         }
 
         for (String taint : taints) {
-            if (tolerations.contains(taint)) {
-                return true;
+            if (!tolerations.contains(taint)) {
```
tbak: Simplify with containsAll?

corindwyer: updated

PR #655, opened 2019-09-06
Support cross account connectivity in AwsInstanceCloudConnector

tbak:


titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonStsAsyncProvider.java
```diff
@@ -0,0 +1,46 @@
...
+@Singleton
+public class AmazonStsAsyncProvider implements Provider<AWSSecurityTokenServiceAsync> {
+
+    private final AWSSecurityTokenServiceAsync amazonStsAsync;
+
+    @Inject
+    public AmazonStsAsyncProvider(AwsConfiguration configuration, AWSCredentialsProvider credentialProvider) {
+        String region = configuration.getRegion().trim().toLowerCase();
+        this.amazonStsAsync = AWSSecurityTokenServiceAsyncClientBuilder.standard()
+                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration("sts." + region + ".amazonaws.com", region))
```
andrew-leung: Can you just set a `.withRegion()` and have the endpoint automatically built? Or do you actually need to set the endpoint URL?

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AwsConfiguration.java
```diff
@@ -63,4 +63,17 @@
      */
     @DefaultValue("60000")
     long getIamRoleCacheTimeoutMs();
+
+    /**
+     * IAM role ARN to assume into to access AWS API for the data plane account. If not set, it is assumed that
+     * the control plane and the data plane run in the same account and no cross access is required.
+     */
+    @DefaultValue("")
+    String getDataPlaneRoleArn();
```
andrew-leung: Once we add IAM validation, the control plane will still need to know the data plane Role ARN even if it's in the same account because it would still need to assume into it to do validation.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonStsAsyncProvider.java
tbak: If I remember correctly, in the past we had to do both. I am following this pattern to keep it consistent with other providers.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AwsConfiguration.java
tbak: Good point. We will need more configuration parameters to control that. For now, I will keep this simple rule.

PR #656, opened 2019-09-09
AB testing for JobRuntimePredictionSelector

tbak:
Initial proposal.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/DateTimeExt.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/DateTimeExt.java
@@ -63,7 +63,7 @@ public final class DateTimeExt {
         if (msSinceEpoch == 0L) {
             return null;
         }
-        return ISO_LOCAL_DATE_TIME_FORMATTER.format(Instant.ofEpochMilli(msSinceEpoch)) + 'Z';
+        return ISO_LOCAL_DATE_TIME_FORMATTER.format(Instant.ofEpochMilli(msSinceEpoch));
     }
 
     public static String toTimeUnitAbbreviation(TimeUnit timeUnit) {
--- a/titus-common/src/main/java/com/netflix/titus/common/util/DateTimeExt.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/DateTimeExt.java
@@ -31,6 +31,7 @@ import com.google.protobuf.util.Durations;
 public final class DateTimeExt {
 
     private static final DateTimeFormatter ISO_UTC_DATE_TIME_FORMATTER = DateTimeFormatter.ISO_LOCAL_DATE_TIME.withZone(ZoneId.of("UTC"));
+    private static final DateTimeFormatter ISO_LOCAL_DATE_TIME_FORMATTER = DateTimeFormatter.ISO_LOCAL_DATE_TIME.withZone(ZoneId.systemDefault());
 
     private DateTimeExt() {
     }
@@ -56,6 +57,13 @@ public final class DateTimeExt {
             return null;
         }
         return ISO_UTC_DATE_TIME_FORMATTER.format(Instant.ofEpochMilli(msSinceEpoch)) + 'Z';
+    }
+
+    public static String toLocalDateTimeString(long msSinceEpoch) {
+        if (msSinceEpoch == 0L) {
+            return null;
+        }
+        return ISO_LOCAL_DATE_TIME_FORMATTER.format(Instant.ofEpochMilli(msSinceEpoch)) + 'Z';
     }
 
     public static String toTimeUnitAbbreviation(TimeUnit timeUnit) {
--- a/titus-common/src/main/java/com/netflix/titus/common/util/closeable/DefaultCloseableReference.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/closeable/DefaultCloseableReference.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.common.util.closeable;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+
+import com.netflix.titus.common.util.ExceptionExt;
+
+class DefaultCloseableReference<T> implements CloseableReference<T> {
+
+    private final T resource;
+    private final Runnable closeAction;
+    private final boolean serialize;
+    private final boolean swallowException;
+    private final Runnable onSuccessCallback;
+    private final Consumer<Throwable> onErrorCallback;
+
+    private final AtomicBoolean closed = new AtomicBoolean();
+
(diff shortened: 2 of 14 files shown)


fabiokung:
@tbak LGTM, the only thing missing is a mechanism to have different configuration depending on what A/B cell is selected

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSelectors.java
```diff
@@ -58,18 +59,30 @@ public static JobRuntimePredictionSelector best(Map<String, String> selectionMet
     /**
      * Select prediction with the lowest confidence level which is above the provided confidenceThreshold.
```
fabiokung: javadoc needs to be updated

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionUtil.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.runtime.endpoint.admission;
+
+import org.apache.commons.math3.distribution.NormalDistribution;
+
+public final class JobRuntimePredictionUtil {
```
fabiokung: minor: this class can probably be package protected

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSelectors.java
```diff
@@ -0,0 +1,121 @@
...
+    public static JobRuntimePredictionSelector aboveThreshold(double runtimeThresholdInSeconds,
+                                                              double sigmaThreshold,
+                                                              Map<String, String> selectionMetadata) {
+        return (jobDescriptor, predictions) -> {
+            SortedSet<JobRuntimePrediction> predictionsSet = predictions.getPredictions();
+            if (CollectionsExt.isNullOrEmpty(predictionsSet) || predictionsSet.size() < 2) {
+                return Optional.empty();
+            }
+
+            JobRuntimePrediction last = predictionsSet.last();
```
fabiokung: verify that the confidence here is `0.95` (and document the assumption), given we assume that when calculating `JobRuntimePredictionUtil.NORM_SIGMA`

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSelectors.java
```diff
@@ -0,0 +1,121 @@
...
+            if (last.getRuntimeInSeconds() > runtimeThresholdInSeconds) {
+                return Optional.empty();
+            }
+
+            JobRuntimePrediction first = predictionsSet.first();
```
fabiokung: verify that the confidence here is `0.05` (and document the assumption), given we assume that when calculating `JobRuntimePredictionUtil.NORM_SIGMA`

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSelectors.java
fabiokung: a few words about how `sigma` is our estimation of a stddev for the distribution of predicted runtimes

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSelectors.java
tbak: In this case we should also validate the low quantile.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionUtil.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private static final int LOW_QUANTILE = 5;
+    private static final int HIGH_QUANTILE = 95;
+
+    static final double NORM_SIGMA = computeNormSigma();
+
+    private JobRuntimePredictionUtil() {
+    }
+
+    private static double computeNormSigma() {
```
rostyboost: Documentation:
```
Estimate the standard deviation of a gaussian distribution given 2 quantiles. See https://www.johndcook.com/quantiles_parameters.pdf
```

titus-common/src/main/java/com/netflix/titus/common/util/DateTimeExt.java
```diff
@@ -58,6 +59,13 @@ public static String toUtcDateTimeString(long msSinceEpoch) {
         return ISO_UTC_DATE_TIME_FORMATTER.format(Instant.ofEpochMilli(msSinceEpoch)) + 'Z';
     }
 
+    public static String toLocalDateTimeString(long msSinceEpoch) {
+        if (msSinceEpoch == 0L) {
+            return null;
+        }
+        return ISO_LOCAL_DATE_TIME_FORMATTER.format(Instant.ofEpochMilli(msSinceEpoch)) + 'Z';
```
fabiokung: I think the `Z` is reserved for times in UTC, we shouldn't append it in this case

titus-common/src/main/java/com/netflix/titus/common/util/closeable/DefaultCloseableReference.java
```diff
@@ -0,0 +1,84 @@
...
+            ExceptionExt.silent(onSuccessCallback::run);
+        } catch (Throwable e) {
+            ExceptionExt.silent(() -> onErrorCallback.accept(e));
+            if (!swallowException) {
+                throw e;
+            }
+        }
+    }
+
+    private void doClose() {
```
fabiokung: is this missing a `closed.set(true)`?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSelectors.java
```diff
@@ -41,51 +61,146 @@ public static JobRuntimePredictionSelector never() {
...
             );
         };
     }
 
+    /**
+     * See {@link #aboveThreshold(double, double, Map)}.
+     */
+    public static JobRuntimePredictionSelector aboveThreshold(Config config, Map<String, String> selectionMetadata) {
+        double runtimeThresholdInSeconds = config.getDouble("runtimeThresholdInSeconds", 600.0);
+        double sigmaThreshold = config.getDouble("sigmaThreshold", 60.0);
```
fabiokung: default here seems to high, it should be probably 2, or 3 from experiments IIRC. @rostyboost confirm?

rostyboost: depends what we consider "default". @tbak wanted to make it so that by default the threshold on `sigma` is a no-op, and we're only thresholding on P95. makes sense to me. in this case I'd argue for the default to be `Double.MAX_VALUE`

titus-common/src/main/java/com/netflix/titus/common/util/closeable/DefaultCloseableReference.java
tbak: Good catch

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSelectors.java
tbak: `Double.MAX_VALUE` as no-op seems to be best here. Will do that

PR #657, opened 2019-09-09
IAM role validation with STS service

tbak:


titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/DataPlaneControllerCredentialsProvider.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public AWSCredentialsProvider get() {
+        String roleArn = configuration.getDataPlaneControllerRoleArn();
+        if (StringExt.isEmpty(roleArn)) {
+            return defaultCredentialsProvider;
+        }
+
+        String roleSessionName = configuration.getDataPlaneControllerRoleSessionName();
+        int roleSessionDurationSeconds = configuration.getDataPlaneControllerRoleSessionDurationSeconds();
```
andrew-leung: Why make local variables?

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamConnector.java
```diff
@@ -126,6 +153,46 @@ public void shutdown() {
...
+                        canAssumeCache.put(iamRoleName, true);
+                        connectorMetrics.success(IamConnectorMetrics.IamMethods.CanAgentAssume, startTime);
+                        return Mono.<Void>empty();
+                    })
+                    .onErrorMap(error -> {
+                        logger.debug("Error: {}", error.getMessage());
+                        connectorMetrics.failure(IamConnectorMetrics.IamMethods.CanAgentAssume, error, startTime);
+
+                        String errorCode = ((AWSSecurityTokenServiceException) error).getErrorCode();
+                        if ("AccessDenied".equals(errorCode)) {
```
andrew-leung: You could also consider caching failed validations to prevent frequent re-requests of bad IAMs.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/DataPlaneControllerCredentialsProvider.java
tbak: Readability only.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/iam/AwsIamConnector.java
tbak: Good point. I will add that

corindwyer: The downside to this is if the assume permissions are fixed and the value is still cached. If the ttl on the cache is low like 1 minute then it seems fine though.

PR #666, opened 2019-09-13
Add cluster membership to task relocation service

tbak:
Leader election callbacks are not handled yet.


titus-common-ext/k8s/src/main/java/com/netflix/titus/ext/k8s/clustermembership/connector/K8ClusterMembershipConnectorComponent.java
```diff
@@ -0,0 +1,86 @@
...
+import com.netflix.titus.common.util.StringExt;
+import io.kubernetes.client.ApiClient;
+import io.kubernetes.client.util.ClientBuilder;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
+
+@Configuration
+public class K8ClusterMembershipConnectorComponent {
```
fabiokung: minor nit: just noticed these classes are missing the `s` in their prefix (`K8 => K8s`). It would be probably better to rename them to `Kube*` to be consistent with other classes in the `master.mesos.kubeapiserver` package.

tbak: So rename `K8ClusterMembershipConnectorComponent` to `KubeClusterMembershipConnectorComponent`?

PR #669, opened 2019-09-18
Do not reduce opportunistic cpus on launchguard failures

fabiokung:
... and other failures considered to be "transient", meaning they should not be acted upon and are expected to go away in the near future.

Also refactor some of the scheduling failure classification code to make it more reusable.


titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailure.java
```diff
@@ -65,7 +67,21 @@
...
 
-        Unrecognized,
+        Unrecognized;
+
+        /**
+         * Failures that are expected to go away, and should not be acted upon
+         *
+         * @see DefaultSchedulingService
+         */
+        public static final Set<FailureKind> TRANSIENT = Sets.immutableEnumSet(WaitingForInUseIpAllocation, LaunchGuard);
```
andrew-leung: Just pointing out that, while highly unlikely, resolution of this condition may be up to user if they are doing something strange so how long this scenario may exist is variable (e.g., we're expecting the user to terminate the the task holding the contended IP Allocation).

titus-server-master/src/test/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScalerTest.java
```diff
@@ -185,7 +188,7 @@ public void testScaleUpForFailingTasks() {
         Map<TaskPlacementFailure.FailureKind, Map<String, List<TaskPlacementFailure>>> taskPlacementFailures = createTaskPlacementFailures(ImmutableMap.of(
                 AllAgentsFull, 10
         ), Tier.Flex);
-        when(schedulingService.getLastTaskPlacementFailures()).thenReturn(taskPlacementFailures);
+        doReturn(taskPlacementFailures).when(schedulingService).getLastTaskPlacementFailures();
```
andrew-leung: Was the `thenReturn()` type checking getting confused?

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailureClassifier.java
```diff
@@ -52,7 +52,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-class TaskPlacementFailureClassifier {
+class TaskPlacementFailureClassifier<T extends TaskRequest> {
```
tbak: I think we should use `TitusQueuableTask` as a base type in our code, so it is easier to add new behaviors if needed in the future.

titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
```diff
@@ -473,9 +474,13 @@ private Completable createResetOverrideStatusesCompletable(List<AgentInstance> r
...
 
-    private Map<Tier, Set<String>> getFailedTaskIds(Map<FailureKind, Map<String, List<TaskPlacementFailure>>> taskPlacementFailures, Set<FailureKind> ignoring) {
+    /**
+     * @param <T> generic helper "trick" to capture the wildcard. See https://docs.oracle.com/javase/tutorial/java/generics/capture.html
+     */
+    private <T> Map<Tier, Set<String>> getFailedTaskIdsByTier(Map<FailureKind, Map<T, List<TaskPlacementFailure>>> taskPlacementFailures,
+                                                              Set<FailureKind> ignoring) {
         Map<Tier, Set<String>> failedTaskIdsByTier = new HashMap<>();
-        for (Map<String, List<TaskPlacementFailure>> failuresByTaskId : taskPlacementFailures.values()) {
+        for (Map<T, List<TaskPlacementFailure>> failuresByTaskId : taskPlacementFailures.values()) {
```
tbak: How the keys are compared here? By object ref or value?

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailure.java
fabiokung: Makes sense. I think it is still right to call it a "transient" failure, and that we don't need to act on it, correct?

titus-server-master/src/test/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScalerTest.java
fabiokung: Yes, this is the way you are supposed to use Mockito with generic types/methods.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailureClassifier.java
fabiokung: that's why it is a generic type, it can be whatever we want. Note in `DefaultSchedulingService` I'm using `TaskPlacementFailureClassifier<V3QueueableTask>`. I think it makes sense to leave the more generic fenzo type only as a lower bound in the generic definition here, because as things are right now, it's the minimum type required for what this class is doing

titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
fabiokung: object ref, we have some other code indexing maps by `TaskRequests` (and Fenzo does it too internally). Since these are mutable objects that we keep references around (forever during its lifecycle), I think this is fine.

titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
```diff
@@ -108,12 +109,16 @@
     private static final Comparator<AgentInstanceGroup> PREFER_ACTIVE_INSTANCE_GROUP_COMPARATOR = Comparator.comparing(ig -> ig.getLifecycleStatus().getState());
     private static final Comparator<AgentInstanceGroup> PREFER_PHASED_OUT_INSTANCE_GROUP_COMPARATOR = PREFER_ACTIVE_INSTANCE_GROUP_COMPARATOR.reversed();
     private static final Set<String> IGNORED_HARD_CONSTRAINT_NAMES = asSet("machineid", "machinetype");
+    private static final Set<FailureKind> IGNORED_FAILURE_KINDS_WITH_LAUNCHGUARD = ImmutableSet.<FailureKind>builder()
+            .addAll(FailureKind.NEVER_TRIGGER_AUTOSCALING)
+            .add(FailureKind.LaunchGuard)
```
corindwyer: Can `LaunchGuard` be moved to `NEVER_TRIGGER_AUTOSCALING`?

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailure.java
andrew-leung: Yup. Just an FYI.

titus-server-master/src/main/java/com/netflix/titus/master/clusteroperations/ClusterAgentAutoScaler.java
fabiokung: No, because we still autoscale in some cases even when all failures are of the `LaunchGuard` kind, e.g.: when breaching task SLOs

PR #671, opened 2019-09-20
Add cluster membership metrics

tbak:


tbak:
@fabiokung I am merging this PR now. I will shortly push another one which adds elected leader activation mechanism.

PR #683, opened 2019-09-26
Fix circleci builds

amitaekbote:
Embedded postgres does not like being run as root, default user for the previous docker image was root, using circleci's jdk8 image. Since the previous cache was in the root directory using a new index for this cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,11 +15,9 @@ jobs:
             sudo apt update
             sudo apt install -qy ca-certificates git
       - checkout
-      - run: sudo chmod -R 777 /root/.gradle/caches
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 executors:
   java:
     docker:
-      - image: azul/zulu-openjdk:8
```
fabiokung: it would be better to continue using the zulu jdk for builds, to match what we will use when running this as close as possible

amitaekbote: that image runs as root by default, we will need to create a new user and login, this resembles running a test locally

PR #685, opened 2019-10-03
Enforce health check quota constraints only for jobs with size above a threshold (>=3)

tbak:
Health based quota limits are provided for larger jobs which can tolerate a few unhealthy
tasks.


titus-server-master/src/main/java/com/netflix/titus/master/eviction/service/quota/job/UnhealthyTasksLimitTracker.java
```diff
@@ -42,10 +42,17 @@
 
 public class UnhealthyTasksLimitTracker implements QuotaTracker {
 
+    /**
+     * Do not track health status of small jobs with size below the threshold.
+     */
+    private static final int JOB_SIZE_TRACKING_THRESHOLD = 3;
```
corindwyer: Can you change this to 2 so the only assumption on not highly available deployments is 1 instance?

tbak: Done

PR #687, opened 2019-10-07
prevent PolledMeter#monitorValue called twice

fabiokung:
It seems `@Activator` methods are being called twice (I will fix that in a different PR), which caused the `PolledMeter` gauge to be double counted. The default behavior of the Spectator client is to sum up multiple PolledMeters with the same metric Id.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeOpportunisticResourceProvider.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeOpportunisticResourceProvider.java
@@ -19,6 +19,7 @@ package com.netflix.titus.master.mesos.kubeapiserver;
 import java.util.Comparator;
 import java.util.Map;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
@@ -67,6 +68,7 @@ public class KubeOpportunisticResourceProvider implements OpportunisticCpuAvaila
     private final SharedIndexInformer<V1OpportunisticResource> informer;
     private final SharedInformerFactory informerFactory;
     private final Id cpuSupplyId;
+    private final ScheduledExecutorService metricsPollerExecutor;
 
     private static long currentOpportunisticCpuCount(KubeOpportunisticResourceProvider self) {
         return self.getOpportunisticCpus().values().stream()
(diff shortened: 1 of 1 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeOpportunisticResourceProvider.java
```diff
@@ -65,12 +67,28 @@
...
     @Inject
     public KubeOpportunisticResourceProvider(MesosConfiguration configuration, TitusRuntime titusRuntime) {
         this.titusRuntime = titusRuntime;
         cpuSupplyId = titusRuntime.getRegistry().createId("titusMaster.opportunistic.supply.cpu");
+
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat("kube-opportunistic-cpu-metrics-poller")
+                .setDaemon(true)
+                .build();
+        PolledMeter.using(titusRuntime.getRegistry())
```
tbak: We should release resources in the `shutdown` method.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeOpportunisticResourceProvider.java
```diff
@@ -176,19 +193,15 @@ private CodeInvariants codeInvariants() {
      */
     @Override
     public Map<String, OpportunisticCpuAvailability> getOpportunisticCpus() {
-        Map<String, OpportunisticCpuAvailability> cpuAvailability = informer.getIndexer().list().stream().collect(Collectors.toMap(
+        return informer.getIndexer().list().stream().collect(Collectors.toMap(
                 V1OpportunisticResource::getInstanceId,
                 resource -> new OpportunisticCpuAvailability(resource.getUid(), resource.getEnd(), resource.getCpus()),
                 BinaryOperator.maxBy(EXPIRES_AT_COMPARATOR)
         ));
-        lastCpuSupply.set(cpuAvailability.values().stream()
```
corindwyer: Since this method is being called many times for scheduling, I think it would be better if you create an [active gauge](http://netflix.github.io/spectator/en/latest/intro/gauge/#active-gauges) in the constructor and then set it in this method so that the min/max is correct. The downside to using the `PolledMeter` is that you don't know the rate at which it is being polled and the min/max values might not be accurate.

fabiokung: I did something similar to what you are suggesting before (each time this method gets called, update an `AtomicLong` with the current value, then install a `PolledMeter` on the `AtomicLong`).

The problem is that the rate in which this method gets called is orders of magnitude higher than the polling frequency (by `PolledMeter`). We know the polling frequency because it's configurable through `RegistryConfig#gaugePollingFrequency()` (default of 10s). Not  computing the metric inline with all calls (including the ones from scheduling hooks) ends up being way cheaper.

As far of min/max, I looked into the code of the active `Gauge`, and it doesn't track min/max for each step, it only keeps the last computed value. To have better control over we would need to use `distributionSummary` or `maxGauge`.

PR #690, opened 2019-10-08
Add scheduler REST/GRPC endpoints to TitusFederation server

tbak:
for now limited to `findLastSchedulingResult` operation only.

### DIFF
--- a/titus-server-federation/src/main/java/com/netflix/titus/federation/service/DefaultAggregatingSchedulerService.java
+++ b/titus-server-federation/src/main/java/com/netflix/titus/federation/service/DefaultAggregatingSchedulerService.java
@@ -50,10 +50,10 @@ public class DefaultAggregatingSchedulerService implements AggregatingSchedulerS
 
     @Override
     public Mono<SchedulingResultEvent> findLastSchedulingResult(String taskId) {
-        return findTaskInAllCells(taskId).map(CellResponse::getResult);
+        return findSchedulingResultInAllCells(taskId).map(CellResponse::getResult);
     }
 
-    private Mono<CellResponse<SchedulerServiceGrpc.SchedulerServiceStub, SchedulingResultEvent>> findTaskInAllCells(String taskId) {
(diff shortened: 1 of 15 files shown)


titus-server-federation/src/main/java/com/netflix/titus/federation/service/DefaultAggregatingSchedulerService.java
```diff
@@ -0,0 +1,78 @@
...
+        this.aggregatingClient = aggregatingClient;
+        this.callMetadataResolver = callMetadataResolver;
+    }
+
+    @Override
+    public Mono<SchedulingResultEvent> findLastSchedulingResult(String taskId) {
+        return findTaskInAllCells(taskId).map(CellResponse::getResult);
+    }
+
+    private Mono<CellResponse<SchedulerServiceGrpc.SchedulerServiceStub, SchedulingResultEvent>> findTaskInAllCells(String taskId) {
```
fabiokung: minor: copy and pasted the name? Probably should rename to `findSchedulingResultInAllCells`

titus-server-master/src/test/java/com/netflix/titus/master/integration/v3/job/JobFederationTest.java
```diff
@@ -106,4 +121,51 @@ public void testJobCreateRouting() {
...
+                .toList()
+                .toBlocking()
+                .first();
+
+        checkSchedulingResult(tasks.get(0));
+        checkSchedulingResult(tasks.get(1));
+    }
+
+    private void checkSchedulingResult(Task task) {
+        while (true) {
```
fabiokung: add timeout? Might be easier to use `Awaitability` like other tests do

tbak: I used `Awaitability` first, but then switched to an explicit loop. I purposefully did not set any timeouts, as it is difficult to estimate the time budget for all the blocking calls we have here. Instead we depend in this test on one global (and quite long) timeout.

tbak: On a second thought, having fine grain timeouts is better. Adding them...

PR #694, opened 2019-10-11
prevent @Activator methods called multiple times

fabiokung:
Guice will call `onProvision()` multiple times even for a @Singleton some times (apparently when it is being injected multiple times).

Make sure each unique instance (based on `Object#equals()`) is only registered once to be activated.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/guice/internal/ActivationProvisionListener.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/guice/internal/ActivationProvisionListener.java
@@ -20,7 +20,9 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


titus-common/src/main/java/com/netflix/titus/common/util/guice/internal/ActivationProvisionListener.java
```diff
@@ -44,7 +46,7 @@
 
     private static final Logger logger = LoggerFactory.getLogger(ActivationProvisionListener.class);
 
-    private final List<ServiceHolder> services = new CopyOnWriteArrayList<>();
+    private final Set<ServiceHolder> services = new CopyOnWriteArraySet<>();
```
tbak: We cannot use set here, as the activation order is important. Deactivation is executed in the reverse order.

fabiokung: good point, will fix

fabiokung: On a second thought, some `Sets` preserve insertion order, it is implementation specific. The one we are using here does preserve the insertion order, so we are fine.

Unfortunately in Java none of the `Collection` interfaces explicitly capture the intent of preserving insertion order. I will add a comment to make it more explicit.

PR #699, opened 2019-10-15
Emit metrics on cases where predictions are skipped

fabiokung:
With some coarse grained categorization for each failure case.

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSanitizerMetrics.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSanitizerMetrics.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionSanitizerMetrics.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package com.netflix.titus.runtime.endpoint.admission;
+
+import com.netflix.spectator.api.Registry;
+
+class JobRuntimePredictionSanitizerMetrics {
+
+    private enum Reason {
+        OPTED_OUT, PREDICTION_SERVICE_ERROR, INVALID_PREDICTIONS, NO_PREDICTION_SELECTED;
```
tbak: Java uses camel case convention for enum types. Why is it different here?

fabiokung: I looked a bit, and it seemed the consensus out there is to use capitalized names for enum entries, like constants. E.g.: https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html and https://stackoverflow.com/questions/3069743/coding-conventions-naming-enums/3069863

Our code base is mixed as well. Since these are more akin to constants (with stronger typing), I'd be in favor to keep them as capitalized. Maybe camel case makes more sense in other cases where we are using enums as case classes.

Do you have a strong opinion about this? I will merge and can change later if that's the case.

fabiokung: The Google style guide (which we don't use, but it's another reference) also favors capitalized names: https://google.github.io/styleguide/javaguide.html#s4.8.1-enum-classes

tbak: All caps is fine. Lets follow this convention.

PR #711, opened 2019-10-24
Prevent job size to go below 0

tbak:


titus-server-master/src/test/java/com/netflix/titus/master/jobmanager/service/integration/scenario/JobsScenarioBuilder.java
```diff
@@ -101,6 +103,7 @@ public JobsScenarioBuilder() {
 
         when(configuration.getActiveNotStartedTasksLimit()).thenReturn(ACTIVE_NOT_STARTED_TASKS_LIMIT);
         when(configuration.getConcurrentReconcilerStoreUpdateLimit()).thenReturn(CONCURRENT_STORE_UPDATE_LIMIT);
```
fabiokung: remove this one now that you installed another handler below?

PR #712, opened 2019-10-28
Add hedged request execution for iamConnector.canAgentAssume method

tbak:


titus-common/src/main/java/com/netflix/titus/common/util/rx/ReactorExt.java
```diff
@@ -128,6 +136,22 @@ private ReactorExt() {
         return new ReactorHeadTransformer<>(headSupplier);
     }
 
+    /**
+     * Creates multiple parallel subscriptions to the source observable. The first subscriptions happens immediately, while
```
corindwyer: subscriptions -> subscription

titus-common/src/main/java/com/netflix/titus/common/util/rx/ReactorExt.java
```diff
@@ -128,6 +136,22 @@ private ReactorExt() {
         return new ReactorHeadTransformer<>(headSupplier);
     }
 
+    /**
+     * Creates multiple parallel subscriptions to the source observable. The first subscription happens immediately, while
```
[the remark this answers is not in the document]
tbak: I am using here `observable` as a generic term. If you think it is misleading, I will be more explicit.

titus-common/src/main/java/com/netflix/titus/common/util/Evaluators.java
```diff
@@ -118,4 +122,39 @@ public static void times(int count, Consumer<Integer> task) {
...
     }
+
+    /**
+     * Remember last function execution result, and return the cached value on subsequent invocations, until the
+     * argument changes.
+     * <h1>
+     * For memoizing suppliers use {@link com.google.common.base.Suppliers#memoize(com.google.common.base.Supplier)}.
+     * </h1>
+     */
+    public static <T, R> Function<T, R> memoizeLast(Function<T, R> function) {
```
fabiokung: this seems very similar to `com.netflix.titus.common.util.cache.MemoizedFunction`, can we combine them?

tbak: I missed that one. Yes, it make sense to combine them. I will extend it to deal with failures as well, so we have two variants available (with error and without error caching).

titus-common/src/main/java/com/netflix/titus/common/util/rx/ReactorExt.java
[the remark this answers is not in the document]
andrew-leung: Not a big deal. Just thought it was slightly confusing.

PR #714, opened 2019-10-30
Remove the deprecated task migrator code

corindwyer:
### Description of the Change
Remove the deprecated task migrator code.


tbak:
@corindwyer When a new job is created, and the legacy migration policy is set we should either reject the request with an error, or silently rewrite it to some default value, like `SystemMigrationPolicy`.

corindwyer:
We are going to handle the `MigrationPolicy` changes in a different PR. This PR is just focused on removing the `TaskMigrator` code.

PR #715, opened 2019-10-30
Add back disruption budget sanitizer

tbak:


titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/DisruptionBudgetSanitizer.java
```diff
@@ -0,0 +1,139 @@
...
+        this.registry = titusRuntime.getRegistry();
+        this.nonCompliantId = registry.createId(METRICS_ROOT + "nonCompliant");
+    }
+
+    public JobDescriptor sanitize(JobDescriptor original) {
+        if (!DisruptionBudgetFunctions.isLegacyJobDescriptor(original)) {
+            return original;
+        }
+
+        if (!configuration.isEnabled()) {
```
corindwyer: I don't think this toggle is needed anymore as everything going into the system should be rewritten going forward.

PR #718, opened 2019-11-07
Shared Informer

corindwyer:
### Description of the Change

Use the shared informer pattern in the k8s integration.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
@@ -23,10 +23,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -36,7 +34,6 @@ import javax.inject.Singleton;
 
 import com.google.common.base.Strings;
 import com.google.gson.JsonSyntaxException;
-import com.google.gson.reflect.TypeToken;
 import com.google.inject.Injector;
 import com.netflix.fenzo.VirtualMachineLease;
 import com.netflix.fenzo.functions.Action1;
@@ -45,15 +42,14 @@ import com.netflix.titus.api.jobmanager.model.job.JobFunctions;
 import com.netflix.titus.api.jobmanager.model.job.Task;
 import com.netflix.titus.api.jobmanager.model.job.TaskState;
 import com.netflix.titus.api.jobmanager.service.V3JobOperations;
-import com.netflix.titus.common.annotation.Experimental;
 import com.netflix.titus.common.framework.scheduler.LocalScheduler;
 import com.netflix.titus.common.framework.scheduler.model.ScheduleDescriptor;
-import com.netflix.titus.common.network.client.ClientMetrics;
 import com.netflix.titus.common.runtime.TitusRuntime;
 import com.netflix.titus.common.util.ExecutorsExt;
 import com.netflix.titus.common.util.NetworkExt;
 import com.netflix.titus.common.util.RegExpExt;
 import com.netflix.titus.common.util.StringExt;
+import com.netflix.titus.common.util.guice.annotation.Deactivator;
 import com.netflix.titus.common.util.time.Clock;
 import com.netflix.titus.master.mesos.ContainerEvent;
 import com.netflix.titus.master.mesos.LeaseRescindedEvent;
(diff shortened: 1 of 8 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -219,20 +199,124 @@ public void enterActiveMode() {
...
+                        null,
+                        null,
+                        params.resourceVersion,
+                        params.timeoutSeconds,
+                        params.watch,
+                        null,
+                        null),
+                V1Node.class,
+                V1NodeList.class,
+                RESYNC_PERIOD_MS
```
tbak: Dynamic property?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -219,20 +199,124 @@ public void enterActiveMode() {
...
     public void launchTasks(List<Protos.TaskInfo> requests, List<VirtualMachineLease> leases) {
         for (Protos.TaskInfo request : requests) {
-            long startTimeMs = clock.wallTime();
             try {
                 V1Pod v1Pod = taskInfoToPod(request);
                 logger.info("creating pod: {}", v1Pod);
-                normalApi.createNamespacedPod(KUBERNETES_NAMESPACE, v1Pod, null, null, null);
-                podsClientMetrics.incrementOnSuccess(POST, PODS, STATUS_200);
-                podsClientMetrics.registerOnSuccessLatency(POST, Duration.ofMillis(clock.wallTime() - startTimeMs));
+                api.createNamespacedPod(KUBERNETES_NAMESPACE, v1Pod, null, null, null);
```
tbak: Are these metrics replaced with something else?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -608,16 +578,15 @@ private void publishContainerEvent(String taskId, TaskState taskState, String re
     }
 
     private void reconcileNodesAndPods() {
-        if (!mesosConfiguration.isReconcilerEnabled()) {
+        if (!mesosConfiguration.isReconcilerEnabled() || !nodeInformer.hasSynced() || !podInformer.hasSynced()) {
```
tbak: Do we need `mesosConfiguration.isReconcilerEnabled()` check here? It is Mesos specific reconciliation logic.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
```diff
@@ -0,0 +1,79 @@
...
+
+        ApiClient client = Config.fromUrl(serverUrl);
+        client.getHttpClient().setReadTimeout(readTimeoutMs, TimeUnit.MILLISECONDS);
+        client.getHttpClient().interceptors().add(metricsInterceptor);
+        return client;
+    }
+
+    public static SharedInformerFactory createSharedInformerFactory(String threadNamePrefix, ApiClient apiClient) {
+        AtomicLong nextThreadNum = new AtomicLong(0);
+        return new SharedInformerFactory(apiClient, Executors.newCachedThreadPool(runnable -> {
```
tbak: Is this executor shutdown in the shutdown process or leaked? In the latter case, it may affect unit tests.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
corindwyer: yes, I instrumented the `OkHttp` client so all the calls has metrics.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
corindwyer: It is within Mesos because this is an integration that translates K8s into Mesos concepts.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
corindwyer: it will be stopped in the `shutdown` method.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeOpportunisticResourceProvider.java
```diff
@@ -92,22 +95,20 @@ public KubeOpportunisticResourceProvider(MesosConfiguration configuration, Titus
...
+                KubeOpportunisticResourceProvider.class.getSimpleName() + "-",
+                apiClient
+        );
         // TODO(fabio): enhance the kube client to support custom JSON deserialization options
         informer = informerFactory.sharedIndexInformerFor(
                 this::listOpportunisticResourcesCall,
                 V1OpportunisticResource.class,
                 V1OpportunisticResourceList.class,
-                configuration.getKubeOpportunisticRefreshIntervalMs()
+                refreshIntervalMs
```
fabiokung: should probably use a slightly larger interval here, just to avoid killing the connection in the middle of it trying to reconcile

corindwyer: changed as it was being overriden

PR #719, opened 2019-11-08
Add load shedding capability to TitusMaster

tbak:


titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
```diff
@@ -0,0 +1,146 @@
...
+        );
+    }
+
+    @Override
+    public AdmissionControllerResponse apply(AdmissionControllerRequest request) {
+        return findTokenBucket(request).map(this::consume).orElse(DEFAULT_OK);
+    }
+
+    private Optional<TokenBucketInstance> findTokenBucket(AdmissionControllerRequest request) {
+        return tokenBucketConfigurations.stream()
```
fabiokung: I don't particularly love the `O(N)` here on every single request, I think it will be fine as the number of `tokenBucketConfigurations` will be very small, but we should probably monitor latency here to make sure we keep an eye on it

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
```diff
@@ -0,0 +1,146 @@
...
+                            ? tokenBucketConfiguration.getCallerPatternString()
+                            : request.getCallerId();
+
+                    return bucketsById.get(id, i -> new TokenBucketInstance(tokenBucketConfiguration));
+                });
+    }
+
+    private boolean matches(TokenBucketConfiguration configuration, AdmissionControllerRequest request) {
+        try {
+            if (!configuration.getCallerPattern().matcher(request.getCallerId()).matches()) {
```
fabiokung: minor nit: simplify the if/else chain to a single logic expression - `return callerPattern.matches() && endpointPatter.matches()`

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
```diff
@@ -0,0 +1,146 @@
...
+        return findTokenBucket(request).map(this::consume).orElse(DEFAULT_OK);
+    }
+
+    private Optional<TokenBucketInstance> findTokenBucket(AdmissionControllerRequest request) {
+        return tokenBucketConfigurations.stream()
+                .filter(tokenBucketConfiguration -> matches(tokenBucketConfiguration, request))
+                .findFirst()
+                .map(tokenBucketConfiguration -> {
+                    // If shared, use pattern name as key, otherwise use caller id
+                    String id = tokenBucketConfiguration.isShared()
```
fabiokung: What if the same callerId has different configurations for different endpoints, don't we need a different `TokenBucketInstance` (id) for each `(callerPatternString, endpointPatternString)` pair?

fabiokung: Thinking more about it, we also need `tokenBucketConfiguration.getName()` as part of the id here

titus-common/src/test/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionControllerTest.java
```diff
@@ -0,0 +1,111 @@
...
+import com.netflix.titus.common.runtime.TitusRuntimes;
+import com.netflix.titus.common.util.loadshedding.AdmissionControllerRequest;
+import com.netflix.titus.common.util.loadshedding.AdmissionControllerResponse;
+import org.junit.Test;
+
+import static com.netflix.titus.common.util.loadshedding.tokenbucket.TokenBucketTestConfigurations.DEFAULT_SHARED_CONFIGURATION;
+import static com.netflix.titus.common.util.loadshedding.tokenbucket.TokenBucketTestConfigurations.NOT_SHARED_CONFIGURATION;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TokenBucketAdmissionControllerTest {
```
fabiokung: add a test for a single `callerId` (or `callerIdPattern`) with different configs for different endpoints

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/SpectatorAdmissionController.java
```diff
@@ -0,0 +1,61 @@
...
+package com.netflix.titus.common.util.loadshedding;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+import com.netflix.spectator.api.Registry;
+import com.netflix.titus.common.runtime.TitusRuntime;
+
+@Singleton
+public class SpectatorAdmissionController implements AdmissionController {
```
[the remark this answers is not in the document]
tbak: Currently this interface is `Function<AdmissionControllerRequest, AdmissionControllerResponse>`. I do not expect more interface methods to be added here.

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/ArchaiusTokenBucketAdmissionController.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public AdmissionControllerResponse apply(AdmissionControllerRequest request) {
+        return delegate.apply(request);
+    }
+
+    private void reload(ExecutionContext context) {
+        List<TokenBucketConfiguration> current = configurationParser.parse();
+        if (current != activeConfiguration) {
```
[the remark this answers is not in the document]
tbak: Changing to `equals`

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
tbak: My assumption is that we should have <10 configurations total. If we need more and it gets more complex, we should optimize this code.

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
tbak: I did this on purpose to improve readability. The evaluated conditions are very long.

tbak: Building both matchers before the condition evaluation may result in unnecessary computation if the first condition fails.

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
tbak: We need use callerId / endpoint name as bucket id.
`tokenBucketConfiguration.getName()` is not important

tbak:
We agreed in the offline discussion with @fabiokung that we do not need different buckets per endpoint. To make it more explicit I am changing `TokenBucketConfiguration#shared` field to `TokenBucketConfiguration#sharedByCallers`.

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/ArchaiusTokenBucketAdmissionController.java
```diff
@@ -0,0 +1,115 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A wrapper around {@link TokenBucketAdmissionController} which dynamically loads configuration from Archaius.
+ * When configuration changes, the current {@link TokenBucketAdmissionController} instance is discarded, a new one
+ * is created. As a result token buckets state is reset, which is fine as long as changes do not happen too
+ * frequently.
+ */
+public class ArchaiusTokenBucketAdmissionController implements AdmissionController, Closeable {
```
corindwyer: I think it would be better to make this a generic configuration loader implementation and put the archaius part  behind a Configuration abstraction as there isn't really anything Archaius specific in this class.

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(TokenBucketAdmissionController.class);
+
+    private static final AdmissionControllerResponse DEFAULT_OK = AdmissionControllerResponse.newBuilder()
+            .withAllowed(true)
+            .withReasonMessage("Rate limits not configured")
+            .build();
+
+    private static final int MAX_CACHE_SIZE = 10_000;
+    private static final Duration CACHE_ITEM_TIMEOUT = Duration.ofSeconds(60);
```
corindwyer: is this timeout needed considering there is a max cache size?

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
corindwyer: I would have rather seen a generated key + lookup to make this more efficient but it isn't needed immediately.

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
tbak: Bumping up timeout to 5min. If the cache gets full, some items will be removed. From caffeine documentation:
```   
/* Specifies the maximum number of entries the cache may contain. Note that the cache <b>may evict
   * an entry before this limit is exceeded or temporarily exceed the threshold while evicting</b>.
   * As the cache size grows close to the maximum, the cache evicts entries that are less likely to
   * be used again. For example, the cache may evict an entry because it hasn't been used recently
   * or very often.
 */
public Caffeine<K, V> maximumSize(@Nonnegative long maximumSize) {
```

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/tokenbucket/TokenBucketAdmissionController.java
tbak: Giving up under the peer pressure. I am implementing the caching now.

titus-common/src/test/java/com/netflix/titus/common/util/loadshedding/tokenbucket/ConfigurableTokenBucketAdmissionControllerTest.java
```diff
@@ -0,0 +1,96 @@
...
+import com.netflix.titus.common.util.loadshedding.AdmissionControllerRequest;
+import com.netflix.titus.common.util.loadshedding.AdmissionControllerResponse;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import static com.jayway.awaitility.Awaitility.await;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ConfigurableTokenBucketAdmissionControllerTest {
```
fabiokung: add a test for invalid configurations (keep the previous last seen valid one)

tbak: In `ArchaiusTokenBucketAdmissionConfigurationParser` we skip all invalid configurations (there are log entries for each failure).
This way a single bad configuration entry will not prevent us from loading other entries, for example on startup.

PR #721, opened 2019-11-14
Limit threads used by ObserveJobs and add instrumentation

fabiokung:
TitusMaster servers can be DoS'd with very few `ObserveJobs` calls. Currently, each `ObserveJobs` stream will consume a new thread in a TitusMaster server. Additionally, gRPC will by default use a different thread pool for each `Channel`. In a typical server configuration, the maximum number of threads is in the dozens of thousands.

These changes considerably improve (by 10x or more) our ability to handle server streaming calls:

* use a single thread pool for all gRPC tasks, instead of a separate thread pool per channel (default with gRPC)
* limit the number of threads used for server streaming (ObserveJobs) tasks, instead of using an unbounded thread pool and potentially spawning a new thread per request while stream snapshots are being generated (an expensive operation). This will cause calls to sit on a queue when snapshots can't be generated fast enough, i.e. requests are queued when we are overloaded
* metrics for thread pool sizes, busy thread counts, and task counts on gRPC thread pools

I did some load testing with an empty Titus stack, and after these changes we were able to sustain ~300 `ObserveJobs` calls/sec with little additional threads from what is pre-allocated in the fixed size thread pools. 100k concurrent streams were no problem and consumed not more than a few hundred threads. This is significantly better than what we saw in the past.

The only remaining problem is with `rx.unsubscribe()`, called when gRPC streams are canceled. That is sadly blocking inside the RxJava stack, and putting pressure on gRPC thread pools and the Rx Computation Scheduler. Canceling 100k `ObserveJobs` streams consumed ~4k threads for a few seconds and caused the hiccup sensor to complain about threads blocked in the Rx Computation Scheduler.

Unsubscription being expensive can still be a problem, and we will need another solution for it. But these changes put us in a much better shape.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/ExecutorsExt.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/ExecutorsExt.java
@@ -18,11 +18,22 @@ package com.netflix.titus.common.util;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.netflix.spectator.api.Registry;
+import com.netflix.spectator.api.patterns.ThreadPoolMonitor;
 
 /**
  * Factory methods and utilities for {@link java.util.concurrent.ExecutorService executors}.
  */
 public final class ExecutorsExt {
+    private ExecutorsExt() {
+    }
 
     public static ExecutorService namedSingleThreadExecutor(String name) {
         return Executors.newSingleThreadExecutor(runnable -> {
@@ -30,6 +41,46 @@ public final class ExecutorsExt {
             thread.setDaemon(true);
             return thread;
         });
+    }
+
+    /**
+     * Unbounded elastic thread pool that grows and shrinks as necessary.
+     */
+    public static ExecutorService instrumentedCachedThreadPool(Registry registry, String name) {
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat(name + "-%d")
+                .setDaemon(true)
+                .build();
+        // similar to Executors.newCachedThreadPool(), but explicitly use the concrete ThreadPoolExecutor type to ensure
+        // it can be instrumented by Spectator
+        ThreadPoolExecutor executor = new ThreadPoolExecutor(
+                0, Integer.MAX_VALUE,
+                60L, TimeUnit.SECONDS,
+                new SynchronousQueue<>(),
+                threadFactory
+        );
+        ThreadPoolMonitor.attach(registry, executor, name);
+        return executor;
+    }
+
+    /**
+     * Fixed size thread pool that pre-allocates all threads. When no more threads are available, requests will block.
+     */
+    public static ExecutorService instrumentedFixedSizeThreadPool(Registry registry, String name, int size) {
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat(name + "-%d")
+                .setDaemon(true)
+                .build();
+        // similar to Executors.newFixedSizeThreadPool(), but explicitly use the concrete ThreadPoolExecutor type
+        // to ensure it can be instrumented by Spectator
+        ThreadPoolExecutor executor = new ThreadPoolExecutor(
+                size, size,
+                0L, TimeUnit.MILLISECONDS,
+                new LinkedBlockingQueue<>(),
+                threadFactory
+        );
+        ThreadPoolMonitor.attach(registry, executor, name);
+        return executor;
     }
 
 }
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
@@ -56,6 +56,7 @@ import com.netflix.titus.common.model.sanitizer.EntitySanitizer;
 import com.netflix.titus.common.model.sanitizer.ValidationError;
 import com.netflix.titus.common.runtime.TitusRuntime;
 import com.netflix.titus.common.util.CollectionsExt;
+import com.netflix.titus.common.util.ExecutorsExt;
 import com.netflix.titus.common.util.ProtobufExt;
 import com.netflix.titus.common.util.archaius2.ObjectConfigurationResolver;
 import com.netflix.titus.common.util.rx.ObservableExt;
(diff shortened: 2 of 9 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/common/grpc/AbstractTitusGrpcServer.java
```diff
@@ -71,7 +77,9 @@ public int getPort() {
     @PostConstruct
     public void start() {
         if (!started.getAndSet(true)) {
-            ServerBuilder serverBuilder = configure(ServerBuilder.forPort(configuration.getPort()));
+            ServerBuilder<?> initial = ServerBuilder.forPort(port);
+            initial.executor(ExecutorsExt.instrumentedCachedThreadPool(runtime.getRegistry(), "grpcCallbackExecutor"));
```
tbak: How are these executors shutdown?

tbak:
@fabiokung What do you mean by channel in this sentence:
`use a single thread pool for all gRPC tasks, instead of a separate thread pool per channel (default with gRPC)`

Is this service endpoint (jobManagement, agetManagement, etc)?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/common/grpc/AbstractTitusGrpcServer.java
fabiokung: Never, only when the JVM shuts down. The executor will live with the gRPC server (one per gRPC server).

fabiokung:
> @fabiokung What do you mean by channel in this sentence:
> `use a single thread pool for all gRPC tasks, instead of a separate thread pool per channel (default with gRPC)`
> 
> Is this service endpoint (jobManagement, agetManagement, etc)?

No, I really mean a gRPC server `ManagedChannel`. Every time a new channel is open in a default gRPC server, it will use a new thread pool for all RPC calls, so we always end up using at least one thread per `ManagedChannel` in the server. With a single shared pool for all channels, we can reuse the same Threads across all async grpc `Call`s (or executor tasks).

fabiokung:
@tbak see @carl-mastrangelo's comment here: https://stackoverflow.com/a/42422045

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/common/grpc/AbstractTitusGrpcServer.java
carl-mastrangelo: Hi, I would recommend that you shut down the gRPC server, even if you don't bother with the executor.  You can add it as a JVM shutdown hook.    This allows in flight RPCs to finish gracefully while turning away new client.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
```diff
@@ -171,6 +174,8 @@ public DefaultJobManagementServiceGrpc(GrpcMasterEndpointConfiguration configura
         this.authorizationService = authorizationService;
         this.titusRuntime = titusRuntime;
         this.schedulingService = schedulingService;
+        this.observeJobsScheduler = Schedulers.from(ExecutorsExt.instrumentedFixedSizeThreadPool(
```
carl-mastrangelo: Personal preference, but I would recommend against using a fixed size executor here\*.   Depending on how spiky your traffic is, a better way to bound the traffic is to keep track of the number of active RPCs, and fail new RPCs at your application layer.  (by keeping track of latency, deadline expirations, etc.)    Feel free to keep it as is, but just wanted to make aware that fixed thread pools have some risky side effects that are hard to debug.

\* I know I said otherwise on the S/O answer, but there is more nuance needed for an accurate answer (but which doesn't fit well into the S/O format).

titus-common/src/main/java/com/netflix/titus/common/util/ExecutorsExt.java
```diff
@@ -32,4 +43,44 @@ public static ExecutorService namedSingleThreadExecutor(String name) {
...
+     * Fixed size thread pool that pre-allocates all threads. When no more threads are available, requests will block.
+     */
+    public static ExecutorService instrumentedFixedSizeThreadPool(Registry registry, String name, int size) {
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat(name + "-%d")
+                .setDaemon(true)
+                .build();
+        // similar to Executors.newFixedSizeThreadPool(), but explicitly use the concrete ThreadPoolExecutor type
+        // to ensure it can be instrumented by Spectator
+        ThreadPoolExecutor executor = new ThreadPoolExecutor(
```
carl-mastrangelo: I see that this is being done to add monitoring, but I would highly recommend using `ForkJoinPool`, which degrades gracefully under load.  When doing a large number of RPCs (Like millons/s), the blocking Queue lock contention begins to serialize the RPC callbacks.  If there is any way to update ThreadPoolMonitor to work with FJP, it would be ideal long term.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/common/grpc/AbstractTitusGrpcServer.java
tbak: We run multiple server instances (on ephemeral ports) when running integration tests. Servers are shutdown properly, so it should be easy to add a shutdown logic to the executor pool as well.

fabiokung: sounds good, I will look into tearing down the executor when the grpc server is shutting down in a follow-up PR

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
fabiokung: Makes sense. There's a lot of nuance here, and even with a concurrency limiter at the application code level (RPC handlers) we would still want to limit the size of this thread pool for isolation purposes, and to protect other components running in the same JVM. This was the intent here, we will look into adding more protection (concurrency limits) for RPCs in the future.

titus-common/src/main/java/com/netflix/titus/common/util/ExecutorsExt.java
fabiokung: Yeah, I agree `ForkJoinPool` could be better here. We are not super concerned with the throughput yet, this is not a grpc server where we need to squeeze every possible RPS, so I'm leaning into more instrumentation for now.

Agree it would be nice to have support for an instrumented `ForkJoinPool` in Spectator, we will revisit this as needed in the future.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/common/grpc/AbstractTitusGrpcServer.java
fabiokung: @tbak I read the code again, and I was wrong and misinterpreted the code. You were right, by default it seems each gRPC *server* gets a single cached thread pool. gRPC *clients* get a thread pool per `ManagedChannel`. I mixed up the two usages of the `callbackExecutor` code.

In any case I still think it is better to replace the default server callback executor, as we will get better instrumentation with our own. I will make a change to avoid leaking executors when we spawn multiple grpc servers in integration tests in a follow-up PR.

PR #724, opened 2019-11-19
add load shedding to federation grpc requests

fabiokung:
Protection for REST endpoints will come at a later date.


tbak:
Looks good, but is missing REST API integration. For completeness we should cover that as well, as a single rouge client can still generate excessive load down to TitusMaster.

titus-server-federation/src/main/java/com/netflix/titus/federation/endpoint/EndpointConfiguration.java
```diff
@@ -34,4 +34,7 @@
      */
     @DefaultValue("30000")
     long getGrpcServerShutdownTimeoutMs();
+
+    @DefaultValue("true")
+    boolean isAdmissionControllerEnabled();
```
tbak: Add REST API integration.

fabiokung:
@tbak I mention the REST API in the PR description.

tbak:
@fabiokung Sorry, missed that.

PR #729, opened 2019-11-27
tear down thread pools when grpc servers shutdown

fabiokung:
Follow-up from #721

### DIFF
--- a/titus-server-gateway/src/main/java/com/netflix/titus/gateway/endpoint/v3/grpc/TitusGatewayGrpcServer.java
+++ b/titus-server-gateway/src/main/java/com/netflix/titus/gateway/endpoint/v3/grpc/TitusGatewayGrpcServer.java
@@ -18,6 +18,7 @@ package com.netflix.titus.gateway.endpoint.v3.grpc;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


titus-server-gateway/src/main/java/com/netflix/titus/gateway/endpoint/v3/grpc/TitusGatewayGrpcServer.java
```diff
@@ -112,67 +115,69 @@ public int getPort() {
...
+            throw new RuntimeException(e);
+        }
+        LOG.info("Started gRPC server on port {}.", port);
     }
 
     @PreDestroy
     public void shutdown() {
-        if (!server.isShutdown()) {
-            long timeoutMs = config.getShutdownTimeoutMs();
+        if (server.isShutdown()) {
```
corindwyer: Can this code be extracted into a single place instead of it being duplicated in each server?

fabiokung: it is already extract in `AbstractTitusGrpcServer`, but these three older server classes (Master, Gateway, Federation) are not using it yet.

fabiokung: (and I didn't want to make that change in this PR)

PR #731, opened 2019-12-04
Add transactionId to `ManyReconciler` events + race condition fix

tbak:


titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/internal/DefaultManyReconciler.java
```diff
@@ -258,26 +258,37 @@ private void connectEventListeners() {
...
-
         for (EventListenerHolder holder; (holder = eventListenerHolders.poll()) != null; ) {
             FluxSink<List<SimpleReconcilerEvent<DATA>>> sink = holder.getSink();
-            sink.next(allState);
-            Disposable disposable = eventStream.subscribe(
-                    sink::next,
-                    sink::error
-            );
+
+            // We build snapshot only after the subscription to 'eventStream' happens, otherwise we might loose events
```
corindwyer: loose -> lose

tbak:
@fabiokung Good point. Made me think again about it. I have changed transaction id from string to integer, with better documentation about semantic.

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/SimpleReconcilerEvent.java
```diff
@@ -29,9 +29,15 @@
     private final Kind kind;
     private final String id;
     private final DATA data;
-    private final String transactionId;
 
-    public SimpleReconcilerEvent(Kind kind, String id, DATA data, String transactionId) {
+    /**
+     * A unique identifier associated with each reconciler action and data element version. Consecutive transaction ids
```
fabiokung: I think it is important to note that this will be only unique within a session (subscription) and will be possibly reset across failovers (and thus reconnections).

fabiokung: ... which brings the question of how are clients supposed to handle transaction ids that are reset?

fabiokung: ... or is this only ever supposed to be used within a single JVM process, and no guarantees about the state of transaction ids will be provided when processes are bounced?

tbak: That is correct. Restarting JVM resets it. In case of GRPC stream it will also terminate the stream, and subscriber must be aware of that. It must be documented in GRPC IDL that uses this information.
However this is just a library, without persistence notion, so i do not think we should talk about failovers here.

PR #732, opened 2019-12-04
Set job accepted timestamp in agent proto

andrew-leung:
Will depend on: https://github.com/Netflix/titus-api-definitions/pull/118 being merged.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
@@ -276,7 +276,7 @@ public class DefaultV3TaskInfoFactory implements TaskInfoFactory<Protos.TaskInfo
     }
 
     private void setJobAcceptedTimestamp(ContainerInfo.Builder containerInfoBuilder, Job<?> job) {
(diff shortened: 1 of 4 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/DefaultV3TaskInfoFactory.java
```diff
@@ -230,6 +232,9 @@ public DefaultV3TaskInfoFactory(MasterConfiguration masterConfiguration,
         // Configure IP address allocation
         setSignedAddressAllocation(containerInfoBuilder, task, containerResources);
 
+        // Configure job accepted timestamp
+        setJobAcceptedTimestamp(containerInfoBuilder, job);
```
tbak: It is not documented here nor in the IDL, why we need this timestamp.

andrew-leung: We pass the job accept timestamp to agents which then passes it along to Metatron on container launch. Metatron uses this field as part of its evaluation of whether to grant credentials.

This address some issues we've had with long running jobs where the key used to sign the job has since expired and new tasks cannot acquire credentials.

PR #733, opened 2019-12-05
Minor improvements

corindwyer:
### Description of the Change
* Add ability to randomize machines for scheduling and keep agent quality tracker data longer to prevent scheduling to the same machine.
* Handle missed event case in K8s integration since we are converting from a level based system to an edge based system.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
@@ -38,9 +38,11 @@ import com.google.inject.Injector;
 import com.netflix.fenzo.VirtualMachineLease;
 import com.netflix.fenzo.functions.Action1;
 import com.netflix.fenzo.plugins.VMLeaseObject;
+import com.netflix.titus.api.jobmanager.model.job.Job;
 import com.netflix.titus.api.jobmanager.model.job.JobFunctions;
 import com.netflix.titus.api.jobmanager.model.job.Task;
 import com.netflix.titus.api.jobmanager.model.job.TaskState;
+import com.netflix.titus.api.jobmanager.model.job.TaskStatus;
 import com.netflix.titus.api.jobmanager.service.V3JobOperations;
 import com.netflix.titus.common.framework.scheduler.LocalScheduler;
 import com.netflix.titus.common.framework.scheduler.model.ScheduleDescriptor;
@@ -51,6 +53,7 @@ import com.netflix.titus.common.util.RegExpExt;
 import com.netflix.titus.common.util.StringExt;
 import com.netflix.titus.common.util.guice.annotation.Deactivator;
 import com.netflix.titus.common.util.time.Clock;
+import com.netflix.titus.common.util.tuple.Pair;
 import com.netflix.titus.master.mesos.ContainerEvent;
 import com.netflix.titus.master.mesos.LeaseRescindedEvent;
 import com.netflix.titus.master.mesos.MesosConfiguration;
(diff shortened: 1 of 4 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -519,37 +528,97 @@ private void podUpdated(V1Pod pod) {
...
+                state = Finished;
+                reasonCode = REASON_FAILED;
+            } else {
+                titusRuntime.getCodeInvariants().inconsistent("Pod: %s has unknown phase mapping: %s", podName, phase);
+                return;
+            }
+
+            // if the new state is Finished and the pod had a container that ran then make sure that the StartInitiated
+            // and Started states were populated in the task state machine due to apiserver sending the latest event.
+            if (state == Finished) {
```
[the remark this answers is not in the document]
fabiokung: I'd suggest calling it something like: `fillInMissingStateTransitions(...)`, and adding the comments above as its javadoc

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -519,37 +528,97 @@ private void podUpdated(V1Pod pod) {
...
         }
     }
 
+    private Optional<V1ContainerStateTerminated> findTerminatedContainerStatus(V1Pod pod) {
+        List<V1ContainerStatus> containerStatuses = pod.getStatus().getContainerStatuses();
+        if (containerStatuses != null) {
+            for (V1ContainerStatus status : containerStatuses) {
+                V1ContainerState state = status.getState();
+                if (state != null) {
+                    return Optional.ofNullable(state.getTerminated());
```
fabiokung: minor: never null, so you can use `Optional.of()`

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -519,37 +528,97 @@ private void podUpdated(V1Pod pod) {
...
+        return Optional.empty();
+    }
+
     private void publishContainerEvent(String taskId, TaskState taskState, String reasonCode, String reasonMessage,
                                        Optional<TitusExecutorDetails> executorDetails) {
+        publishContainerEvent(taskId, taskState, reasonCode, reasonMessage, executorDetails, 0);
+    }
 
+    private void publishContainerEvent(String taskId, TaskState taskState, String reasonCode, String reasonMessage,
+                                       Optional<TitusExecutorDetails> executorDetails, long timestamp) {
```
fabiokung: minor: use `Optional<Long>` to avoid magic numbers with special meaning

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/SchedulerConfiguration.java
```diff
@@ -57,6 +57,14 @@
     @DefaultValue("true")
     boolean isSchedulingMachinesFilterEnabled();
 
+    /**
+     * An option to enable shuffle the machines list from the machines filter.
+     *
+     * @return whether to enable filtering of machines before scheduling iteration.
```
fabiokung: minor: copy-and-pasta describing the wrong return

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
corindwyer: It is possible that it is null.

fabiokung: my bad, I misread the null check just above

PR #734, opened 2019-12-06
loadbalancer: store state for targets being managed

fabiokung:
Store information about all load balancer targets being managed by a particular titusmaster instance, so multiple stacks can share the same load balancer (AWS TargetGroup). When reconciling targets that should be in a load balancer, a titusmaster instance now only ever tries to deregister something that was registered by itself in the past. Targets added by someone else are left untouched.

I'm opening this PR early for reviews. It is not ready to be merged yet, as I'm still running benchmarks to ensure the impact of conditional deletes (`DELETE ... IF`) I'm using here for optimistic concurrency control will not be significant, and will not harm job/task management operations on C*.

Conditional deletes use lightweight transactions on C* and will require coordinators to run Paxos, which has a non-negligible cost. It should be fine for the out-of-band garbage collection I'm doing here, but I want to get some confidence it will not hurt other operations. I'm hoping I can use this feature to avoid having to serialize **all** store operations with the reconciliation loops.


titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
```diff
@@ -274,9 +322,69 @@ public int getNumLoadBalancersForJob(String jobId) {
...
+        Set<ValidationError> violations = entitySanitizer.validate(target);
+        if (!violations.isEmpty()) {
+            if (configuration.isFailOnInconsistentLoadBalancerData()) {
+                return Mono.error(LoadBalancerStoreException.badData(target, violations));
+            }
+            logger.warn("Ignoring bad target record of {} due to validation constraint violations: violations={}", target, violations);
+            return Mono.empty();
+        }
+
+        BoundStatement statement = insertTarget.bind(
```
tbak: If `addOrUpdateTargets` would retry, this statement would be reused which is probably incorrect. Use `Mono.defer(...)`?

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
```diff
@@ -274,9 +322,69 @@ public int getNumLoadBalancersForJob(String jobId) {
...
+                target.getIpAddress(),
+                target.getLoadBalancerTarget().getTaskId(),
+                target.getState().name()
+        );
+        return ReactorExt.toMono(storeHelper.execute(statement).toCompletable());
+    }
+
+    @Override
+    public Mono<Void> removeDeregisteredTargets(Collection<LoadBalancerTarget> toRemove) {
+        List<Mono<Void>> deleteOperations = toRemove.stream()
```
tbak: Like above, bind happens too early.

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
tbak: According to the documentation:
```This class is not thread-safe. Do not share instances among requests that will execute concurrently (e.g. requests run from separate application threads, but also separate executeAsync calls, even if they're triggered from the same thread).```
If we can guarantee strict non-concurrency, we could leave it like that, but it is probably easier to create a new instance for each attempt.

File: titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java [hunk 3]
```diff
@@ -274,9 +322,69 @@ public int getNumLoadBalancersForJob(String jobId) {
...
+                .ignoreElements()
+                .doOnSubscribe(ignored -> logger.debug("Removing targets {}", toRemove));
+    }
+
+    @Override
+    public Flux<LoadBalancerTargetState> getLoadBalancerTargets(String loadBalancerId) {
+        Statement selectStmt = selectTargetsForLoadBalancer.bind(loadBalancerId).setFetchSize(FETCH_SIZE);
+        return ReactorExt.toFlux(storeHelper.execute(selectStmt))
+                .timeout(Duration.ofMillis(FETCH_TIMEOUT_MS))
+                .next()
```
tbak: As above

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
fabiokung: Good catch. I will fix.

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
```diff
@@ -152,17 +206,11 @@ public void init() {
...
-    }
-
-    @Override
-    public Observable<JobLoadBalancerState> getLoadBalancersForJob(String jobId) {
-        logger.debug("Getting all load balancer states for job {}", jobId);
-        return Observable.from(loadBalancerStateMap.entrySet())
-                .filter(entry -> entry.getKey().getJobId().equals(jobId))
-                .map(JobLoadBalancerState::from);
+                })
+                .ignoreElement()
```
andrew-leung: Are `ignore()` + `cast()` doing something different than `then()`?

fabiokung: Not really, it was on oversight. Good catch.

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
```diff
@@ -86,30 +106,43 @@
...
-                    TABLE_LOAD_BALANCER,
+                    TABLE_LOAD_BALANCER_ASSOCIATIONS,
                     COLUMN_JOB_ID,
                     COLUMN_LOAD_BALANCER);
 
+    private static Pair<JobLoadBalancer, JobLoadBalancer.State> buildLoadBalancerStatePairFromRow(Row row) {
+        String state = row.getString(COLUMN_STATE);
+        return Pair.of(
+                new JobLoadBalancer(row.getString(COLUMN_JOB_ID), row.getString(COLUMN_LOAD_BALANCER)),
+                state == null ? JobLoadBalancer.State.DISSOCIATED : JobLoadBalancer.State.valueOf(state.toUpperCase())
```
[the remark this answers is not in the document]
fabiokung: No expected cases that I know, I just made the code a bit more defensive since it's being used in `Enum.valueOf`. I will add a warning log when it is `null`.

titus-api/src/main/java/com/netflix/titus/api/loadbalancer/store/LoadBalancerStore.java
```diff
@@ -85,13 +68,32 @@
...
 
     /**
-     * Blocking call the returns the current snapshot page of the given offset/size of
-     * all load balancers.
-     * As a blocking call, data must be served from cached/in-memory data and avoid doing external calls.
-     *
-     * @param offset
-     * @param limit
-     * @return
+     * Blocking call the returns the current snapshot page of the given offset/size of * all load balancers. As a
```
corindwyer: call the returns -> call that returns

corindwyer: * -> times

titus-api/src/main/java/com/netflix/titus/api/loadbalancer/store/LoadBalancerStore.java
```diff
@@ -85,13 +68,32 @@
...
      */
     List<JobLoadBalancer> getAssociationsPage(int offset, int limit);
+
+    /**
+     * Adds or updates targets with the provided states.
+     */
+    Mono<Void> addOrUpdateTargets(Collection<LoadBalancerTargetState> targets);
+
+    /**
+     * Adds a or updates targets with the provided states.
```
corindwyer: Adds a or -> Adds or

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
```diff
@@ -86,30 +106,49 @@
...
+    private static Pair<JobLoadBalancer, JobLoadBalancer.State> buildLoadBalancerStatePairFromRow(Row row) {
+        String jobId = row.getString(COLUMN_JOB_ID);
+        String loadBalancerId = row.getString(COLUMN_LOAD_BALANCER);
+        String state = row.getString(COLUMN_STATE);
+        JobLoadBalancer.State parsedState;
+        if (state == null) {
+            logger.warn("Unexpected null state for association {}:{}", jobId, loadBalancerId);
+            parsedState = JobLoadBalancer.State.DISSOCIATED;
+        } else {
+            parsedState = JobLoadBalancer.State.valueOf(state.toUpperCase());
```
tbak: Try/cach in case the value is corrupted? Unless this error would be handled somewhere else.

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
```diff
@@ -86,30 +106,49 @@
...
+    }
+
+    private static LoadBalancerTargetState buildLoadBalancerTargetStateFromRow(Row row) {
+        return new LoadBalancerTargetState(
+                new LoadBalancerTarget(
+                        row.getString(COLUMN_LOAD_BALANCER),
+                        row.getString(COLUMN_TASK_ID),
+                        row.getString(COLUMN_IP_ADDRESS)
+                ),
+                LoadBalancerTarget.State.valueOf(row.getString(COLUMN_STATE))
```
tbak: try/catch?

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
```diff
@@ -122,9 +161,27 @@ public CassandraLoadBalancerStore(CassandraStoreConfiguration configuration,
...
 
-        this.getAllJobIdsStmt = session.prepare(GET_ALL_JOB_IDS);
-        this.insertLoadBalancerStmt = session.prepare(INSERT_JOB_LOAD_BALANCER);
-        this.deleteLoadBalancerStmt = session.prepare(DELETE_JOB_LOAD_BALANCER);
+        this.selectAssociations = session.prepare(GET_ALL_ASSOCIATIONS).setConsistencyLevel(ConsistencyLevel.LOCAL_QUORUM);
+        this.insertAssociation = session.prepare(INSERT_ASSOCIATION).setConsistencyLevel(ConsistencyLevel.LOCAL_QUORUM);
+        this.deleteAssociation = session.prepare(DELETE_ASSOCIATION).setConsistencyLevel(ConsistencyLevel.LOCAL_QUORUM);
+
+        this.selectTargetsForLoadBalancer = session.prepare(
+                select(COLUMN_LOAD_BALANCER, COLUMN_IP_ADDRESS, COLUMN_TASK_ID, COLUMN_STATE)
```
tbak: Two different styles of building queries?

titus-server-master/src/main/java/com/netflix/titus/master/loadbalancer/service/DefaultLoadBalancerReconciler.java
```diff
@@ -163,97 +176,134 @@ public void shutdown() {
...
         return updatesForLoadBalancer
                 .doOnError(e -> logger.error("Error while reconciling load balancer {}", loadBalancerId, e))
                 .onErrorResumeNext(Observable.empty());
     }
 
     /**
-     * Generates a stream of necessary updates based on what jobs are currently associated with a load balancer, and the
-     * ip addresses currently registered on it.
+     * Generates a stream of necessary updates based on what jobs are currently associated with a load balancer, the
+     * ip addresses currently registered on it, and what ips were previously registered by us.
```
corindwyer: by us -> with the load balancer

titus-server-master/src/main/java/com/netflix/titus/master/loadbalancer/service/DefaultLoadBalancerReconciler.java
```diff
@@ -163,97 +176,134 @@ public void shutdown() {
...
      * <p>
      * {@link JobLoadBalancer} associations in the <tt>Dissociated</tt> state will be removed when it is safe to do so,
-     * i.e.: when there is nothing from them to be deregistered on the load balancer anymore.
+     * i.e.: when there is no more stored state to be cleaned up, and nothing to be deregistered on the load balancer.
      *
-     * @param loadBalancer id, state, and ip addresses currently registered on the load balancer
+     * @param loadBalancer tuple with ip addresses currently registered on the load balancer, and ip addresses
+     *                     previously registered by us
```
corindwyer: by us -> with the load balancer

titus-server-master/src/main/java/com/netflix/titus/master/loadbalancer/service/DefaultLoadBalancerReconciler.java
```diff
@@ -163,97 +176,134 @@ public void shutdown() {
...
+                        .doOnSubscribe(ignored -> logger.info("Cleaning up {} deregistered targets for load balancer {}",
+                                updates.toRemove.size(), loadBalancer.current.getId()))
+                        .compose(removeTargetsMetrics.asCompletable())
+                        .doOnError(e -> logger.error("Error while cleaning up targets for " + loadBalancer.current.getId(), e))
+                        .onErrorComplete()
+                : Completable.complete();
 
-        // clean up Dissociated entries when all their targets have been deregistered
-        final Completable removeOperations = toDeregister.isEmpty() ?
+        // clean up dissociated entries only it is safe: all targets have been deregistered and there is nothing to be removed
```
corindwyer: only it is safe -> only if it is safe

titus-server-master/src/main/java/com/netflix/titus/master/loadbalancer/service/DefaultLoadBalancerService.java
```diff
@@ -198,4 +212,60 @@ public void deactivate() {
     Observable<Batch<TargetStateBatchable, String>> events() {
         return engine.events();
     }
+
+    /**
+     * Temporary utility to backfill all targets from registered load balancers to the store, since targets were not
+     * being persisted before.
+     * <p>
+     * Once all current state has been backfilled (i.e.: this runs successfully at least once), this is not needed
```
tbak: This is only correct, if the load balancer is managed by one cell?

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
fabiokung: I think we should let the error bubble up and prevent activation here, since this would be a case of data corruption. FWIW the code previous to changes here was doing the same (only without `.toUpperCase()`).

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
fabiokung: same as before, I think we should let the error bubble up since it's data corruption (i.e.: this should never fail to parse). What we could do here is add a configuration flag to ignore corrupt entries when set, but only if you feel strong about it.

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraLoadBalancerStore.java
fabiokung: I liked the type/grammar safety of the dsl builder, but didn't want to go and change everything else in this PR. Any strong preferences?

titus-server-master/src/main/java/com/netflix/titus/master/loadbalancer/service/DefaultLoadBalancerService.java
fabiokung: existing code requires that a load balancer is managed only by one cell (single titusmaster leader), so that is assumed. After backfill happens and this property is disabled, multiple cells can start managing the same load balancer.

PR #740, opened 2020-01-02
Add PaginationEvaluator to simplify the pagination processing

tbak:


titus-common-api/src/main/java/com/netflix/titus/api/model/PaginationEvaluator.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        String jobId = matcher.group(1);
+        long timestamp = Long.parseLong(matcher.group(2));
+
+        return Pair.of(jobId, timestamp);
+    }
+
+    private static int numberOfPages(Page page, int totalItems) {
+        return (totalItems + page.getPageSize() - 1) / page.getPageSize();
```
andrew-leung: Do we validate somewhere that the user hasn't set pageSize to 0?

andrew-leung: Also, I'm a little confused about the calculation here. If totalItems is 10 and pageSize is 4, I would expect there to be 3 pages. Seems like this would return 4.

titus-common-api/src/main/java/com/netflix/titus/api/model/PaginationEvaluator.java
```diff
@@ -0,0 +1,200 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
andrew-leung: 2020?

titus-common-api/src/main/java/com/netflix/titus/api/model/PaginationEvaluator.java
tbak: No, we do not do that. I will add this check.
(10 + 4  - 1)/4 = 3. The calculation is correct.

tbak: I will support page size 0 as a special case, which returns empty list. This is useful if a client is interested only in the pagination data (mostly total items).

andrew-leung: Yeah, I must have misread it before. Looks correct.

PR #754, opened 2020-01-31
ManyReconciler refactoring

tbak:


titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/internal/DefaultManyReconciler.java
```diff
@@ -144,14 +148,19 @@ public DefaultManyReconciler(
...
 
     @Override
     public Mono<Void> close() {
         return Mono.defer(() -> {
             stateRef.compareAndSet(ReconcilerState.Running, ReconcilerState.Closing);
-            return closedProcessor;
+
+            return closedProcessor.doFinally(signal -> {
+                reconcilerSchedulerRef.close();
+                // TODO There may be pending notifications, so we have to delay the actual close. There must be a better solution than timer.
```
andrew-leung: Is there no way to get a pending count here?

tbak: Yes, we could do it but it would be a bit more complicated. I will fix that when working again on this code.

PR #764, opened 2020-02-07
Add containers per node limit placement constraint

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailure.java
```diff
@@ -54,6 +54,11 @@
          */
         LaunchGuard,
 
+        /**
+         * Task not launched on an agent, as it runs the maximum allowed number of containers.
+         */
+        AgentContainerLimit,
```
fabiokung: add this new kind to `FailureKind.IGNORED_FOR_OPPORTUNISTIC_SCHEDULING` since it should not cause opportunistic behavior to change

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/AgentContainerLimitSystemConstraint.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public Result evaluate(TaskRequest taskRequest, VirtualMachineCurrentState targetVM, TaskTrackerState taskTrackerState) {
+        int total = targetVM.getRunningTasks().size() + targetVM.getTasksCurrentlyAssigned().size();
+        int max = Math.max(1, configuration.getMaxTasksPerMachine());
+        return total < max ? VALID : TOO_MANY_CONTAINERS;
+    }
+
+    public static boolean isAgentContainerLimitSystemConstraint(String reason) {
+        return reason != null && TOO_MANY_CONTAINERS.getFailureReason().contains(reason);
```
fabiokung: should this be protected against the empty string? Why not use `.equals()` instead of `.contains()`?

tbak: Copy paste. It should be equals.

PR #766, opened 2020-02-07
publish more detailed metrics on failed placements

fabiokung:
Currently, we only publish a gauge with the total number of tasks that failed placement in the last scheduling iteration.

This adds a new (counter) metric with more insights into why placements failed and for what application/capacityGroup. It will also count *all* failures from all scheduling iterations.

Failures are grouped by `(applicationName, capacityGroup)` to limit the cardinality of the new metric.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -54,7 +54,7 @@ services:
       - 1.0.0.1
 
   mesos-master:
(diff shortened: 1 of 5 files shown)


fabiokung:
@corindwyer yeah, I think it makes sense to add the tier. I will do it.

PR #769, opened 2020-02-13
process C* resultsets outside of C* driver threads

fabiokung:
This prevents async (on demand) fetching of `ResultSet` pages blocking C* driver threads.

The Datastax driver has a deadlock detector that complains (throws an exception) when pagination of resultsets would cause a deadlock, particularly when results are processed in C* driver threads and blocking the C* driver event loop.

### DIFF
--- a/titus-ext/cassandra/src/test/java/com/netflix/titus/ext/cassandra/store/CassStoreHelperTest.java
+++ b/titus-ext/cassandra/src/test/java/com/netflix/titus/ext/cassandra/store/CassStoreHelperTest.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


titus-ext/cassandra/src/test/java/com/netflix/titus/ext/cassandra/store/CassStoreHelperTest.java
```diff
@@ -0,0 +1,104 @@
...
+                // force pagination, and pages to be fetched on demand as the ResultSet is iterated on
+                .setFetchSize(numberOfRecords / numberOfPages));
+
+        results.doOnNext(rows -> rows.forEach(row ->
+                assertThat(row.getString(0)).startsWith("job-"))
+        ).toCompletable().await(1, TimeUnit.MINUTES);
+    }
+
+    @Test(expected = IllegalStateException.class /* the datastax driver complains that page fetching will cause a deadlock */ )
+    public void paginatedResultInCassandraThreadsThrowsException() {
```
amit-git: Minor - but potentially a common 'fetchAllPages' method that takes scheduler as a parameter could simplify both of these test case code

fabiokung: > Minor - but potentially a common 'fetchAllPages' method that takes scheduler as a parameter could simplify both of these test case code

That would require pre-loading all rows in memory, which iterating on resultsets is trying to avoid. I can see a callback or stream/reactive based API for it (e.g. `walkAllPages(Consumer<T> processor)`), but I didn't see a need to overcomplicate ATM

fabiokung: nvm, I see what mean. I agree code duplication on both test cases would be good, next time.

PR #770, opened 2020-02-13
Add grpc to reactor server support for cglib proxies

andrew-leung:
Spring security annotations result in cglib proxied classes. These classes do no retain enough generic type information for the grpc to reactor method handler to work with, so we pass additional class details in these cases.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/grpc/reactor/GrpcToReactorServerFactory.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/grpc/reactor/GrpcToReactorServerFactory.java
@@ -23,5 +23,10 @@ public interface GrpcToReactorServerFactory {
 
     <REACT_SERVICE> ServerServiceDefinition apply(ServiceDescriptor serviceDefinition, REACT_SERVICE reactService);
 
+    /**
+     * This method is used when the {@link REACT_SERVICE} is a CGLIB proxy and additional class details need to be provided to the factory.
+     * If the {@link REACT_SERVICE} is a CGLIB proxy class it does not retain generic type information from the proxy target class so a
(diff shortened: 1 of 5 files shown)


titus-common/src/main/java/com/netflix/titus/common/util/grpc/reactor/GrpcToReactorServerFactory.java
```diff
@@ -22,4 +22,6 @@
 public interface GrpcToReactorServerFactory {
 
     <REACT_SERVICE> ServerServiceDefinition apply(ServiceDescriptor serviceDefinition, REACT_SERVICE reactService);
+
+    <REACT_SERVICE> ServerServiceDefinition apply(ServiceDescriptor serviceDefinition, REACT_SERVICE reactService, Class<REACT_SERVICE> reactorDetailedFallbackClass);
```
tbak: Could you add a comment why we need `reactorDetailedFallbackClass`, and cannot just do `reactService.getClass()`?

titus-common/src/main/java/com/netflix/titus/common/util/grpc/reactor/server/DefaultGrpcToReactorServerFactory.java
```diff
@@ -34,8 +34,14 @@ public DefaultGrpcToReactorServerFactory(Class<CONTEXT> contextType, Supplier<CO
 
     @Override
     public <REACT_SERVICE> ServerServiceDefinition apply(ServiceDescriptor serviceDescriptor, REACT_SERVICE reactService) {
+        return apply(serviceDescriptor, reactService, null);
```
tbak: Can we pass here `reactService.getClass()`, not null?

PR #772, opened 2020-02-13
Split Kube notifications based on scheduler type (Fenzo/Kube)

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobAndTaskMetrics.java
```diff
@@ -257,7 +256,7 @@ private void updateJobCounts(List<Pair<Job, List<Task>>> jobsAndTasks) {
             if (JobFunctions.getJobDesiredSize(job) == 0) {
                 emptyJobs++;
             } else {
-                boolean hasPods = tasks.stream().anyMatch(this::hasPod);
+                boolean hasPods = tasks.stream().anyMatch(JobFunctions::hasOwnedByKubeSchedulerAttribute);
```
fabiokung: minor nit: rename the variable to reflect its new meaning

PR #773, opened 2020-02-14
Support task terminate with Kube scheduler

tbak:


tbak:
It is used in `terminateTask` when a pod for the task does not exist in
Kube.

fabiokung:
@tbak this is where it is sent, but where are you planning on consuming it? Anyway, no big deal, most out of curiosity.

tbak:
@fabiokung Good catch. I have forgotten to handle this event in the `KubeNotificationProcessor`. I will address this in my next PR.

PR #775, opened 2020-02-14
Shared Kube informer

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultKubeApiFactory.java
```diff
@@ -0,0 +1,180 @@
...
+import io.kubernetes.client.models.V1Pod;
+import io.kubernetes.client.models.V1PodList;
+import io.kubernetes.client.util.CallGeneratorParams;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.netflix.titus.master.mesos.kubeapiserver.KubeUtil.createSharedInformerFactory;
+
+@Singleton
+public class DefaultKubeApiFactory implements KubeApiFactory {
```
fabiokung: minor nit: first time I read the code, the name `Factory` implied to me that new instances would be returned for each call, since this is providing singletons (memoized instances), maybe rename this class to `...Provider`, or `...Holder`?

tbak: `Provider` would be better suffix, but it is used by Guice, and it would be confusing.
I will try to find a better name.

tbak:
@fabiokung  Yes, I have missed that one.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultKubeApiFactory.java
fabiokung: well, I already used `...Provider` in some other places :-)

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultKubeApiFacade.java
```diff
@@ -119,8 +148,9 @@ private void activate() {
                         apiClient
                 );
 
-                this.nodeInformer = createNodeInformer(sharedInformerFactory, coreV1Api);
-                this.podInformer = createPodInformer(sharedInformerFactory, coreV1Api);
+                this.nodeInformer = createNodeInformer(sharedInformerFactory);
```
corindwyer: Since these are now centrally managed, would be nice to have gauges that show their current size.

tbak: I am working on metrics in another PR. I will include this as well.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultKubeApiFacade.java
```diff
@@ -0,0 +1,243 @@
...
+                        null
+                ),
+                V1Pod.class,
+                V1PodList.class,
+                configuration.getKubeApiServerIntegratorRefreshIntervalMs()
+        );
+    }
+
+    private SharedIndexInformer<V1OpportunisticResource> createOpportunisticResourceInformer(SharedInformerFactory sharedInformerFactory) {
+        // TODO(fabio): enhance the kube client to support custom JSON deserialization options
```
fabiokung: you can drop this TODO, I wanted to use jackson instead of gson to get support for immutable POJOs, but it's not worth the effort

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultKubeApiFacade.java
```diff
@@ -0,0 +1,243 @@
...
+        return sharedInformerFactory.sharedIndexInformerFor(
+                this::listOpportunisticResourcesCall,
+                V1OpportunisticResource.class,
+                V1OpportunisticResourceList.class,
+                configuration.getKubeOpportunisticRefreshIntervalMs()
+        );
+    }
+
+    private Call listOpportunisticResourcesCall(CallGeneratorParams params) {
+        try {
```
fabiokung: probably safe to remove this try catch now and the `return null` below

tbak: This method throws checked exception. Wrapping it with `IllegalStateException`

PR #779, opened 2020-02-18
Shared Kube/Titus job reconciler

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeJobManagementReconciler.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.titus.master.mesos.kubeapiserver;
+
+import com.netflix.titus.master.mesos.ContainerEvent;
+import com.netflix.titus.master.mesos.kubeapiserver.direct.model.PodEvent;
+import reactor.core.publisher.Flux;
+
+public interface KubeJobManagementReconciler {
```
corindwyer: please add javadocs

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
```diff
@@ -0,0 +1,286 @@
...
+                                .withTimestamp(clock.wallTime())
+                                .build()
+                );
+            }
+            logger.info("Finished orphaned task transitions to finished ({})", kind);
+        });
+    }
+
+    private Optional<V1Node> findNode(Task task, Map<String, V1Node> nodes) {
+        String instanceId = task.getTaskContext().get(TaskAttributes.TASK_ATTRIBUTES_AGENT_INSTANCE_ID);
```
corindwyer: Instead of doing this, I suggest looking up the pod object and then using the pods nodeName so nodeName can change.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
```diff
@@ -0,0 +1,286 @@
...
+                String reasonCode;
+                String reasonMessage;
+                switch (kind) {
+                    case KILL_INITIATED:
+                        reasonCode = TaskStatus.REASON_TASK_KILLED;
+                        reasonMessage = "Task killed";
+                        break;
+                    case NODE_TERMINATED:
+                        reasonCode = TaskStatus.REASON_TASK_LOST;
+                        reasonMessage = "Agent terminated";
```
corindwyer: I think there should be a better message here. Maybe `Agent terminated due to underlying issue`?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/TaskStatus.java
```diff
@@ -153,6 +153,14 @@ public static boolean isEvicted(Task task) {
                 ).orElse(false);
     }
 
+    public static boolean hasPod(Task task) {
+        TaskStatus acceptedStatus = JobFunctions.findTaskStatus(task, TaskState.Accepted).orElse(null);
```
corindwyer: If there are many Accepted entries in the status history then you might not get the correct one.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultContainerResultCodeResolver.java
```diff
@@ -0,0 +1,76 @@
...
+        this.localSystemErrorMessageMatcherFactory = RegExpExt.dynamicMatcher(mesosConfiguration::getLocalSystemErrorMessagePattern, "localSystemErrorMessagePattern", Pattern.DOTALL, logger);
+        this.unknownSystemErrorMessageMatcherFactory = RegExpExt.dynamicMatcher(mesosConfiguration::getUnknownSystemErrorMessagePattern, "unknownSystemErrorMessagePattern", Pattern.DOTALL, logger);
+    }
+
+    public Optional<String> resolve(TaskState taskState, String reasonMessage) {
+        if (taskState != com.netflix.titus.api.jobmanager.model.job.TaskState.Finished || StringExt.isEmpty(reasonMessage)) {
+            return Optional.empty();
+        }
+
+        if (invalidRequestMessageMatcherFactory.apply(reasonMessage).matches()) {
```
fabiokung: minor: add `(matcher, reasonString)` pairs to a collection and `findFirst()` on it to reduce duplication here

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    @Override
+    public Flux<PodEvent> getPodEventSource() {
+        return podEventProcessor.compose(ReactorExt.badSubscriberHandler(logger));
+    }
+
+    private void reconcile() {
+        if (!mesosConfiguration.isReconcilerEnabled() || !kubeApiFacade.getNodeInformer().hasSynced() || !kubeApiFacade.getPodInformer().hasSynced()) {
+            return;
```
fabiokung: log something and/or emit a metric so we know when reconciliation is not running? Or alternatively, emit a metric when it runs

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
fabiokung: :+1:, something like: `Terminated due to an issue with the underlying host machine`

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
```diff
@@ -0,0 +1,286 @@
...
+                        reasonMessage = "Task killed";
+                        break;
+                    case NODE_TERMINATED:
+                        reasonCode = TaskStatus.REASON_TASK_LOST;
+                        reasonMessage = "Agent terminated";
+                        break;
+                    case UNKNOWN:
+                    default:
+                        reasonCode = TaskStatus.REASON_TASK_LOST;
+                        reasonMessage = "Task lost between control plane and machine";
```
fabiokung: suggestion, something like: `Abandoned with unknown state due to lack of status updates from the host machine`

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
```diff
@@ -0,0 +1,286 @@
...
+            return Optional.empty();
+        }
+        return Optional.ofNullable(nodes.get(instanceId));
+    }
+
+    private boolean shouldTaskBeInApiServer(Task task) {
+        boolean isRunning = TaskState.isRunning(task.getStatus().getState());
+
+        if (JobFunctions.isOwnedByKubeScheduler(task)) {
+            return isRunning || (task.getStatus().getState() == TaskState.Accepted && TaskStatus.hasPod(task));
```
fabiokung: In the `TaskState.Accepted && hasPod()` case, I think we need a grace time to allow state propagation through kube-apiserver and our informers, otherwise we may be killing tasks prematurely

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultDirectKubeApiServerIntegrator.java
```diff
@@ -146,7 +150,14 @@ public void shutdown() {
                 metrics.terminateError(task, e, timer.elapsed(TimeUnit.MILLISECONDS));
 
                 if (e.getMessage().equalsIgnoreCase(NOT_FOUND) && task.getStatus().getState() == TaskState.Accepted) {
-                    sendEvent(PodEvent.onPodNotFound(task));
+                    sendEvent(PodEvent.onPodNotFound(task,
+                            TaskStatus.newBuilder()
+                                    .withState(TaskState.Finished)
+                                    .withReasonCode(TaskStatus.REASON_TASK_LOST)
+                                    .withReasonMessage("Pod terminate requested, but the pod is not found")
```
fabiokung: minor: maybe `s/pod/task/` since this is an user facing message

fabiokung:
@tbak everything other than the grace time for terminations in the reconciliation loop is a minor and can be ignored

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
tbak: Is it something that may happen today, or the goal is to make the code more flexible.
If instance id in the task context is different from node instance id, it is a major consistency issue.

corindwyer: It is to make it more flexible as the nodeName in the future may not match the instance id.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultContainerResultCodeResolver.java
tbak: I would prefer to keep it as it is, in case in the future we need more complex logic when parsing reason messages or checking additional information in the task state.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultDirectKubeApiServerIntegrator.java
tbak: Here the task exists in Titus, but there is no pod. Maybe something like this:
`Task terminate requested, but its container is not found`

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
tbak: Adding a new task attribute: 
```java
public static final String TASK_ATTRIBUTES_KUBE_NODE_NAME = "kube.nodeName"
```

titus-api/src/main/java/com/netflix/titus/api/jobmanager/TaskAttributes.java
```diff
@@ -22,15 +22,19 @@
      */
     public static final String TASK_ATTRIBUTES_AGENT_REGION = "agent.region";
     public static final String TASK_ATTRIBUTES_AGENT_ZONE = "agent.zone";
+    public static final String TASK_ATTRIBUTES_AGENT_AMI = "agent.ami";
     public static final String TASK_ATTRIBUTES_AGENT_ASG = "agent.asg";
     public static final String TASK_ATTRIBUTES_AGENT_HOST = "agent.host";
+    public static final String TASK_ATTRIBUTES_AGENT_CLUSTER = "agent.cluster";
     public static final String TASK_ATTRIBUTES_AGENT_INSTANCE_ID = "agent.instanceId";
     public static final String TASK_ATTRIBUTES_AGENT_ITYPE = "agent.itype";
+    public static final String TASK_ATTRIBUTES_AGENT_RES = "agent.res";
```
fabiokung: minor: I know it is not related to your PR, but what is `agent.res`? A comment here may help

tbak: Adding a comment now in another PR. This attributes holds agent ENI resources.

PR #780, opened 2020-02-19
use kube config file

corindwyer:
### Description of the Change
Use kube config file instead of a configured url to kube-apiserver.


titus-server-master/src/main/java/com/netflix/titus/master/mesos/MesosConfiguration.java
```diff
@@ -109,10 +109,10 @@
     boolean isKubeApiServerIntegrationEnabled();
 
     /**
-     * @return the url of the Kube API Server
+     * @return the path to the kubeconfig file
      */
-    @DefaultValue("localhost:8080")
-    String getKubeApiServerUrl();
+    @DefaultValue("/run/kubernetes/config")
+    String getKubeConfigPath();
```
tbak: It would be good to support both modes (the former for testing).

PR #783, opened 2020-02-21
Add Kube scheduler tolerations

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/KubeConstants.java
```diff
@@ -46,4 +46,34 @@
...
+
+    /**
+     * Machines with this taint and value 'titus' can be used for pod placement.
+     */
+    public static final String TAINT_VIRTUAL_KUBLET = "virtual-kubelet.io/provider";
+
+    /**
+     * Common prefix for Titus taints.
+     */
+    public static final String TITUS_TAINT_DOMAIN = "com.netflix.titus.agent.taint/";
```
corindwyer: where are these taints defined?

PR #787, opened 2020-02-25
Enable junit categories be extended

andrew-leung:
Switch JUnit categories to plain interfaces from annotations to allow extending and move Smoke and Functional test categories from CMB to Titus testkit common.


andrew-leung:
Yes, CMB will be using the `RemoteIntegrationTest` from here instead of its own.

PR #788, opened 2020-02-25
Miscellaneous updates

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
```diff
@@ -193,11 +196,31 @@ public void shutdown() {
...
+                        newStatus = TaskStatus.newBuilder()
+                                .withState(newTaskState)
+                                .withReasonCode(TaskStatus.REASON_NORMAL)
+                                .withReasonMessage("Kube pod notification")
+                                .build();
+                    } else {
+                        TaskStatus.Builder newStatusBuilder = TaskStatus.newBuilder().withState(TaskState.Finished);
+
+                        if("failed".equalsIgnoreCase(pod.getStatus().getPhase())) {
+                            newStatusBuilder
```
corindwyer: `Failed` phase can also have a deletion timestamp.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -327,6 +334,34 @@ private void nodeUpdated(V1Node node) {
...
+     */
+    private boolean isNodeOwnedByFenzo(V1Node node) {
+        List<V1Taint> taints = node.getSpec().getTaints();
+        if (CollectionsExt.isNullOrEmpty(taints)) {
+            return true;
+        }
+
+        Set<String> schedulerTaintValues = taints.stream()
+                .filter(t -> KubeConstants.TAINT_SCHEDULER.equals(t.getKey()))
+                .map(t -> StringExt.safeTrim(t.getValue()))
```
corindwyer: You can use `findFirst()` here since you only want the first value.

titus-common/src/main/java/com/netflix/titus/common/util/loadshedding/SpectatorAdmissionController.java
```diff
@@ -42,18 +44,18 @@ public AdmissionControllerResponse apply(AdmissionControllerRequest request) {
             AdmissionControllerResponse result = delegate.apply(request);
 
             registry.counter(METRIC_NAME,
-                    "callerId", request.getCallerId(),
-                    "endpointName", request.getEndpointName(),
+                    "callerId", getNonEmptyOrDefault(request.getCallerId(), "notSet"),
```
fabiokung: minor: `s/notSet/unknown`

tbak: I have purposefully chose different word, as we use `unknown` in many places in the code already. I want here something that is distinct, as we know that an empty value is a bug. Maybe better would be `requiredButNotSet`.

PR #791, opened 2020-02-28
upgrade gradle and nebula plugins

fabiokung:
... enabling immutable snapshots and gradle composite builds.

Some of our gradle configuration and dependencies needed to be fixed to work with the newer versions.


titus-supplementary-component/task-relocation-springboot/build.gradle
```diff
@@ -14,14 +14,8 @@
...
 
-buildscript {
-    dependencies {
-        classpath "org.springframework.boot:spring-boot-gradle-plugin:2.1.2.RELEASE"
-    }
-}
-
 plugins {
-    id 'org.springframework.boot' version '2.1.2.RELEASE'
+    id 'org.springframework.boot' version '2.1.3.RELEASE'
```
corindwyer: I noticed this version is duplicated a lot. Is it possible to use a variable in the top level build.gradle?

fabiokung: maybe. I noticed variables are not picked up by all gradle blocks. Let me give it a quick try. If it is simple I will make the change.

PR #795, opened 2020-02-29
Refactoring ConstraintValidators to sanitize violation message templates

joshi-keyur:
### Description of the Change

Sanitize message string so it is not an EL expression by escaping special characters.
The message string in this case is a collection of user supplied values which is being used as a part of message template for message interpolation.


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/SchedulingConstraintSetValidator.java
```diff
@@ -53,7 +54,7 @@ public void initialize(SchedulingConstraintSet constraintAnnotation) {
     }
 
     @Override
-    public boolean isValid(Container container, ConstraintValidatorContext context) {
+    public boolean isValid(Container container, ConstraintValidatorContextWrapper context) {
```
tbak: `protected`? We want clients to call the original validator method?

titus-common/src/main/java/com/netflix/titus/common/model/sanitizer/internal/AbstractConstraintValidator.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Annotation;
+
+public abstract class AbstractConstraintValidator<A extends Annotation, T> implements ConstraintValidator<A, T> {
+
+    @Override
+    final public boolean isValid(T type, ConstraintValidatorContext context) {
+        return this.isValid(type, new ConstraintValidatorContextWrapper(context));
+    }
+
+    abstract public boolean isValid(T type, ConstraintValidatorContextWrapper wrapper);
```
tbak: `protected`?

titus-common/src/main/java/com/netflix/titus/common/model/sanitizer/internal/ConstraintValidatorContextWrapper.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.common.model.sanitizer.internal;
+
+import javax.validation.ConstraintValidatorContext;
+
+public class ConstraintValidatorContextWrapper {
```
tbak: Replace this class with `Function< String, ConstraintViolationBuilder >`, and the default implementation as here?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/SchedulingConstraintValidator.java
```diff
@@ -66,17 +65,16 @@
 
     @Override
     public void initialize(SchedulingConstraint constraintAnnotation) {
-    }
+   }
```
fabiokung: minor: weird indentation here

titus-common/src/main/java/com/netflix/titus/common/model/sanitizer/internal/ConstraintValidatorContextWrapper.java
joshi-keyur: Thank you for the feedback. Please review the updates.

PR #796, opened 2020-03-02
create accountId amazon client provider and use it in the load balancer integration

corindwyer:


titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonClientProvider.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Inject
+    public AmazonClientProvider(AwsConfiguration configuration,
+                                AWSSecurityTokenServiceAsync stsClient) {
+        this.configuration = configuration;
+        this.stsClient = stsClient;
+    }
+
+    public AmazonElasticLoadBalancingAsync getLoadBalancingClient(String accountId) {
+        return loadBalancerClients.computeIfAbsent(accountId, id -> {
```
tbak: Is it ok to create many of those during initialization if there are concurrent calls?

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/loadbalancer/AwsLoadBalancerConnector.java
```diff
@@ -173,14 +175,19 @@ public Completable isValid(String loadBalancerId) {
...
-                        return Single.just(new LoadBalancer(id, LoadBalancer.State.REMOVED, Collections.emptySet()));
+                        return Single.just(new LoadBalancer(loadBalancerId, LoadBalancer.State.REMOVED, Collections.emptySet()));
                     }
                     return Single.error(throwable);
                 });
     }
+
+    private AmazonElasticLoadBalancingAsync getClient(String loadBalancerId) {
+        Arn arn = Arn.fromString(loadBalancerId);
+        return clientProvider.getLoadBalancingClient(arn.getAccountId());
```
tbak: Is it an invariant that account id is always set in the load balancer id?

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonClientProvider.java
```diff
@@ -0,0 +1,65 @@
...
+            Arn roleArn = getControlPlaneRoleArnForAccount(id);
+
+            return new STSAssumeRoleSessionCredentialsProvider.Builder(roleArn.toString(), roleSessionName)
+                    .withStsClient(stsClient)
+                    .withRoleSessionDurationSeconds(roleSessionDurationSeconds)
+                    .build();
+        });
+    }
+
+    private Arn getControlPlaneRoleArnForAccount(String accountId) {
```
joshi-keyur: Looks like we are using string representation arn.toString() returned by this function. Do we need to build Arn object or would we rather build a string and return just that?

corindwyer: I am using the ARN builder in order for it to provide validations and guarantee what is outputted is a valid ARN.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonClientProvider.java
fabiokung: :+1: probably not a huge deal, it might just increase the number of thread (pools) and connections during initialization a bit, but I agree it would be probably better to do a simple double-checked lock of the whole thing here and serialize calls

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/loadbalancer/AwsLoadBalancerConnector.java
corindwyer: yes, a valid ARN must always contain the accountId and `Arn.fromString(loadBalancerId)` does the validation.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonClientProvider.java
joshi-keyur: How do we make sure the credentials provider in the Map is still valid and has not expired the session duration when first constructed?

corindwyer: The `STSAssumeRoleSessionCredentialsProvider` has a background thread that does the automatic refreshing.

corindwyer: Fixed

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonClientProvider.java
```diff
@@ -29,27 +29,43 @@ public AmazonClientProvider(AwsConfiguration configuration,
...
-            String region = configuration.getRegion().trim().toLowerCase();
-            AWSCredentialsProvider credentialsProvider = getAwsCredentialsProvider(id);
-            return AmazonElasticLoadBalancingAsyncClientBuilder.standard()
-                    .withCredentials(credentialsProvider)
-                    .withRegion(region)
-                    .build();
-        });
+        AmazonElasticLoadBalancingAsync client = loadBalancerClients.get(accountId);
+        if (client == null) {
+            synchronized (AmazonClientProvider.class) {
```
fabiokung: minor: sinchronize on an instance level mutex just to avoid issues when this class is not being used as a singleton (e.g.: tests)

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonClientProvider.java
```diff
@@ -29,27 +29,43 @@ public AmazonClientProvider(AwsConfiguration configuration,
...
     }
 
     private AWSCredentialsProvider getAwsCredentialsProvider(String accountId) {
-        return awsCredentialsByAccountId.computeIfAbsent(accountId, id -> {
-            String roleSessionName = configuration.getControlPlaneRoleSessionName();
-            int roleSessionDurationSeconds = configuration.getControlPlaneRoleSessionDurationSeconds();
-            Arn roleArn = getControlPlaneRoleArnForAccount(id);
+        AWSCredentialsProvider credentialsProvider = awsCredentialsByAccountId.get(accountId);
+        if (credentialsProvider == null) {
+            synchronized (AmazonClientProvider.class) {
```
fabiokung: same as above

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonClientProvider.java
corindwyer: Fixed

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonClientProvider.java
corindwyer: Fixed

PR #800, opened 2020-03-04
Add farzone support

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
```diff
@@ -264,6 +266,16 @@ private Task attachNodeMetadata(Task task, V1Node node) {
 
         acceptNotNull(node.getMetadata().getName(), nodeName -> agentAttributes.put(TaskAttributes.TASK_ATTRIBUTES_KUBE_NODE_NAME, nodeName));
 
+        List<V1NodeAddress> addresses = node.getStatus().getAddresses();
```
corindwyer: The first ipv4 ip address should be used as the host ip like in:
https://github.com/Netflix/titus-control-plane/blob/master/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java#L453

tbak: @corindwyer I have seen your recent PR. We should share your address filter in both places. Once you merge, I will rebase and create a function in `KubeUtil`.

corindwyer: Sounds good. I already merged.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
```diff
@@ -191,4 +197,12 @@ public static boolean isOwnedByKubeScheduler(V1Pod v1Pod) {
         }
         return false;
     }
+
+    public static String getNodeIpV4Address(V1Node node) {
+        return   node.getStatus().getAddresses().stream()
```
corindwyer: this looks incorrectly indented

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/batch/BatchDifferenceResolver.java
```diff
@@ -238,7 +244,7 @@ public BatchDifferenceResolver(
         }
 
         Map<String, String> taskContext = getTaskContext(previousTask, unassignedIpAllocations);
-        if (kubeSchedulerPredicate.test(refJobView.getJob().getJobDescriptor())) {
+        if (KubeUtil.findFarzoneId(kubeConfiguration, refJobView.getJob()).isPresent() || kubeSchedulerPredicate.test(refJobView.getJob().getJobDescriptor())) {
```
fabiokung: why not centralize this check into the `kubeSchedulerPredicate`?

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/service/ServiceDifferenceResolver.java
```diff
@@ -261,7 +267,7 @@ public ServiceDifferenceResolver(
         }
 
         Map<String, String> taskContext = getTaskContext(previousTask, unassignedIpAllocations);
-        if (kubeSchedulerPredicate.test(refJobView.getJob().getJobDescriptor())) {
+        if (KubeUtil.findFarzoneId(kubeConfiguration, refJobView.getJob()).isPresent() || kubeSchedulerPredicate.test(refJobView.getJob().getJobDescriptor())) {
```
fabiokung: same as above, centralize this into the predicate

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
```diff
@@ -148,4 +161,48 @@ public static String formatV1ContainerState(V1ContainerState containerState) {
 
         return "{state=<not set>}";
     }
+
+    /**
+     * If a job has an availability zone hard constraint with a farzone id, return this farzone id.
+     */
+    public static Optional<String> findFarzoneId(DirectKubeConfiguration configuration, Job job) {
+        List<String> farzones = configuration.getFarzones();
```
fabiokung: could this be generalized to any availabilityZones? Does it matter that it is a farzone? I'd recommend naming this `isAvailabilityZoneOwnedByKubeScheduler()` and `configuration.getAvailabilityZonesOwnedByKubeScheduler()`, or similar.

IOW, can we abstract away that some AZs are farzones?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultPodAffinityFactory.java
```diff
@@ -181,6 +186,12 @@ private void addNodeAffinityPreferredSelectorConstraint(String key, String value
             term.addMatchExpressionsItem(requirement);
         }
 
+        private void processFarzoneConstraints() {
```
fabiokung: same, generalize to any AZs, farzone or not

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/batch/BatchDifferenceResolver.java
tbak: I have tried that, but the code is in a different subproject.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/service/ServiceDifferenceResolver.java
tbak: I have tried that, but the code is in a different subproject.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
tbak: It is more than an additional zone. Farzones are excluded from the default scheduling, and there is extra logic around that. I would prefer to keep it explicit.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultPodAffinityFactory.java
tbak: See my comment above ^^^.

PR #801, opened 2020-03-05
always use the first InternalIP address as the host ip

corindwyer:
### Description of the Change
* Always use the first ipv4 InternalIP address as the host ip.


jigish:
For Context:

https://github.com/kubernetes/legacy-cloud-providers/blob/master/aws/aws.go#L1409

PR #803, opened 2020-03-06
Fix flaky integration tests

tbak:
Also restore the security manager in `BaseIntegrationTest`. There are a few more places where `System.exit` is called, and cannot be easily patched.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,13 @@ matrix:
   fast_finish: true
   include:
     - name: "Unit Tests"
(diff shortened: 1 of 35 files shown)


tbak:
@fabiokung Launch guard was disabled in several places, but when we made it default, we added max concurrency limit per node, which defaults to 1.  This broke several tests. They were sometimes passing, as tasks in launched state would expire and by doing this would allow a blocked task to be scheduled.

fabiokung:
Awesome, makes sense. Thanks for clarifying!

PR #804, opened 2020-03-09
add job descriptor annotation

corindwyer:
### Description of the Change
* Add configurable job descriptor annotation to the pod object.


titus-common/src/main/java/com/netflix/titus/common/util/StringExt.java
```diff
@@ -540,4 +543,24 @@ public static String startWithLowercase(String text) {
...
+     */
+    public static String gzipAndBase64Encode(String s) {
+        if (StringExt.isEmpty(s)) {
+            return s;
+        }
+        try {
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            GZIPOutputStream gzip = new GZIPOutputStream(out);
+            gzip.write(s.getBytes());
+            gzip.close();
```
fabiokung: worth using a try-with-resources here (and move some of the declarations outside of the try-catch block) just to avoid leaking the stream on errors?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
```diff
@@ -209,4 +213,24 @@ public static String getNodeIpV4Address(V1Node node) {
...
+            boolean includeJobDescriptor
+    ) {
+        String encodedContainerInfo = Base64.getEncoder().encodeToString(containerInfoData);
+
+        Map<String, String> annotations = new HashMap<>(passthroughAttributes);
+        annotations.putAll(PerformanceToolUtil.toAnnotations(job));
+        annotations.put("containerInfo", encodedContainerInfo);
+
+        if (includeJobDescriptor) {
+            String jobDescriptorJson = CommonObjectMappers.writeValueAsString(ObjectMappers.storeMapper(), job.getJobDescriptor());
```
fabiokung: minor: use `ObjectMappers.storeMapper()#writeValue(OutputStream, Object)` to save some heap space, avoid `String` allocations and unnecessary UTF-16 shenanigans.

That way you can also pass in a `GZIPOutputStream` directly to the `writeValue()` method and allow chunked encoding of the whole jobDescriptor.

titus-common/src/test/java/com/netflix/titus/common/util/StringExtTest.java
```diff
@@ -81,4 +81,11 @@ public void testParseDurationList() {
                 Duration.ofMillis(1), Duration.ofMillis(2), Duration.ofMillis(3)
         );
     }
+
+    @Test
+    public void testGzipAndBase64Encode() {
+        String input = "{aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa:bbbbbbbbbbbbbbbbbbbbbbbbbbb}";
+        String output = StringExt.gzipAndBase64Encode(input);
+        assertThat(output.length()).isLessThan(input.length());
```
joshi-keyur: Looks like this assertion is unique to the input in this case.
Should we generalize the test and possibly validate the decoded-gunzipped output is the same as input?
Also empty string is an interesting case since its output is non-empty. Maybe also handle that instead of returning an empty string. Curious to know what you think.

titus-common/src/main/java/com/netflix/titus/common/util/StringExt.java
corindwyer: I like this idea, but will address it in a followup PR.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
corindwyer: I like this idea, but will address it in a followup PR.

titus-common/src/test/java/com/netflix/titus/common/util/StringExtTest.java
corindwyer: I didn't go do that because I didn't write the decode and ungzip method, but I will think about this some more and potentially change it in a followup PR.

PR #805, opened 2020-03-11
Archaius2 dynamic proxy for Spring environment

tbak:


titus-common/src/main/java/com/netflix/titus/common/util/archaius2/SpringProxyInvocationHandler.java
```diff
@@ -0,0 +1,216 @@
...
+        if (prefix == null) {
+            Configuration configurationAnnotation = apiInterface.getAnnotation(Configuration.class);
+            effectivePrefix = configurationAnnotation != null ? configurationAnnotation.prefix() : null;
+        }
+        this.prefix = StringExt.isEmpty(effectivePrefix) ? "" : (effectivePrefix.endsWith(".") ? effectivePrefix : effectivePrefix + '.');
+        this.environment = environment;
+
+        Map<Method, MethodHandler> methodWrappers = new HashMap<>();
+        for (Method method : apiInterface.getMethods()) {
+            Preconditions.checkArgument(method.getParameterCount() == 0, "Method with no parameters expected");
```
fabiokung: wouldn't it be safer to just ignore this method and consider it not a configurable property? Especially as we are looking at all methods, including the inherited ones from parent interfaces.

tbak: We cannot just ignore them. We could either:
1. execute them if a default implementation is provided (our type is interface not a class)
2. throw an exception if the method is called by the client, but do not fail eagerly in the constructor

PR #807, opened 2020-03-12
Support Kube scheduler zone balancing

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultTaskToPodConverter.java
```diff
@@ -81,6 +84,12 @@
     private static final String ARN_SUFFIX = ":role/";
     private static final Pattern IAM_PROFILE_RE = Pattern.compile(ARN_PREFIX + "(\\d+)" + ARN_SUFFIX + "\\S+");
 
+    /**
+     * Max allowed skew for zone load balancing as a soft constraint. As we do not want to prevent placement in any
+     * case, it must be >= max job size.
+     */
+    private static final int SOFT_MAX_SKEW = 100_000;
```
fabiokung: minor: configurable

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultTaskToPodConverter.java
```diff
@@ -331,4 +341,24 @@ private void setJobAcceptedTimestamp(TitanProtos.ContainerInfo.Builder container
...
+
+        V1TopologySpreadConstraint constraint = new V1TopologySpreadConstraint()
+                .topologyKey(KubeConstants.NODE_LABEL_ZONE)
+                .labelSelector(new V1LabelSelector().matchLabels(Collections.singletonMap(KubeConstants.POD_LABEL_JOB_ID, job.getId())));
+
+        if (hard) {
+            constraint.maxSkew(1).whenUnsatisfiable("DoNotSchedule");
+        } else {
+            constraint.maxSkew(SOFT_MAX_SKEW).whenUnsatisfiable("ScheduleAnyway");
+        }
```
fabiokung: nice 👏

fabiokung:
@tbak given jobs on farzones always have an explicit hard constraint on the farzone id, do we still need a toleration in `DefaultTaintTolerationFactory#resolveAvailabilityZoneToleration(Job)`?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultTaskToPodConverter.java
tbak: It is set to a very high value so it is always larger than any job.
Would you like to make it configurable, in case we have such big jobs, or to actually have a skew < job size. In the latter case, it should be computed as a % of desired job size.

fabiokung: just a minor quality of life change in case we observe skew issues when running this and need to tweak it. Given we are using this with `.whenUnsatisfiable("ScheduleAnyway")` maybe it doesn't matter much

tbak:
@fabiokung It is not strictly required. I would keep it for some time, in case we need more strict rules around farzones and other special resources.

PR #810, opened 2020-03-16
Upgrade Spring/SpringBoot

tbak:


build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     dependencies {
         // force a more recent spring since some nebula plugins were bringing an older version
         // before springboot plugins had any saying
-        classpath "org.springframework:spring-core:5.1.5.RELEASE"
+        classpath "org.springframework:spring-core:5.2.4.RELEASE"
```
corindwyer: can this use a global variable?

tbak: I am not sure, so I am keeping this as it was previously.

PR #811, opened 2020-03-17
More routing options for Fenzo/Kube scheduler

tbak:
With this PR, we can route using an application name, capacity group, tier or job type.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/feature/FeatureGuards.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/feature/FeatureGuards.java
@@ -19,10 +19,31 @@ package com.netflix.titus.common.util.feature;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
+import com.netflix.titus.common.util.feature.FeatureGuard.FeatureGuardResult;
+
 public class FeatureGuards {
 
-    public static <T> Predicate<T> toPredicate(FeatureGuard<T> featureGuard) {
-        return value -> featureGuard.matches(value) == FeatureGuard.FeatureGuardResult.Approved;
+    public static <T> Predicate<T> toPredicate(FeatureGuard<T>... featureGuard) {
+        if (featureGuard.length == 0) {
+            return value -> false;
+        }
+        if (featureGuard.length == 1) {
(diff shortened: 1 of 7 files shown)


titus-common/src/main/java/com/netflix/titus/common/util/feature/FeatureGuards.java
```diff
@@ -19,10 +19,31 @@
 import java.util.function.Function;
 import java.util.function.Predicate;
 
+import com.netflix.titus.common.util.feature.FeatureGuard.FeatureGuardResult;
+
 public class FeatureGuards {
 
-    public static <T> Predicate<T> toPredicate(FeatureGuard<T> featureGuard) {
-        return value -> featureGuard.matches(value) == FeatureGuard.FeatureGuardResult.Approved;
+    public static <T> Predicate<T> toPredicate(FeatureGuard<T>... featureGuard) {
```
corindwyer: Can you document the intended behavior of this method? I can see it being confusing due to having to figure out how multiple feature guards are composed together.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/FeatureFlagModule.java
```diff
@@ -155,16 +158,22 @@ public FeatureGuardWhiteListConfiguration getJobActivityPublishFeatureGuardConfi
...
     @Singleton
     @Named(KUBE_SCHEDULER_FEATURE)
-    public Predicate<JobDescriptor> getKubeSchedulerFeaturePredicate(@Named(KUBE_SCHEDULER_FEATURE) FeatureGuardWhiteListConfiguration configuration) {
-
-        Predicate<JobDescriptor> routingPredicate = FeatureGuards.toPredicate(
-                FeatureGuards.fromField(
-                        JobDescriptor::getApplicationName,
-                        FeatureGuards.newWhiteListFromConfiguration(configuration).build()
-                )
+    public Predicate<Pair<JobDescriptor, ApplicationSLA>> getKubeSchedulerFeaturePredicate(ConfigProxyFactory factory) {
```
corindwyer: Can you write a test with common cases for this predicate?

PR #812, opened 2020-03-17
Keep task retryer in the direct Kube integration

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
```diff
@@ -316,7 +318,38 @@ private Task fillInMissingStates(V1Pod pod, Task task) {
...
+
+        List<TaskStatus> newStatusHistory = new ArrayList<>(task.getStatusHistory());
+        newStatusHistory.add(TaskStatus.newBuilder()
+                .withState(TaskState.Launched)
+                .withReasonCode(TaskStatus.REASON_STATE_MISSING)
+                .withReasonMessage("Filled in")
+                .withTimestamp(startAtTimestamp)
+                .build()
+        );
+        newStatusHistory.sort(Comparator.comparing(ExecutableStatus::getState));
```
fabiokung: minor: can probably extract the comparator here as a constant to avoid allocations

PR #813, opened 2020-03-17
TitusFederation on SpringBoot (initial)

tbak:


titus-server-federation-springboot/src/main/resources/application.yaml
```diff
@@ -0,0 +1 @@
+titus.federation.cells: cell1=titusapi-devtbakcell001-gateway.cluster.us-east-1.test.cloud.netflix.net:7007
```
fabiokung: this should probably not be checked in here

PR #817, opened 2020-03-19
Add configurable Fenzo taints

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -351,24 +351,49 @@ private void nodeUpdated(V1Node node) {
...
+            }
+        }
+
+        logger.debug("Owned by fenzo (all taints tolerated): {}", node.getMetadata().getName());
+        return true;
+    }
+
+    /**
+     * Returns true if there is {@link KubeConstants#TAINT_SCHEDULER} taint with {@link KubeConstants#TAINT_SCHEDULER_VALUE_FENZO} value
+     * or this taint is missing (no explicit scheduler taint == Fenzo).
```
fabiokung: the code below seems to be returning `false` when there is no explicit scheduler taint (i.e. `schedulerTaintValues.size() < 1`). What is the expected behavior?

Some unit tests here would be nice.

tbak: Yes, this is correct. If there is no taint, we assume the node belongs to fenzo (see the comment).
We do not have unit tests for this class. I tested this extensively manually on a running system.

fabiokung: Let me rephrase: the comment here is not consistent with the behavior of the code on line 396. If a Node has taints, but none of them are `KubeConstants.TAINT_SCHEDULER` the return of this method will be `false` and the node will be considered not owned by Fenzo, which is the opposite of what the javadoc says.

fabiokung: If what I said is correct and I'm not missing anything, one more argument in favor of having a unit test for the behavior here (how taints are being used and expected).

tbak: It is a bug. I am adding a unit test for this function.

PR #821, opened 2020-03-25
Minor Improvements

corindwyer:
### Description of the Change
* Change jobDescriptor pod annotation payload to use the json API representation.
* Add ipv4 task attribute to be more explicit that it is an ipv4 address.
* Add ipv6 task attribute if the container has an ipv6 address.
* Add job id and job application name to passthrough attributes.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
@@ -33,17 +33,19 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import com.google.common.base.Strings;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
 import com.netflix.titus.api.jobmanager.JobConstraints;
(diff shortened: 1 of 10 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
```diff
@@ -234,8 +238,13 @@ public static String getNodeIpV4Address(V1Node node) {
         annotations.put("containerInfo", encodedContainerInfo);
 
         if (includeJobDescriptor) {
-            String jobDescriptorJson = CommonObjectMappers.writeValueAsString(ObjectMappers.storeMapper(), job.getJobDescriptor());
-            annotations.put("jobDescriptor", StringExt.gzipAndBase64Encode(jobDescriptorJson));
+            JobDescriptor grpcJobDescriptor = GrpcJobManagementModelConverters.toGrpcJobDescriptor(job.getJobDescriptor());
```
tbak: Does it mean this field is not processed yet by VK?

corindwyer: correct

PR #825, opened 2020-03-25
Make RDS timeout configurable in the Titus relocation service

tbak:
and increast the default from 30sec to 90sec


titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/workflow/step/MustBeRelocatedTaskStoreUpdateStep.java
```diff
@@ -119,7 +123,7 @@ private void updateInStore(List<TaskRelocationPlan> toUpdate) {
         Map<String, Optional<Throwable>> result;
         try {
             result = store.createOrUpdateTaskRelocationPlans(toUpdate)
-                    .timeout(STORE_UPDATE_TIMEOUT)
+                    .timeout(Duration.ofMillis(configuration.getRdsTimeoutMs()))
```
corindwyer: if you timeout the Mono operation, does the underlying sql query also get cancelled or is the query just going to run until completion?

tbak: I think it will not be cancelled. This request is run synchronously on an internal executor.

PR #834, opened 2020-04-02
Initial jooq flyway commit to connect to job activity

amitaekbote:
still a work in progress

### DIFF
--- a/titus-ext/jooqflyway/src/main/java/com/netflix/titus/ext/jooqflyway/jobactivity/RDSSSLSocketFactory.java
+++ b/titus-ext/jooqflyway/src/main/java/com/netflix/titus/ext/jooqflyway/jobactivity/RDSSSLSocketFactory.java
@@ -34,7 +34,7 @@ public class RDSSSLSocketFactory extends WrappedFactory {
             tmf.init(trustStore);
             SSLContext sslContext = SSLContext.getInstance("TLS");
             sslContext.init(null, tmf.getTrustManagers(), null);
-            _factory = sslContext.getSocketFactory();
+            factory = sslContext.getSocketFactory();
         } catch (Exception e) {
             throw new IllegalStateException("Cannot initialize RDS socket factory", e);
         }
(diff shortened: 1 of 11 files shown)


titus-ext/jooqflyway/src/main/java/com/netflix/titus/ext/jooqflyway/jobactivity/RDSSSLSocketFactory.java
```diff
@@ -0,0 +1,42 @@
...
+import java.security.KeyStore;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManagerFactory;
+
+import org.postgresql.ssl.WrappedFactory;
+
+public class RDSSSLSocketFactory extends WrappedFactory {
+    public RDSSSLSocketFactory() {
+        try {
+            KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
```
tbak, line 29: This code is duplicated (we should call into RdsUtil added below).

titus-ext/jooqflyway/src/main/java/com/netflix/titus/ext/jooqflyway/jobactivity/RdsUtils.java
```diff
@@ -0,0 +1,46 @@
...
+public final class RdsUtils {
+    /**
+     * Setting as a constant as it is not critical.
+     */
+    private static final String TRUST_STORE_PASSWORD = "titus123";
+
+    public static SSLSocketFactory createRdsSSLSocketFactory() {
+        try {
+            KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
+            try (InputStream inputStream = RdsUtils.class.getClassLoader().getResourceAsStream("RDS-2019.truststore")) {
```
tbak, line 34: For security reason, we should allow injecting the truststore file name and the password here. Thus internally we can use a different file, and a different (ideally encrypted) password.

tbak: We should reference the attached truststure in this repo for running the integration tests only.

titus-ext/jooqflyway/src/main/java/com/netflix/titus/ext/jooqflyway/jobactivity/RDSSSLSocketFactory.java
amitaekbote: fixed it.

titus-ext/jooqflyway/src/main/java/com/netflix/titus/ext/jooqflyway/jobactivity/RdsUtils.java
amitaekbote: will fix this in the following pr, finally able to test this. was fighting gradle builds. definitely needs more work here

PR #835, opened 2020-04-10
TitusFederation SpringBoot REST resources

tbak:


titus-server-federation/src/main/java/com/netflix/titus/federation/endpoint/rest/FederationV2CapacityGroupSpringResource.java
```diff
@@ -0,0 +1,100 @@
...
+     */
+    @GetMapping
+    public List<ApplicationSlaRepresentation> getApplicationSLAs(@RequestParam(name = "extended", defaultValue = "false") boolean extended) {
+        Either<List<ApplicationSlaRepresentation>, WebApplicationException> result = CellWebClientConnectorUtil.doGetAndMerge(
+                cellWebClientConnector,
+                API_PATH + "?extended=" + extended,
+                APPLICATION_SLA_LIST_TP,
+                configuration.getRestRequestTimeoutMs()
+        );
+        if (result.hasError()) {
```
fabiokung, line 73: minor: an utility method like `<T, E extends Exception> T FunctionExt::must(Either<T, E>) throws E` would be nice to reduce repetition of dealing with the result here, e.g.: `return must(result)`

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/common/rest/Responses.java
```diff
@@ -79,6 +90,15 @@ public static Response fromCompletable(Completable completable, Response.Status
         return Response.status(statusCode).build();
     }
 
+    public static ResponseEntity<Void> fromCompletable(Completable completable, HttpStatus statusCode) {
+        try {
+            completable.await(1, TimeUnit.MINUTES);
```
fabiokung, line 95: why no `REST_TIMEOUT_DURATION`?

PR #836, opened 2020-04-12
Moving elastic search http client into titus-common module

amit-git:
### Moving Elastic Search Client code to titus-common module

Elastic search client interface along with its http based implementation and the corresponding model classes are reusable beyond its current usage in a Titus supplementary component - taskspublisher. This involves a bunch of refactoring around taskspublisher code as well as the test code.


titus-common-ext/elasticsearch/src/main/java/com/netflix/titus/ext/elasticsearch/EsClientConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.titus.ext.elasticsearch;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class EsClientConfiguration {
```
tbak, line 22: I think for consistency we should use the Archaius2 annotations and factories provided by `Archaius2Ext`.

titus-common-ext/elasticsearch/src/test/java/com/netflix/titus/ext/elasticsearch/DefaultEsWebClientFactoryTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.assertj.core.api.Java6Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class DefaultEsWebClientFactoryTest {
+    private static final Logger logger = LoggerFactory.getLogger(DefaultEsWebClientFactoryTest.class);
```
tbak, line 27: Not used?

titus-common-ext/elasticsearch/src/test/java/com/netflix/titus/ext/elasticsearch/EsExternalResource.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        buildEsClientConfiguration(esHostName, esPort);
+    }
+
+    public EsClientConfiguration getEsClientConfiguration() {
+        return esClientConfiguration;
+    }
+
+    private void buildEsClientConfiguration(String esHostName, int esPort) {
```
tbak, line 60: With Archaius2 interface getting defaults is a one liner.

titus-common-ext/elasticsearch/src/test/java/com/netflix/titus/ext/elasticsearch/TestDoc.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public void setState(String state) {
+        this.state = state;
+    }
+
+    public void setTs(long ts) {
+        this.ts = ts;
+    }
+
+    public TestDoc(String id, String state, long ts) {
```
tbak, line 38: Minor. Constructors should be first, followed by getters and last setters.

titus-common-ext/elasticsearch/src/main/java/com/netflix/titus/ext/elasticsearch/model/BulkEsIndexResp.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.netflix.titus.ext.elasticsearch.model;
+
+import java.util.List;
+
+/**
+ * Elastic search data model as defined by REST API documentation
+ * https://www.elastic.co/guide/en/elasticsearch/reference/master/rest-apis.html
+ */
+public class BulkEsIndexResp {
```
fabiokung, line 24: are these POJOs auto-generated code? Can they be immutable? They are inconsistent in some other ways as well, it would be nice to normalize them all with:

* many don't have private fields
* non-standard coding conventions for field and getter/setter methods on some (with underscores)
* only some implement `toString()`
* do they all need `equals()` and `hashcode()`?

amit-git: Good catch.
My initial thought on this was to not make assumptions about Webclient strategy it uses for deserializing ES responses into POJOs. But it seems like spring docs make it clear that it is going to depend on Jackson bindings for WebClient support.

PR #837, opened 2020-04-17
Adding schedulerName attribute to ApplicationSLA

joshi-keyur:
### Description of the Change

By introducing scheduler attribute as part of ApplicationSLA, this change paves a path for moving  from default scheduler (fenzo) to kube-scheduler allowing control planes components to filter and limit processing to exclusive sets of ApplicationSLAs.


titus-api/src/main/java/com/netflix/titus/api/model/ApplicationSLA.java
```diff
@@ -30,20 +38,29 @@
...
      * fulfill this requirement.
      */
     private int instanceCount;
 
-    public ApplicationSLA(String appName,
-                          Tier tier,
-                          ResourceDimension resourceDimension,
-                          int instanceCount) {
+    @JsonCreator
+    public ApplicationSLA(@JsonProperty("appName") String appName,
```
tbak, line 50: Jackson annotation are in ApplicationSlaMixIn. You should not add them here.

titus-server-master/src/main/java/com/netflix/titus/master/service/management/internal/DefaultApplicationSlaManagementService.java
```diff
@@ -52,6 +52,12 @@ public DefaultApplicationSlaManagementService(CapacityMonitoringService capacity
                 .toList().toBlocking().firstOrDefault(Collections.emptyList());
     }
 
+    @Override
+    public Collection<ApplicationSLA> getApplicationSLAsForScheduler(String schedulerName) {
+        return storage.findBySchedulerName(schedulerName).onErrorReturn(t -> null)
```
tbak, line 57: Does `.onErrorReturn(t -> null)` emit null value or  nothing at all?

titus-api/src/main/java/com/netflix/titus/api/model/ApplicationSLA.java
joshi-keyur: Fixed. Thanks

titus-server-master/src/main/java/com/netflix/titus/master/service/management/internal/DefaultApplicationSlaManagementService.java
joshi-keyur: It emits a null value that would appear in the return collection. This behavior is borrowed from the other get methods in this service. Should we do anything differently here and possibly at other places?

tbak: Yes, this is incorrect. Could you also fix the other method? What happens if a list of null is returned?

joshi-keyur: Changed `onErrorReturn()` to `doOnError()` and added error logging. Please review.

titus-api/src/main/java/com/netflix/titus/api/store/v2/ApplicationSlaStoreCache.java
```diff
@@ -32,15 +36,17 @@
 
 /**
  * Caching proxy for {@link ApplicationSlaStore}. It loads all data on startup in a blocking mode to fail fast,
- * in case the is any problem with the storage.
+ * in case there is any problem with the storage.
```
corindwyer, line 39: there is -> there are

titus-api/src/main/java/com/netflix/titus/api/store/v2/ApplicationSlaStoreCache.java
```diff
@@ -50,12 +56,22 @@ public ApplicationSlaStoreCache(ApplicationSlaStore delegate) {
...
         this.cache = loadCache(delegate);
+        this.cacheBySchedulerName = loadSchedulerMapCache(this.cache);
         return Observable.empty();
     }
 
     @Override
     public Observable<Void> create(ApplicationSLA applicationSLA) {
-        return delegate.create(applicationSLA).doOnCompleted(() -> cache.put(applicationSLA.getAppName(), applicationSLA));
+        return delegate.create(applicationSLA).doOnCompleted(() ->
+        {
```
corindwyer, line 66: autoformat class should move this curly up a line

PR #840, opened 2020-04-22
Change task dependency rules to avoid issues caused by the recent gradle upgrade

tbak:
protobuf tasks are now created lazily which broke the previous override rule setup

### DIFF
--- a/titus-common-grpc-api/build.gradle
+++ b/titus-common-grpc-api/build.gradle
@@ -22,13 +22,11 @@ dependencies {
 }
 
 if (project.hasProperty("idlLocal")) {
-    task extractProto {
-        doLast {
-            copy {
-                from '../../titus-api-definitions/src/main/proto'
-                into new File(buildDir, "extracted-protos/main")
-            }
-        }
+    tasks.register('copyProtos', Exec) {
+        commandLine "/bin/bash", "-c", "cp -R " + new File(buildDir, '../../../titus-api-definitions/src/main/proto/netflix')  + " " + new File(buildDir, "extracted-protos/main")
+    }
+    afterEvaluate {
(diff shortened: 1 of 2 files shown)


titus-common-grpc-api/build.gradle
```diff
@@ -22,13 +22,11 @@ dependencies {
...
 if (project.hasProperty("idlLocal")) {
-    task extractProto {
-        doLast {
-            copy {
-                from '../../titus-api-definitions/src/main/proto'
-                into new File(buildDir, "extracted-protos/main")
-            }
-        }
+    tasks.register('copyProtos', Exec) {
+        commandLine "/bin/bash", "-c", "cp -R " + new File(buildDir, '../../../titus-api-definitions/src/main/proto/netflix')  + " " + new File(buildDir, "extracted-protos/main")
```
fabiokung, line 26: could this be done inside the gradle JVM instead of shelling out here? Shelling out has all sorts of implications and assumptions about where it is running

E.g.: https://docs.gradle.org/current/userguide/working_with_files.html#sec:moving_files_example or directly registering a task of `type: Copy` instead of `type: Exec`

tbak: I will try ant task. Copy does not work (gradle decides to skip the task).

tbak: I am not getting any success with ant task as well.
Anyway this property is only set in the local builds to use local version of titus-api-definitions. It is not used in the regular build process.

corindwyer: Yeah, it was strange to me that you were shelling out instead of using the gradle task but I did not know the gradle task was the issue in the first place.

PR #843, opened 2020-04-28
Fix producer/consumer jooq configs

amitaekbote:
### Description of the Change

**FIXME**: please add at least one sentence explaining why changes here are being made.


titus-ext/jooqflyway/src/main/java/com/netflix/titus/ext/jooqflyway/jobactivity/producer/JooqJobActivityProducerConnectorComponent.java
```diff
@@ -0,0 +1,54 @@
...
+import org.jooq.SQLDialect;
+import org.jooq.impl.DSL;
+import org.postgresql.PGProperty;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
+
+@Configuration
+public class JooqJobActivityProducerConnectorComponent {
```
tbak, line 18: What is the difference between this database connection and the other one?

amitaekbote: It was complaining about having two data sources with one configuration, I tried separating it out but after talking to runtime, marking one as primary with correct qualifiers seems to work and is the recommended way

PR #844, opened 2020-05-07
Add a Kube backend constraint

tbak:
to restrict placement to nodes running a specific Kube backend version

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
@@ -235,7 +235,7 @@ public class KubeNotificationProcessor {
                     Task fixedTask = fillInMissingStates(pod, updatedTask);
                     Task taskWithPlacementData = JobManagerUtil.attachPlacementData(fixedTask, executorDetailsOpt);
                     Task taskWithNodeMetadata = node.map(n -> attachNodeMetadata(taskWithPlacementData, n)).orElse(taskWithPlacementData);
-                    Task taskWithAnnotations = addMissingAttributes(pod, updatedTask);
+                    Task taskWithAnnotations = addMissingAttributes(pod, taskWithNodeMetadata);
 
                     return Optional.of(taskWithAnnotations);
                 },
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
@@ -26,6 +26,7 @@ import java.util.Optional;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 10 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeConstants.java
```diff
@@ -83,18 +90,23 @@ private KubeConstants() {
...
-    public static final String TAINT_TIER = TITUS_TAINT_DOMAIN + "tier";
+    public static final String TAINT_TIER = TITUS_NODE_DOMAIN + "tier";
+
+    /**
+     * Taint added to each GPU instance.
+     */
+    public static final String TAINT_GPU_INSTANCE = TITUS_NODE_DOMAIN + "gpu";
 
     /**
      * Taint added to each GPU instance.
```
corindwyer, line 107: javadoc is wrong

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
```diff
@@ -235,7 +235,7 @@ public void shutdown() {
                     Task fixedTask = fillInMissingStates(pod, updatedTask);
                     Task taskWithPlacementData = JobManagerUtil.attachPlacementData(fixedTask, executorDetailsOpt);
                     Task taskWithNodeMetadata = node.map(n -> attachNodeMetadata(taskWithPlacementData, n)).orElse(taskWithPlacementData);
-                    Task taskWithAnnotations = addMissingAttributes(pod, updatedTask);
+                    Task taskWithAnnotations = addMissingAttributes(pod, taskWithNodeMetadata);
```
fabiokung, line 238: nice catch! Can you please add test cases to `KubeNotificationProcessorTest` so we ensure no regressions going forward?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeConstants.java
```diff
@@ -41,21 +41,33 @@ private KubeConstants() {
...
+     */
+    public static final String TITUS_V3_JOB_DOMAIN = "v3.job.titus.netflix.com/";
+
     /*
-     * Titus node labels.
+     * Titus pod labels.
      */
 
-    public static final String NODE_LABEL_MACHINE_ID = "titus/machine-id";
+    public static final String POD_LABEL_JOB_ID = TITUS_V3_JOB_DOMAIN + "jobId";
```
fabiokung, line 58: I think we should stick with DNS naming conventions for annotations and labels, so all lowercase characters:

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

I suggest `.../job-id` and `.../task-id`. See how `machine-id` and `machine-group` were chosen below, rather than `machineId`. Alternatively, it could also be `.../jobid` and `.../taskid`.

tbak: I like the idea. Changing this to `job-id` and `task-id`.

PR #848, opened 2020-05-20
Configurable KubeApiServerIntegrator orphaned pods token bucket

tbak:


titus-common/src/main/java/com/netflix/titus/common/util/limiter/tokenbucket/internal/FixedIntervalTokenBucketSupplier.java
```diff
@@ -0,0 +1,147 @@
...
+        return activeConfiguration.getCapacity() == configuration.getCapacity()
+                && activeConfiguration.getInitialNumberOfTokens() == configuration.getInitialNumberOfTokens()
+                && activeConfiguration.getIntervalMs() == configuration.getIntervalMs()
+                && activeConfiguration.getNumberOfTokensPerInterval() == configuration.getNumberOfTokensPerInterval();
+    }
+
+    private ActiveConfiguration reload() {
+        boolean same;
+        synchronized (lock) {
+            same = activeConfiguration != null && isSame();
```
joshi-keyur, line 73: Should the null check be also part of the isSame method?

titus-common/src/main/java/com/netflix/titus/common/util/limiter/tokenbucket/FixedIntervalTokenBucketConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.common.util.limiter.tokenbucket;
+
+import com.netflix.archaius.api.annotations.DefaultValue;
+
+public interface FixedIntervalTokenBucketConfiguration {
```
joshi-keyur, line 21: Should we add Configuration prefix for better namespacing here?

tbak: Prefix must be provided by a client, but we can create a default value here.

titus-common/src/main/java/com/netflix/titus/common/util/limiter/tokenbucket/internal/FixedIntervalTokenBucketSupplier.java
tbak: Good idea. I will change it as suggested.

PR #850, opened 2020-05-21
Feature/kube informer sync

tbak:
This PR adds a new metric to track an informer (pod, node, etc) sync status.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/client/KubeApiClients.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/client/KubeApiClients.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/client/KubeApiClients.java
```diff
@@ -0,0 +1,89 @@
...
+                        .addInterceptor(metricsInterceptor)
+                        .readTimeout(readTimeoutMs, TimeUnit.SECONDS)
+                        .build()
+        );
+        return client;
+    }
+
+    public static SharedInformerFactory createSharedInformerFactory(String threadNamePrefix, ApiClient apiClient) {
+        AtomicLong nextThreadNum = new AtomicLong(0);
+        return new SharedInformerFactory(apiClient, Executors.newCachedThreadPool(runnable -> {
```
fabiokung, line 83: use `ExecutorsExt.instrumentedCachedThreadPool(...)` to get thread pool metrics

tbak:
@corindwyer @fabiokung @joshi-keyur `Informer.hasSynced` is set once after the first successful sync, which is not what we are aiming for. I am going to merge this PR anyway, as it is better than nothing. I will continue on improving the monitoring, and make another PR when it is ready.

PR #858, opened 2020-06-04
add scheduling constraint for kube nodes

corindwyer:
### Description of the Change
Add scheduling constraint for kube nodes instead of trying to do lease management when conditions and taints change.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
@@ -329,18 +329,10 @@ public class KubeApiServerIntegrator implements VirtualMachineMasterService {
 
     private void nodeUpdated(V1Node node) {
         try {
-            boolean removeStopped = node.getStatus().getConditions().stream()
-                    .anyMatch(c -> c.getType().equalsIgnoreCase(STOPPED) && Boolean.parseBoolean(c.getStatus()));
-            boolean removeNotReady = node.getStatus().getConditions().stream()
-                    .anyMatch(c -> c.getType().equalsIgnoreCase(READY) && !Boolean.parseBoolean(c.getStatus()));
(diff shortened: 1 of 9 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -329,18 +329,10 @@ public void onDelete(V1Node node, boolean deletedFinalStateUnknown) {
 
     private void nodeUpdated(V1Node node) {
         try {
-            boolean removeStopped = node.getStatus().getConditions().stream()
```
tbak, line 332: Do we depend on GC process here to do the cleanup?

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/KubeConstraint.java
```diff
@@ -0,0 +1,154 @@
...
+            return VALID;
+        }
+
+        Optional<AgentInstance> instanceOpt = SchedulerUtils.findInstance(agentManagementService, schedulerConfiguration.getInstanceAttributeName(), targetVM);
+        if (!instanceOpt.isPresent()) {
+            return Failure.INSTANCE_NOT_FOUND.toResult();
+        }
+
+        String instanceId = instanceOpt.get().getId();
+        V1Node node = kubeApiFacade.getNodeInformer().getIndexer().getByKey(instanceId);
```
tbak, line 116: What if node is not there? We should check for null at least + double check that an exception is not thrown if a node is missing or due to other problems.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
corindwyer: Yes.

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/KubeConstraint.java
corindwyer: good catch. updated

PR #863, opened 2020-06-06
Support remote data plane region

joshi-keyur:
### Description of the Change

An aws configuration property is added for data plane region. When set, this property drives the AWS region the agent resources are fetched from. Otherwise, we fallback to default behavior.


titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonAutoScalingAsyncProvider.java
```diff
@@ -33,7 +34,11 @@
 
     @Inject
     public AmazonAutoScalingAsyncProvider(AwsConfiguration configuration, AWSCredentialsProvider credentialProvider) {
-        String region = configuration.getRegion().trim().toLowerCase();
+        String region = configuration.getDataPlaneRegion().trim().toLowerCase();
```
tbak, line 37: This code looks identical in all provider classes. We should move it to a utility class and share one implementation.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonClientProvider.java
```diff
@@ -34,7 +35,10 @@ public AmazonElasticLoadBalancingAsync getLoadBalancingClient(String accountId)
             synchronized (this) {
                 client = loadBalancerClients.get(accountId);
                 if (client == null) {
-                    String region = configuration.getRegion().trim().toLowerCase();
+                    String region = configuration.getDataPlaneRegion().trim().toLowerCase();
```
corindwyer, line 38: like above, this logic appears many times.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AmazonEC2AsyncProvider.java
```diff
@@ -33,7 +34,11 @@
 
     @Inject
     public AmazonEC2AsyncProvider(AwsConfiguration configuration, AWSCredentialsProvider credentialProvider) {
-        String region = configuration.getRegion().trim().toLowerCase();
+        String region = configuration.getDataPlaneRegion().trim().toLowerCase();
```
corindwyer, line 37: same as above

PR #864, opened 2020-06-08
Do not attempt scheduling if Kube integrator is not ready or healthy

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/batch/BatchDifferenceResolver.java
```diff
@@ -269,17 +269,19 @@ public BatchDifferenceResolver(
...
-                    missingTasks.add(BasicTaskActions.launchTaskInKube(
-                            kubeApiServerIntegrator,
-                            configuration,
-                            engine,
-                            runningJobView.getJob(),
-                            refTask,
-                            RECONCILER_CALLMETADATA.toBuilder().withCallReason("Launching task in Kube").build(),
-                            titusRuntime
-                    ));
+                if (kubeApiServerIntegrator.isReadyForScheduling()) {
```
fabiokung, line 272: minor: combine the `if` below into a single boolean logic operation (`&&`)

fabiokung: even better, maybe consider extracting the boolean logic chain into a separate method/function

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailureClassifier.java
```diff
@@ -153,6 +155,23 @@ private boolean processNoActiveAgent(T taskRequest, List<TaskAssignmentResult> a
         return true;
     }
 
+    private boolean processKubeAbiNotReady(T taskRequest, List<TaskAssignmentResult> assignmentResults,
```
fabiokung, line 158: typo `s/Abi/Api/`

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/batch/BatchDifferenceResolver.java
tbak: I was thinking about extracting this into a util function, but there would be too many arguments, which would not improved the code readability.

PR #868, opened 2020-06-10
Read archived data from active job/task tables if not found in the archive tables

tbak:
It is possible that finished jobs/tasks are not moved after they are finished to
the archive tables. This may happen during failover which happens
immediately after a job completes, but is not archived yet. To read such
abandoned items (which are ignored when loading active data into memory), we
have to make a fallback query in case the primary archive table does not contain
an entry.

### DIFF
--- a/titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraJobStore.java
+++ b/titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraJobStore.java
@@ -604,6 +604,15 @@ public class CassandraJobStore implements JobStore {
         ).toCompletable();
     }
 
+    /**
+     * This method reads data from the archive table, and if not found checks the active table for its existence.
+     * The latter is needed as sometimes a job may not be correctly archived, and we do not have a reconciliation process
+     * that would fix it.
+     * <p>
+     * This method should be only used to get state of a finished job. It would return a persisted state for an active job
+     * as well, as a side effect for the workaround implemented here. A caller should not piggyback on this behavior, as
+     * it may change at any point in time.
+     */
     @Override
     public Observable<Job<?>> retrieveArchivedJob(String jobId) {
         Observable<Job> action = retrieveEntityById(jobId, Job.class, retrieveArchivedJobStatement)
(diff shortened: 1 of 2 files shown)


titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraJobStore.java
```diff
@@ -606,43 +606,62 @@ public Completable deleteTask(Task task) {
...
-                    return (Job<?>) ObjectMappers.readValue(mapper, value, Job.class);
-                }));
+        Observable<Job> action = retrieveEntityById(jobId, Job.class, retrieveArchivedJobStatement)
+                .switchIfEmpty(retrieveEntityById(jobId, Job.class, retrieveActiveJobStatement))
+                .switchIfEmpty(Observable.error(JobStoreException.jobDoesNotExist(jobId)));
+        return (Observable) action;
     }
 
     @Override
     public Observable<Task> retrieveArchivedTasksForJob(String jobId) {
```
fabiokung, line 616: please add some documentation here (and consider renaming the method) since now it will also always returns active tasks for the job. It seems to be handled correctly by the only current caller, but it may cause confusion in the future

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraJobStore.java
```diff
@@ -606,43 +606,62 @@ public Completable deleteTask(Task task) {
...
-                            .map(value -> deserializeTask(value))
-                            .collect(Collectors.toList()));
-                }));
+        return retrieveArchivedTasksForJob(jobId, retrieveArchivedTaskIdsForJobStatement, retrieveArchivedTaskStatement)
+                .switchIfEmpty(retrieveArchivedTasksForJob(jobId, retrieveActiveTaskIdsForJobStatement, retrieveActiveTaskStatement));
+    }
+
+    private Observable<Task> retrieveArchivedTasksForJob(String jobId, PreparedStatement taskIdStatement, PreparedStatement taskStatement) {
+        return Observable.fromCallable(() -> taskIdStatement.bind(jobId).setFetchSize(Integer.MAX_VALUE))
+                .flatMap(retrieveActiveTaskIdsForJob ->
```
andrew-leung, line 623: Is the variable name correct? Seems like these could be active or archived depending on the statement provided.

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraJobStore.java
tbak: Active state should never be read like that (all of it is read on startup). But technically you are right that we could read active data with this method. I will add a comment documenting this.

titus-ext/cassandra/src/main/java/com/netflix/titus/ext/cassandra/store/CassandraJobStore.java
tbak: I will change this, and add a comment.

PR #873, opened 2020-06-16
Check if all required node attributes are present

tbak:
before adding a lease for a node to Fenzo


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -124,6 +126,15 @@
 
     static final String GC_UNKNOWN_PODS = "gcUnknownPods";
 
+    static final String NODE_ATTRIBUTE_ID = "id";
+    static final String NODE_ATTRIBUTE_HOST_IP = "hostIp";
+    static final String NODE_ATTRIBUTE_RES = "res";
+    private static final Set<String> REQUIRED_NODE_ATTRIBUTES = CollectionsExt.asSet(
```
corindwyer, line 132: id, res, hostIp, zone, region, itype

PR #875, opened 2020-06-18
Adding compliance check for container accountId and subnets

joshi-keyur:
### Description of the Change

When default values for accountId and subnets container attributes are present in the JobManagerConfiguration but not defined in the JobDescriptor (in container attributes), a violation is recorded. Consequently, the job descriptor is sanitized with the default values from the configuration properties.

Currently we are limiting the compliance check only when default values for these two attributes are present as configuration properties since not all deployment stacks may be ready for us to enforce this strict requirement. Note that we are not providing any extra safety guards yet to make sure the accountId and subnet list are compatible. It is possible that an invalid combination is used which will result in failure to run the task. This is the current behavior and  unchanged.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/FeatureRolloutPlans.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/FeatureRolloutPlans.java
@@ -76,4 +76,19 @@ public interface FeatureRolloutPlans {
             description = "Integrate Kube scheduler"
     )
     String KUBE_SCHEDULER_FEATURE = "kubeSchedulerFeature";
+
+    @FeatureRollout(
+            featureId = "accountId",
+            deadline = "10/01/2020",
+            description = "Jobs should explicity provide AWS accountId the container should launch in"
+    )
+    String CONTAINER_ACCOUNT_ID_REQUIRED_FEATURE = "accountId";
+
+    @FeatureRollout(
+            featureId = "subnets",
+            deadline = "10/01/2020",
+            description = "Each job specifying the accountId to run the containers in must also specify a list of " +
+                    "subnets available in the accountId for provisioning networking."
+    )
+    String SUBNETS_REQUIRED_FEATURE = "subnets";
 }
--- a/titus-api/src/main/java/com/netflix/titus/api/FeatureRolloutPlans.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/FeatureRolloutPlans.java
@@ -78,17 +78,10 @@ public interface FeatureRolloutPlans {
     String KUBE_SCHEDULER_FEATURE = "kubeSchedulerFeature";
 
     @FeatureRollout(
-            featureId = "accountId",
+            featureId = "accountIdAndSubnetsFeature",
             deadline = "10/01/2020",
-            description = "Jobs should explicity provide AWS accountId the container should launch in"
+            description = "Jobs should provide AWS accountId and corresponding subnets the container should launch in"
     )
-    String CONTAINER_ACCOUNT_ID_REQUIRED_FEATURE = "accountId";
-
-    @FeatureRollout(
-            featureId = "subnets",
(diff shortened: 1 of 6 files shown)


titus-api/src/main/java/com/netflix/titus/api/FeatureRolloutPlans.java
```diff
@@ -76,4 +76,19 @@
...
     String KUBE_SCHEDULER_FEATURE = "kubeSchedulerFeature";
+
+    @FeatureRollout(
+            featureId = "accountId",
+            deadline = "10/01/2020",
+            description = "Jobs should explicity provide AWS accountId the container should launch in"
+    )
+    String CONTAINER_ACCOUNT_ID_REQUIRED_FEATURE = "accountId";
+
+    @FeatureRollout(
```
tbak, line 87: Isn't account/subnets configuration the same feature?

titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/ExtendedJobSanitizer.java
```diff
@@ -151,6 +157,26 @@ public ExtendedJobSanitizer(JobManagerConfiguration jobManagerConfiguration,
                 }
             });
 
+            Map<String, String> defaultContainerAttributes = new HashMap<>();
+            violations.findViolation(CONTAINER_ACCOUNT_ID_REQUIRED_FEATURE).ifPresent(nonCompliance -> {
+                if (!StringExt.isEmpty(jobManagerConfiguration.getDefaultContainerAccountId())) {
+                    defaultContainerAttributes.put(JobAttributes.JOB_CONTAINER_ATTRIBUTE_ACCOUNT_ID, jobManagerConfiguration.getDefaultContainerAccountId());
+                }
+            });
+            violations.findViolation(SUBNETS_REQUIRED_FEATURE).ifPresent(nonCompliance -> {
```
tbak, line 166: What is a user sets an account id (non-default), but not subnets?  Setting default subnets will be incorrect.
I think we should validate/set this data together, as they must be mutually consistent.

titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/JobFeatureComplianceChecks.java
```diff
@@ -106,6 +111,39 @@
...
+            if (StringExt.isEmpty(defaultAccountId) || StringExt.isEmpty(defaultSubnets)) {
+                return Optional.empty();
+            }
+
+            String accountIdContainerAttribute = jobDescriptor.getContainer().getAttributes().get(JOB_CONTAINER_ATTRIBUTE_ACCOUNT_ID);
+            String subnetContainerAttribute = jobDescriptor.getContainer().getAttributes().get(JOB_CONTAINER_ATTRIBUTE_SUBNETS);
+
+            // Feature compliance is violated if either (i) both attributes are not set, or (ii) accountId is set to the default value but the subnets value is not set.
+            // In either case, we should take action in response to this violation.
+            if ((StringExt.isEmpty(accountIdContainerAttribute) || defaultAccountId.equals(accountIdContainerAttribute)) && StringExt.isEmpty(subnetContainerAttribute)) {
```
tbak, line 135: If I understand it correctly, setting accontId parameter to the default account id is non-compliant?

joshi-keyur: Non compliant if:
(1). default account_id and subnets are set in the configuration (expected for the stack),
AND
(2a). account_id and subnets are empty in the job descriptor, or
(2b). account_id is set to the default account_id but subnets is not set in the job descriptor

All other cases are given a pass under current/default behavior.

PR #878, opened 2020-06-24
Use KubeClient async API to launch pods

tbak:
This update is limited to `KubeApiServerIntegrator`. We should refactor KubeScheduler code path after verifying that the async API is stable enough. The sync/async mode is controlled via persisted property.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
@@ -90,6 +90,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import reactor.util.concurrent.Queues;
 import rx.Observable;
 import rx.subjects.PublishSubject;
 import rx.subjects.Subject;
@@ -255,7 +256,7 @@ public class KubeApiServerIntegrator implements VirtualMachineMasterService {
     public void launchTasks(List<TaskInfoRequest> requests, List<VirtualMachineLease> leases) {
         Stopwatch stopwatch = Stopwatch.createStarted();
         if (directKubeConfiguration.isAsyncApiEnabled()) {
-            launchTasksAsync(requests);
+            launchTasksConcurrently(requests);
             logger.info("Async pod launches completed: pods={}, elapsed={}[ms]", requests.size(), stopwatch.elapsed(TimeUnit.MILLISECONDS));
         } else {
             launchTasksSync(requests);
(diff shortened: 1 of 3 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -259,6 +278,36 @@ public void launchTasks(List<TaskInfoRequest> requests, List<VirtualMachineLease
         }
     }
 
+    private void launchTasksAsync(List<TaskInfoRequest> requests) {
```
fabiokung, line 281: add docs explaining that this method will still block until all requests are finished (so we still expect the Fenzo thread to be blocked while all requests are in-flight, but requests will be executed concurrently)

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -259,6 +278,36 @@ public void launchTasks(List<TaskInfoRequest> requests, List<VirtualMachineLease
...
+            Mono<Void> podAddAction = KubeUtil
+                    .<V1Pod>toReact(handler -> kubeApiFacade.getCoreV1Api().createNamespacedPodAsync(
+                            KUBERNETES_NAMESPACE, v1Pod, null, null, null, handler
+                    ))
+                    .doOnSubscribe(subscription -> {
+                        launchTaskCounter.increment();
+                        logger.info("creating pod: {}", v1Pod);
+                        podSizeMetrics.record(KubeUtil.estimatePodSize(v1Pod));
+                    })
+                    .timeout(Duration.ofMillis(directKubeConfiguration.getKubeApiClientTimeoutMs()))
```
fabiokung, line 294: I suggest adding a `doOnComplete()` callback to record latencies on each individual operation (`DefaultSchedulingService` will still record the latency of the whole operation until all requests are finished).

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -259,6 +278,36 @@ public void launchTasks(List<TaskInfoRequest> requests, List<VirtualMachineLease
...
+                    .cast(Void.class)
+                    .onErrorResume(error -> {
+                        logger.error("Unable to create pod with error: {}", KubeUtil.toErrorDetails(error), error);
+                        return Mono.empty();
+                    });
+            podAddActions.add(podAddAction);
+        }
+
+        try {
+            Flux.merge(podAddActions).blockLast();
```
fabiokung, line 305: consider adding a concurrency limit to the `merge` so we avoid overloading kube-apiserver during big bursts

fabiokung: also consider `mergeDelayError` to allow partial failures mirroring the previous behavior

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
tbak: We record request level latencies in okhttp interceptor.

PR #881, opened 2020-07-13
Fix scheduling classifier handling of ExclusiveHost hard constraint failures

fabiokung:
A failure for a task with the `ExclusiveHost` hard constraint is different from other tasks failing on a machine because it is already running another task with `ExclusiveHost`.

Both failure modes are classified with different priority to avoid prematurely triggering the ClusterAgentAutoscaler, when tasks that don't have the hard constraint are failing due to other transient scheduling failures.

Automated testing of this part of the codebase is unfortunately pretty difficult due to some Fenzo classes not being public and/or mockable, so I'm punting on junit tests.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailure.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailure.java
@@ -81,6 +81,8 @@ public class TaskPlacementFailure {
 
         /**
          * Task not launched due to other tasks with the exclusiveHost hard constraint on the same agent. It has the lowest priority.
+         *
+         * @see TaskPlacementFailureClassifier for how {@link FailureKind failure kinds} are used and what are their priorities
          */
         ExclusiveHost,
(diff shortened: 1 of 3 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/scheduler/TaskPlacementFailure.java
```diff
@@ -79,31 +79,46 @@
          */
         OpportunisticResource,
 
+        /**
+         * Task not launched due to other tasks with the exclusiveHost hard constraint on the same agent. It has the lowest priority.
+         */
+        ExclusiveHost,
```
tbak, line 85: Is the priority defined by the declaration order in the enum type?

fabiokung: No, the priority is defined by how `TaskPlacementFailureClassifier` uses each `FailureKind` defined here. I will add more comments to reflect that.

PR #885, opened 2020-07-17
kube-scheduler: case insensitive constraint names

fabiokung:
Input in jobDescriptors does not consider constraint names to be case sensitive. Likely the same for constraint values, but since some of them are identifiers, I opted to leave them untouched.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/CollectionsExt.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/CollectionsExt.java
@@ -595,6 +595,31 @@ public final class CollectionsExt {
         return -from - 1;
     }
 
+    /**
+     * Convert all keys to lower case, the behavior is undefined when values clash, and only one will be kept
+     */
+    public static <V> Map<String, V> toLowerCaseKeys(Map<String, V> entries) {
+        return entries.entrySet().stream()
+                .collect(Collectors.toMap(
+                        e -> e.getKey().toLowerCase(),
+                        Map.Entry::getValue,
+                        (o, n) -> n
+                ));
+    }
+
+    /**
+     * Convert all keys to lower case, the behavior is undefined when values clash, and only one will be kept
+     */
+    public static <V> Map<String, V> toLowerCaseKeys(Map<String, V> entries, Supplier<Map<String, V>> mapSupplier) {
+        return entries.entrySet().stream()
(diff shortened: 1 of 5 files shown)


titus-common/src/main/java/com/netflix/titus/common/util/CollectionsExt.java
```diff
@@ -595,6 +595,31 @@ private CollectionsExt() {
         return -from - 1;
     }
 
+    /**
+     * Convert all keys to lower case, the behavior is undefined when values clash, and only one will be kept
+     */
+    public static <V> Map<String, V> toLowerCaseKeys(Map<String, V> entries) {
+        return entries.entrySet().stream()
```
tbak, line 602: Direct for loop for better performance, as this code might be used in many places?

tbak: Is there a unit test for these methods?

titus-server-master/src/main/java/com/netflix/titus/master/scheduler/DefaultSchedulingService.java
```diff
@@ -531,6 +531,9 @@ private void processTaskSchedulingFailureCallbacks(Map<FailureKind, Map<V3Queuea
     }
 
     private void launchTasks(TaskAssignments taskAssignments) {
+        if (taskAssignments.getCount() <= 0) {
```
corindwyer, line 534: This isn't entirely a no-op as it sets a gauge

corindwyer: I guess it doesn't matter though as it is actually adding a 0 to the AtomicLong which is getting polled.

fabiokung: exactly, this mirrors the previous behavior better, before #881

titus-common/src/main/java/com/netflix/titus/common/util/CollectionsExt.java
fabiokung: the same performance argument could be made to any other of the `CollectionsExt` methods, I think this is fine, I will leave it as is for now to avoid early optimization.

Good point about the unit test, I will add one.

PR #887, opened 2020-07-20
Spring REST resources minor bug fixes

tbak:


titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/appscale/AppAutoScalingCallbackSpringResource.java
```diff
@@ -43,18 +43,18 @@ public AppAutoScalingCallbackSpringResource(AppAutoScalingCallbackService awsGat
         this.awsGatewayCallbackService = awsGatewayCallbackService;
     }
 
-    @GetMapping(path = "{scalableTargetDimensionId}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @GetMapping(path = "/{scalableTargetDimensionId}", produces = MediaType.APPLICATION_JSON_VALUE)
```
amit-git, line 46: Is the GetMapping builds path differently than javax.ws.rs.Path ? I would not have expected we'd need a '/' prefix for the path spec on the resource methods

tbak: No, it is the same. `/` is probably not needed, but I think it is more explicit.

PR #889, opened 2020-07-29
Extract GC controller logic from KubeApiServerIntegrator

corindwyer:
### Description of the Change
* Extract gc controller logic from KubeApiServerIntegrator into new GC controllers.
* Separate podAdd,podUpdate event handling from podDelete.
* In podDelete, add a missing KillInitiated status to indicate something outside of Titus terminate the container.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.master.kubernetes.controller;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.netflix.titus.api.agent.model.AgentInstance;
+import com.netflix.titus.api.agent.model.InstanceLifecycleState;
+import com.netflix.titus.api.agent.service.AgentManagementService;
+import com.netflix.titus.common.framework.scheduler.LocalScheduler;
+import com.netflix.titus.common.runtime.TitusRuntime;
+import com.netflix.titus.common.util.StringExt;
+import com.netflix.titus.common.util.limiter.tokenbucket.FixedIntervalTokenBucketConfiguration;
+import com.netflix.titus.common.util.time.Clock;
+import com.netflix.titus.master.mesos.kubeapiserver.KubeUtil;
+import com.netflix.titus.runtime.connector.kubernetes.KubeApiFacade;
+import io.kubernetes.client.openapi.models.V1Node;
+import io.kubernetes.client.openapi.models.V1NodeCondition;
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.netflix.titus.runtime.kubernetes.KubeConstants.NODE_LABEL_ACCOUNT_ID;
+import static com.netflix.titus.runtime.kubernetes.KubeConstants.READY;
+
+/**
+ * Responsible for deleting Kubernetes node objects when they no longer exist in agent management. This will eventually be moved
+ * to a new component.
+ */
+
+@Singleton
+public class NodeGcController extends BaseGcController<V1Node> {
+    public static final String NODE_GC_CONTROLLER = "nodeGcController";
+    public static final String NODE_GC_CONTROLLER_DESCRIPTION = "GC nodes that are not in agent management and are not reachable.";
+
+    private static final Logger logger = LoggerFactory.getLogger(NodeGcController.class);
+    private final AgentManagementService agentManagementService;
+    private final KubeApiFacade kubeApiFacade;
+    private final Clock clock;
+    private final KubeControllerConfiguration kubeControllerConfiguration;
+
+    @Inject
+    public NodeGcController(
+            TitusRuntime titusRuntime,
+            LocalScheduler scheduler,
+            @Named(NODE_GC_CONTROLLER) FixedIntervalTokenBucketConfiguration tokenBucketConfiguration,
+            @Named(NODE_GC_CONTROLLER) ControllerConfiguration controllerConfiguration,
+            AgentManagementService agentManagementService,
+            KubeApiFacade kubeApiFacade,
+            KubeControllerConfiguration kubeControllerConfiguration
+    ) {
+        super(
+                NODE_GC_CONTROLLER,
+                NODE_GC_CONTROLLER_DESCRIPTION,
+                titusRuntime,
+                scheduler,
+                tokenBucketConfiguration,
+                controllerConfiguration
+        );
+        this.agentManagementService = agentManagementService;
+        this.kubeApiFacade = kubeApiFacade;
+        this.clock = titusRuntime.getClock();
+        this.kubeControllerConfiguration = kubeControllerConfiguration;
+    }
+
+    @Override
+    public boolean shouldGc() {
+        return kubeApiFacade.getNodeInformer().hasSynced();
+    }
+
+    @Override
+    public List<V1Node> getItemsToGc() {
+        return kubeApiFacade.getNodeInformer().getIndexer().list().stream()
+                .filter(this::isNodeInConfiguredAccount)
+                .filter(this::isNodeEligibleForGc)
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    public boolean gcItem(V1Node item) {
+        return GcControllerUtil.deleteNode(kubeApiFacade, logger, item);
+    }
+
+    @VisibleForTesting
+    boolean isNodeInConfiguredAccount(V1Node node) {
+        V1ObjectMeta metadata = node.getMetadata();
+        if (metadata == null) {
+            return false;
+        }
+
+        Map<String, String> annotations = metadata.getAnnotations();
+        if (annotations == null || annotations.isEmpty()) {
+            return false;
+        }
+
+        String accountId = annotations.getOrDefault(NODE_LABEL_ACCOUNT_ID, "");
+        return accountId.equals(kubeControllerConfiguration.getAccountId());
+    }
+
+    @VisibleForTesting
+    boolean isNodeEligibleForGc(V1Node node) {
+        String nodeName = KubeUtil.getMetadataName(node.getMetadata());
+        Optional<V1NodeCondition> readyNodeConditionOpt = KubeUtil.findNodeCondition(node, READY);
+        if (StringExt.isNotEmpty(nodeName) && readyNodeConditionOpt.isPresent()) {
+            V1NodeCondition readyNodeCondition = readyNodeConditionOpt.get();
+            boolean isReadyConditionTimestampPastGracePeriod = hasConditionGracePeriodElapsed(readyNodeCondition,
+                    kubeControllerConfiguration.getNodeGcGracePeriodMs());
+            return isReadyConditionTimestampPastGracePeriod && isAgentInstanceNotAvailable(nodeName);
+        }
+        return false;
+    }
+
+    private boolean hasConditionGracePeriodElapsed(V1NodeCondition condition, long gracePeriodMs) {
+        DateTime lastHeartbeatTime = condition.getLastHeartbeatTime();
+        return lastHeartbeatTime != null && clock.isPast(lastHeartbeatTime.getMillis() + gracePeriodMs);
+    }
+
+    private boolean isAgentInstanceNotAvailable(String nodeName) {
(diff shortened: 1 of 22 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -0,0 +1,180 @@
...
+        scheduler.schedule(reconcileSchedulerDescriptor, e -> gcNodes(), ExecutorsExt.namedSingleThreadExecutor("node-gc-controller"));
+    }
+
+    void gcNodes() {
+        if (!kubeApiFacade.getNodeInformer().hasSynced()) {
+            logger.info("Node informer has not synced");
+            return;
+        }
+
+        if (!configuration.isNodeGcControllerEnabled()) {
```
joshi-keyur, line 105: Move this first in the function for better flow?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -0,0 +1,180 @@
...
+            gcNode(nodeName);
+        }
+        logger.info("Finished node GC");
+    }
+
+    @VisibleForTesting
+    boolean isNodeEligibleForGc(V1Node node) {
+        String nodeName = KubeUtil.getMetadataName(node.getMetadata());
+        Optional<V1NodeCondition> readyNodeConditionOpt = KubeUtil.findNodeCondition(node, READY);
+        if (StringExt.isNotEmpty(nodeName) && readyNodeConditionOpt.isPresent()) {
```
joshi-keyur, line 129: I am wondering how likely is the scenario where nodeName is missing/empty and if that is indicative of a different issue.

corindwyer: This should not happen but the models/API don't provide any guarantees so there is a lot of defensive checks.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -0,0 +1,180 @@
...
+            V1NodeCondition readyNodeCondition = readyNodeConditionOpt.get();
+            boolean isReadyConditionTimestampPastGracePeriod = hasConditionGracePeriodElapsed(readyNodeCondition, configuration.getNodeGcGracePeriodMs());
+            boolean isAgentInstanceNotAvailable = isAgentInstanceNotAvailable(nodeName);
+            return isReadyConditionTimestampPastGracePeriod && isAgentInstanceNotAvailable;
+        }
+        return false;
+    }
+
+    private boolean hasConditionGracePeriodElapsed(V1NodeCondition condition, long gracePeriodMs) {
```
joshi-keyur, line 138: What is the heartbeat rate for the Ready condition? Should we document how the gracePeriodMs relates to the default heartbeat rate and provide a guardrail check so we don't accidentally set the gracePeriod lower?

corindwyer: The heartbeat rate is determined by the kubelet. This check is only used as an extra safety check, but the real source is the cloud provider information from agent management.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PodGcController.java
```diff
@@ -0,0 +1,246 @@
...
+
+        if (!configuration.isPodGcControllerEnabled()) {
+            logger.info("Pod GC is disabled");
+            return;
+        }
+
+        List<V1Node> nodes = kubeApiFacade.getNodeInformer().getIndexer().list();
+        List<V1Pod> pods = kubeApiFacade.getPodInformer().getIndexer().list();
+
+        gcPodsOnUnknownNode(nodes, pods);
```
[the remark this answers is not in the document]
corindwyer: The idea is that the code that can fail in each one does its own exception handling but I think it is fine to be safer here.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -0,0 +1,180 @@
...
+    @Activator
+    public void enterActiveMode() {
+        ScheduleDescriptor reconcileSchedulerDescriptor = ScheduleDescriptor.newBuilder()
+                .withName("nodeGcController")
+                .withDescription("GC nodes that no longer exist")
+                .withInitialDelay(Duration.ofMillis(configuration.getNodeGcControllerInitialDelayMs()))
+                .withInterval(Duration.ofMillis(configuration.getNodeGcControllerIntervalMs()))
+                .withTimeout(Duration.ofMillis(configuration.getNodeGcControllerTimeoutMs()))
+                .build();
+        scheduler.schedule(reconcileSchedulerDescriptor, e -> gcNodes(), ExecutorsExt.namedSingleThreadExecutor("node-gc-controller"));
```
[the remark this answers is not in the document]
fabiokung: Same for the `ExecutorService`

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -653,6 +620,49 @@ private void podUpdated(V1Pod pod) {
         }
     }
 
+    private void podDeleted(V1Pod pod) {
+        try {
+            V1ObjectMeta metadata = pod.getMetadata();
+            V1PodStatus status = pod.getStatus();
+            String podName = KubeUtil.getMetadataName(metadata);
+
+            if (!isPodValid(pod, metadata, status, podName)) {
```
fabiokung, line 629: if a Pod is deleted (by anyone) and it didn't have a proper status field set yet, we wait for full reconciliation to catch it?

Shouldn't we consider emitting a `ContainerEvent` with `REASON_TASK_LOST` (or `REASON_UNKNOWN`) here, for that case?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @VisibleForTesting
+    boolean isNodeEligibleForGc(V1Node node) {
+        String nodeName = KubeUtil.getMetadataName(node.getMetadata());
+        Optional<V1NodeCondition> readyNodeConditionOpt = KubeUtil.findNodeCondition(node, READY);
+        if (StringExt.isNotEmpty(nodeName) && readyNodeConditionOpt.isPresent()) {
+            V1NodeCondition readyNodeCondition = readyNodeConditionOpt.get();
+            boolean isReadyConditionTimestampPastGracePeriod = hasConditionGracePeriodElapsed(readyNodeCondition, configuration.getNodeGcGracePeriodMs());
+            boolean isAgentInstanceNotAvailable = isAgentInstanceNotAvailable(nodeName);
+            return isReadyConditionTimestampPastGracePeriod && isAgentInstanceNotAvailable;
```
fabiokung, line 133: very minor optimization here: inline the second variable and prevent some wasted cycles when the first check already fails. The method call is already very descriptive and readability isn't much improved by the extra variable.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -0,0 +1,180 @@
...
+        if (!agentInstanceOpt.isPresent()) {
+            return true;
+        }
+
+        AgentInstance agentInstance = agentInstanceOpt.get();
+        return agentInstance.getLifecycleStatus().getState() == InstanceLifecycleState.Stopped;
+    }
+
+    private void gcNode(String nodeName) {
+        if (StringExt.isNotEmpty(nodeName)) {
```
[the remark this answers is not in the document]
fabiokung: minor: consider inverting the boolean check and do an early return to reduce indentation for the rest of the method

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -0,0 +1,180 @@
...
+                        null
+                );
+            } catch (JsonSyntaxException e) {
+                // this is probably successful. the generated client has the wrong response type
+            } catch (ApiException e) {
+                if (!e.getMessage().equalsIgnoreCase(NOT_FOUND)) {
+                    logger.error("Failed to delete node: {} with error: ", nodeName, e);
+                }
+            } catch (Exception e) {
+                logger.error("Failed to delete node: {} with error: ", nodeName, e);
```
fabiokung, line 176: `CodeInvariants#unexpectedError()`?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -0,0 +1,180 @@
...
+        Set<String> nodeNamesToGc = nodes.stream()
+                .filter(this::isNodeEligibleForGc)
+                .map(n -> KubeUtil.getMetadataName(n.getMetadata()))
+                .filter(StringExt::isNotEmpty)
+                .collect(Collectors.toSet());
+
+        logger.info("Attempting to GC {} nodes: {}", nodeNamesToGc.size(), nodeNamesToGc);
+        this.nodesToGc.set(nodeNamesToGc.size());
+        for (String nodeName : nodeNamesToGc) {
+            gcNode(nodeName);
```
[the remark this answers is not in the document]
fabiokung: the periodic scheduler has a timeout, but that doesn't guarantee any progress is made on each iteration. I agree a limit here is probably good

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
```diff
@@ -167,13 +205,15 @@ private void reconcile() {
         List<Task> tasks = v3JobOperations.getTasks();
 
         Map<String, V1Node> nodesById = nodes.stream().collect(Collectors.toMap(
-                node -> node.getMetadata().getName(),
+                node -> KubeUtil.getMetadataName(node.getMetadata()),
```
fabiokung, line 208: should probably filter out entries with `null` (or empty strings) as their name since this is an index

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/DefaultKubeJobManagementReconciler.java
```diff
@@ -167,13 +205,15 @@ private void reconcile() {
...
                 Function.identity()
         ));
         Map<String, Task> currentTasks = tasks.stream().collect(Collectors.toMap(Task::getId, Function.identity()));
-        Set<String> currentPodNames = pods.stream().map(p -> p.getMetadata().getName()).collect(Collectors.toSet());
+        Set<String> currentPodNames = pods.stream().map(p -> KubeUtil.getMetadataName(p.getMetadata())).collect(Collectors.toSet());
```
fabiokung, line 212: same here, filter out the empty string to void trying to call delete with an empty name later

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PodGcController.java
```diff
@@ -0,0 +1,246 @@
...
+     * GC pods in Pending phase with a deletion timestamp.
+     */
+    private void gcPendingPodsWithDeletionTimestamp(List<V1Pod> pods) {
+        List<V1Pod> pendingPodsWithDeletionTimestampToGc = pods.stream()
+                .filter(this::isPodInPendingPhaseWithDeletionTimestamp)
+                .collect(Collectors.toList());
+
+        logger.info("Attempting to GC {} pending pods: {} with deletion timestamp", pendingPodsWithDeletionTimestampToGc.size(),
+                pendingPodsWithDeletionTimestampToGc);
+        pendingPodsWithDeletionTimestampToGcGauge.set(pendingPodsWithDeletionTimestampToGc.size());
```
joshi-keyur, line 200: This gauge could be set to an inflated value depending on how many pods are actually GC'd successfully vs how many fail in the gcPod() function (where the exception is swallowed). Do you think it'd be better to wait until the gcPod loop to set the gauge to the correctly number or possibly introduce a separate gauge to track attempted vs actual pod GC?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PodGcController.java
```diff
@@ -0,0 +1,246 @@
...
+                        DEFAULT_NAMESPACE,
+                        null,
+                        null,
+                        0,
+                        null,
+                        BACKGROUND,
+                        null
+                );
+            } catch (JsonSyntaxException e) {
+                // this is probably successful. the generated client has the wrong response type
```
joshi-keyur, line 236: Do we know how frequently and when this case could happen? Does it make sense to log at DEBUG level in case we need to troubleshoot a GC issue?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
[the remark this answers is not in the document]
corindwyer: it is already checked earlier so just going to remove it

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PodGcController.java
corindwyer: it happens 100% of the time so I don't think we need a log line

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
corindwyer: I wouldn't say this is unexpected as there could be networking/client errors that happen.

fabiokung (approved):

tbak (approved):

joshi-keyur (approved):

PR #892, opened 2020-08-05
core task log locations populated from grpcTask

amit-git:
### Updates log location attributes for core Task type

RemoteJobManagementClient is used by JobDataReplicator to build a local cache of running jobs and tasks. It depends on converting a grpc Task object into a core Task object. In that process, it seems to be losing information about task log location that is provided by the grpc Task. This change is needed to enable client cache layers using RemoteJobManagementClient to access task log information including ui-link, s3 bucket / key, s3 account Id etc.

### DIFF
--- a/titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
+++ b/titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
@@ -518,7 +518,7 @@ public final class GrpcJobManagementModelConverters {
         return builder.build();
     }
 
-    static Map<String, String> buildAttributesMapForCoreTask(com.netflix.titus.grpc.protogen.Task grpcTask) {
+    private static Map<String, String> buildAttributesMapForCoreTask(com.netflix.titus.grpc.protogen.Task grpcTask) {
         Map<String, String> attributes = new HashMap<>(grpcTask.getAttributesMap());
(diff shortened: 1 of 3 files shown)


titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -512,6 +518,27 @@ public static Task toCoreTask(com.netflix.titus.grpc.protogen.Task grpcTask) {
         return builder.build();
     }
 
+    static Map<String, String> buildAttributesMapForCoreTask(com.netflix.titus.grpc.protogen.Task grpcTask) {
```
andrew-leung, line 521: private?

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -512,6 +518,27 @@ public static Task toCoreTask(com.netflix.titus.grpc.protogen.Task grpcTask) {
...
+        Map<String, String> attributes = new HashMap<>(grpcTask.getAttributesMap());
+        if (grpcTask.hasLogLocation()) {
+            LogLocation logLocation = grpcTask.getLogLocation();
+            if (logLocation != null) {
```
tbak, line 525: Redundant check as we already call `grpcTask.hasLogLocation()` which checks for that.

PR #894, opened 2020-08-06
Configurable prediction quantiles

tbak:


rostyboost:
Thank you @tbak !

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobRuntimePredictionUtil.java
```diff
@@ -16,21 +16,31 @@
...
-        return low.getConfidence() == LOW_QUANTILE && high.getConfidence() == HIGH_QUANTILE;
+    static boolean expectedLowest(JobRuntimePrediction low) {
+        return low.getConfidence() == LOW_QUANTILE;
+    }
+
+    static Optional<JobRuntimePrediction> findRequested(SortedSet<JobRuntimePrediction> predictions, double quantile) {
+        if (predictions.isEmpty()) {
+            return Optional.empty();
+        }
+        return predictions.stream().filter(p -> p.getConfidence() == quantile).findFirst();
```
fabiokung, line 43: minor: I'm always a bit worried about comparing floating point numbers with `==`. Might be good to compare against a tiny `epsilon` range here to make it less brittle to floating point math in the future

tbak: I agree in general. In this case however exact match is what we want, as we compare two configured values.

PR #897, opened 2020-08-13
Upgrade SpringBoot to 2.3

tbak:


build.gradle
```diff
@@ -61,8 +61,9 @@ subprojects {
 
         awsSdkVersion = '1.11.+'
         javaxElVersion = '3.+'
-        springVersion = '5.2.+'
-        springBootVersion = '2.2.5.RELEASE'
+        springVersion = '5.2.8.RELEASE'
+        springSecurityVersion = '5.2.5.RELEASE'
```
andrew-leung, line 65: Looks like internally we're using 5.3 for spring-security-web. Any reason to stay behind?

tbak: Good catch. I will fix that.

PR #898, opened 2020-08-18
Set the default DNS policy for k8s pods to 'Default'

solarkennedy:
Per the docs:
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
the default policy is indeed not "Default", but is "ClusterFirst".

We don't configure cluster ips in titus, nor do we want
k8s to manipulate our DNS configuration so setting the policy to
the actual "Default" is what we want.

This is what we are doing anyway, this just removes an annoying warning:
```
kubelet does not have ClusterDNS IP configured and cannot create Pod using "ClusterFirst" policy. Falling back to "Default" policy.
```

----

This is my first titus api contribution. I could use some guidance on verifying this works.


solarkennedy:
> LGTM. We don't need the same on the VK path, right?

Nope (pretty sure). On the vk path the DNS is being set by the CNI because it can configure it via docker directly.

PR #903, opened 2020-08-26
RFC - Make the kubeapiserver connector understand pod ips from kubelet

solarkennedy:
I kinda don't know what I'm doing here.

It looks like vk + executor add extra annotations after the fact ([here](https://github.com/Netflix-Skunkworks/titus-virtual-kubelet/blob/68ea9b6c6183b91eaa12d2d2505f18732d81aae6/provider/runtime_pod.go#L188)) to provide networking metadata back to the control plane.

This doesn't work on kubelet at all, so the control plane has no idea what ip address a kubelet pod has.

I don't know how easy it is going to be to add these annotations via kubelet, but this unblocks things a little bit by getting at least an ip address?

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
@@ -52,6 +52,7 @@ import com.netflix.titus.common.util.StringExt;
 import com.netflix.titus.common.util.tuple.Pair;
 import com.netflix.titus.master.jobmanager.service.event.JobManagerReconcilerEvent;
 import com.netflix.titus.master.mesos.TitusExecutorDetails;
+import com.netflix.titus.master.mesos.kubeapiserver.direct.model.PodWrapper;
 import com.netflix.titus.master.service.management.ApplicationSlaManagementService;
 import org.apache.mesos.Protos;
 
@@ -124,12 +125,12 @@ public final class JobManagerUtil {
             }
 
             Task newTask = JobFunctions.changeTaskStatus(oldTask, newTaskStatus);
-            Task newTaskWithPlacementData = attachPlacementData(newTask, detailsOpt);
+            Task newTaskWithPlacementData = attachTitusExecutorData(newTask, detailsOpt);
             return Optional.of(newTaskWithPlacementData);
         };
     }
 
-    public static Task attachPlacementData(Task task, Optional<TitusExecutorDetails> detailsOpt) {
+    public static Task attachTitusExecutorData(Task task, Optional<TitusExecutorDetails> detailsOpt) {
         return detailsOpt.map(details -> {
             TitusExecutorDetails.NetworkConfiguration networkConfiguration = details.getNetworkConfiguration();
             if (networkConfiguration != null) {
@@ -146,6 +147,22 @@ public final class JobManagerUtil {
             }
             return task;
         }).orElse(task);
+    }
+
+    public static Task attachKubeletData(Task task, PodWrapper podWrapper) {
+        if (podWrapper.getV1Pod().getStatus() == null) {
+            return task;
+        }
+        String ipv4 = podWrapper.getV1Pod().getStatus().getPodIP();
+        if (ipv4 == null || ipv4.isEmpty()) {
+            return task;
+        }
+
+        Map<String, String> newContext = new HashMap<>(task.getTaskContext());
+        BiConsumer<String, String> contextSetter = (key, value) -> StringExt.applyIfNonEmpty(value, v -> newContext.put(key, v));
+        contextSetter.accept(TaskAttributes.TASK_ATTRIBUTES_CONTAINER_IP, ipv4);
+        contextSetter.accept(TaskAttributes.TASK_ATTRIBUTES_CONTAINER_IPV4, ipv4);
+        return task.toBuilder().addAllToTaskContext(newContext).build();
     }
 
     public static Optional<String> parseEniResourceId(String resourceId) {
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
@@ -290,8 +290,13 @@ public class KubeNotificationProcessor {
                 .build();
 
         Task fixedTask = fillInMissingStates(podWrapper, updatedTask);
-        Task taskWithPlacementData = JobManagerUtil.attachPlacementData(fixedTask, executorDetailsOpt);
-        Task taskWithNodeMetadata = node.map(n -> attachNodeMetadata(taskWithPlacementData, n)).orElse(taskWithPlacementData);
+        Task taskWithExecutorData;
+        if (executorDetailsOpt.isPresent()) {
+            taskWithExecutorData = JobManagerUtil.attachTitusExecutorData(fixedTask, executorDetailsOpt);
+        } else {
+            taskWithExecutorData = JobManagerUtil.attachKubeletData(fixedTask, podWrapper);
+        }
+        Task taskWithNodeMetadata = node.map(n -> attachNodeMetadata(taskWithExecutorData, n)).orElse(taskWithExecutorData);
         Task taskWithAnnotations = addMissingAttributes(podWrapper, taskWithNodeMetadata);
 
         return taskWithAnnotations;
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
@@ -132,6 +132,9 @@ public class KubeUtil {
     }
 
     public static Optional<V1ContainerState> findContainerState(V1Pod pod) {
+        if (pod.getStatus() == null) {
+            return Optional.empty();
+        }
         List<V1ContainerStatus> containerStatuses = pod.getStatus().getContainerStatuses();
         if (containerStatuses != null) {
             for (V1ContainerStatus status : containerStatuses) {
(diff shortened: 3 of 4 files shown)


solarkennedy:
@tbak said

```
In KubeNotificationProcessor

In line 293:
Task taskWithPlacementData = JobManagerUtil.attachPlacementData(fixedTask, executorDetailsOpt);
we would have to do what you already done in your PR.


I would check if executorDetailsOpt is empty, and if it is set the task context directly based on pod data.
```

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
```diff
@@ -148,6 +149,15 @@ public static Task attachPlacementData(Task task, Optional<TitusExecutorDetails>
         }).orElse(task);
     }
 
+    public static Task attachKubeletData(Task task, PodWrapper podWrapper) {
+        String ipv4 = podWrapper.getV1Pod().getStatus().getPodIP();
+        Map<String, String> newContext = new HashMap<>(task.getTaskContext());
+        BiConsumer<String, String> contextSetter = (key, value) -> StringExt.applyIfNonEmpty(value, v -> newContext.put(key, v));
```
tbak, line 155: You can use equivalent method from Evaluators class in titus-common.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
```diff
@@ -148,6 +149,15 @@ public static Task attachPlacementData(Task task, Optional<TitusExecutorDetails>
         }).orElse(task);
     }
 
+    public static Task attachKubeletData(Task task, PodWrapper podWrapper) {
+        String ipv4 = podWrapper.getV1Pod().getStatus().getPodIP();
```
tbak, line 153: You should check defensively that status is not null as well.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
```diff
@@ -290,8 +290,13 @@ static Task updateTaskStatus(PodWrapper podWrapper,
                 .build();
 
         Task fixedTask = fillInMissingStates(podWrapper, updatedTask);
-        Task taskWithPlacementData = JobManagerUtil.attachPlacementData(fixedTask, executorDetailsOpt);
-        Task taskWithNodeMetadata = node.map(n -> attachNodeMetadata(taskWithPlacementData, n)).orElse(taskWithPlacementData);
+        Task taskWithExecutorData;
+        if (executorDetailsOpt == null) {
```
tbak, line 294: `executorDetailsOpt == null || executorDetailsOpt.isEmpty()`

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
```diff
@@ -127,7 +127,21 @@ public static boolean isPodPhaseTerminal(String phase) {
                     )
             );
             return Optional.of(titusExecutorDetails);
+        } else if (!Strings.isNullOrEmpty(pod.getStatus().getPodIP())) {
```
tbak, line 130: That is not need needed?

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
solarkennedy: I see `applyNotNull` in Evaulators, but that won't help me against Empty strings though.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
solarkennedy: Good idea

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
solarkennedy: There is no method like that on this object (because it is optional?)

There is a `isPresent`, but it just checks for nullness too.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
tbak: We should also add `if(podWrapper.getV1Pod().getStatus() == null) { return task; }`

solarkennedy:
Ok. Looks like `updateTaskStatus` also doesn't guard against that, so I can
add the same defense there.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
```diff
@@ -148,6 +149,20 @@ public static Task attachPlacementData(Task task, Optional<TitusExecutorDetails>
         }).orElse(task);
     }
 
+    public static Task attachKubeletData(Task task, PodWrapper podWrapper) {
+        if(podWrapper.getV1Pod().getStatus() == null) { return task; }
```
fabiokung, line 153: minor: formatting is odd here. There's code style config file for IntelliJ you can use in the `codequality/` folder

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
```diff
@@ -148,6 +149,20 @@ public static Task attachPlacementData(Task task, Optional<TitusExecutorDetails>
...
+        String ipv4 = podWrapper.getV1Pod().getStatus().getPodIP();
+        if (ipv4 == null || ipv4.isEmpty()) {
+            return task;
+        } else {
```
fabiokung, line 157: minor: remove the `else` (previous `if` clause already has an early `return`) and reduce indentation level

titus-server-master/src/test/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessorTest.java
```diff
@@ -45,16 +45,7 @@
...
-import io.kubernetes.client.openapi.models.V1ContainerStateRunning;
-import io.kubernetes.client.openapi.models.V1ContainerStateWaiting;
-import io.kubernetes.client.openapi.models.V1ContainerStatus;
-import io.kubernetes.client.openapi.models.V1Node;
-import io.kubernetes.client.openapi.models.V1NodeAddress;
-import io.kubernetes.client.openapi.models.V1NodeStatus;
-import io.kubernetes.client.openapi.models.V1ObjectMeta;
-import io.kubernetes.client.openapi.models.V1Pod;
-import io.kubernetes.client.openapi.models.V1PodStatus;
+import io.kubernetes.client.openapi.models.*;
```
fabiokung, line 48: odd formatting, please use the code-style rules in `codequality/`

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
```diff
@@ -290,8 +290,13 @@ static Task updateTaskStatus(PodWrapper podWrapper,
                 .build();
 
         Task fixedTask = fillInMissingStates(podWrapper, updatedTask);
-        Task taskWithPlacementData = JobManagerUtil.attachPlacementData(fixedTask, executorDetailsOpt);
-        Task taskWithNodeMetadata = node.map(n -> attachNodeMetadata(taskWithPlacementData, n)).orElse(taskWithPlacementData);
+        Task taskWithExecutorData;
+        if (executorDetailsOpt.isPresent()) {
```
fabiokung, line 294: instead of doing this here, did you consider changing the logic in `KubeUtil.getTitusExecutorDetails(pod)`, and checking whether `pod.getStatus().getPodIP()` should be used instead of annotations there? Centralizing this logic in that method looks cleaner to me. As more metadata is added to the kubelet path, I think it will be easier to keep both code paths more consistent in there.

titus-server-master/src/test/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessorTest.java
solarkennedy: Ah! I should have read CONTRIBUTING!

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
solarkennedy: I feel like I'm getting conflicting review here. I originally had this logic in `getTitusExecutorDetails`, but @tbak asked me to check if executorDetailsOpt was present first and do it this way (pretty sure if I'm interpreting his intent).

tbak: My point was that `TitusExecutorDetails` is a legacy data structure that we want to get rid off.
I think it would be better to move the new function into `KubeUtil`, but keep it separate as it is now.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeUtil.java
```diff
@@ -132,6 +132,7 @@ public static boolean isPodPhaseTerminal(String phase) {
     }
 
     public static Optional<V1ContainerState> findContainerState(V1Pod pod) {
+        if(pod.getStatus() == null) { return Optional.empty(); }
```
fabiokung, line 135: minor: missed formatting here

PR #904, opened 2020-08-27
Feature/resource pool assignment

tbak:
Resource pools replace tiers. When a KubeScheduler pod is created for a task, a resource pool affinity rules must be associated with it.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeModule.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeModule.java
@@ -16,6 +16,7 @@
 
 package com.netflix.titus.master.mesos.kubeapiserver;
 
+import java.util.Arrays;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
@@ -24,15 +25,13 @@ import com.google.inject.Injector;
 import com.google.inject.Provides;
 import com.google.inject.name.Names;
 import com.netflix.archaius.ConfigProxyFactory;
+import com.netflix.archaius.api.Config;
 import com.netflix.titus.api.FeatureActivationConfiguration;
 import com.netflix.titus.common.runtime.TitusRuntime;
 import com.netflix.titus.common.util.limiter.tokenbucket.FixedIntervalTokenBucketConfiguration;
 import com.netflix.titus.master.mesos.MesosConfiguration;
 import com.netflix.titus.master.mesos.VirtualMachineMasterService;
 import com.netflix.titus.master.mesos.kubeapiserver.client.JobControllerKubeApiFacade;
-import com.netflix.titus.runtime.connector.kubernetes.KubeApiClients;
-import com.netflix.titus.runtime.connector.kubernetes.KubeApiFacade;
-import com.netflix.titus.runtime.connector.kubernetes.NoOpKubeApiFacade;
 import com.netflix.titus.master.mesos.kubeapiserver.direct.DefaultDirectKubeApiServerIntegrator;
 import com.netflix.titus.master.mesos.kubeapiserver.direct.DefaultPodAffinityFactory;
 import com.netflix.titus.master.mesos.kubeapiserver.direct.DefaultTaskToPodConverter;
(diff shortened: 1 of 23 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/PodAffinityFactory.java
```diff
@@ -26,5 +29,5 @@
  */
 public interface PodAffinityFactory {
 
-    V1Affinity buildV1Affinity(Job<?> job, Task task);
+    Pair<V1Affinity, Map<String, String>> buildV1Affinity(Job<?> job, Task task);
```
fabiokung, line 32: please document what the `Map` return should contain

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/resourcepool/CapacityGroupPodResourcePoolResolver.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public CapacityGroupPodResourcePoolResolver(DirectKubeConfiguration configuration,
+                                                Config config,
+                                                ApplicationSlaManagementService capacityGroupService,
+                                                TitusRuntime titusRuntime) {
+        this.configuration = configuration;
+        this.config = config;
+        this.capacityGroupService = capacityGroupService;
+        this.titusRuntime = titusRuntime;
+        refreshResourcePoolToCapacityGroupMappers();
```
fabiokung, line 60: minor: thoughts on deferring this to after leader activation?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeModule.java
```diff
@@ -77,6 +87,24 @@ public FixedIntervalTokenBucketConfiguration getGcUnknownPodsTokenBucketConfigur
         return factory.newProxy(FixedIntervalTokenBucketConfiguration.class, "titusMaster.kube.gcUnknownPodsTokenBucket");
     }
 
+    @Provides
+    @Singleton
+    public PodResourcePoolResolver getPodResourcePoolResolver(DirectKubeConfiguration configuration,
+                                                              Config config,
+                                                              ApplicationSlaManagementService capacityGroupService,
+                                                              TitusRuntime titusRuntime) {
+        return new PodResourcePoolResolverChain(Arrays.asList(
```
fabiokung, line 96: minor: consider adding a `GpuResourcePoolResolver` into the chain to automatically assign jobs requesting GPUs to `RESOURCE_POOL_GPU_PREFIX` pools

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/taint/DefaultTaintTolerationFactory.java
```diff
@@ -56,23 +49,13 @@ public DefaultTaintTolerationFactory(DirectKubeConfiguration configuration,
         tolerations.add(Tolerations.TOLERATION_VIRTUAL_KUBLET);
         tolerations.add(Tolerations.TOLERATION_KUBE_SCHEDULER);
 
-        tolerations.add(resolveTierToleration(job));
```
fabiokung, line 59: is this backwards compatible? I think we may need to keep double writing for now until all other components are aligned

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeModule.java
tbak: I have intentionally left GPU out of the picture for now. The way we handle them is complicated, and I want to address it in a separate PR.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/resourcepool/CapacityGroupPodResourcePoolResolver.java
tbak: It has a very minor CPU impact and does not make any external calls. I would prefer to keep it as it is.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/taint/DefaultTaintTolerationFactory.java
tbak: I think you are right. It would be too risky to assume we can make all updates in one shot.

tbak:
@fabiokung `resource-pool` is a label associated with a node. We use node label affinity rules to make the desired selection. If you would create a pod without any resource pool affinity rule, it could run anywhere. Nodes that we do not want to be available to everybody should use taints to prevent that. One example could be GPU instances. Or in `reserved` resource pool, we will enforce limits directly in the scheduler, but will allow to run any container in the leftover capacity or opportunistic resources.

PR #905, opened 2020-08-27
Rate limit reconciliation of tasks stuck in state

fabiokung:
... with dynamic configuration.

This will help prevent bad feedback loops when task state gets stale (e.g.: pod informers are not receiving updates), or when downstream systems are slow, adding some back-pressure to reconciliation.

### DIFF
--- a/titus-common/src/main/java/com/netflix/titus/common/util/limiter/Limiters.java
+++ b/titus-common/src/main/java/com/netflix/titus/common/util/limiter/Limiters.java
@@ -37,15 +37,20 @@ public class Limiters {
     }
 
     /**
+     * Useful for testing.
+     */
+    public static TokenBucket unlimited(String name) {
+        return new Unlimited(name);
+    }
+
+    /**
      * Create a {@link TokenBucket} with a fixed interval {@link RefillStrategy}.
      */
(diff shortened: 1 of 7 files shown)


titus-common/src/main/java/com/netflix/titus/common/util/limiter/Limiters.java
```diff
@@ -75,4 +80,54 @@ public static TokenBucket createInstrumentedFixedIntervalTokenBucket(String name
...
+        public void take(long numberOfTokens) {
+        }
+
+        @Override
+        public void refill(long numberOfToken) {
+        }
+
+        @Override
+        public RefillStrategy getRefillStrategy() {
+            return null;
```
tbak, line 130: We should return an actual strategy here.

tbak:
LGTM except the single comment I have made.

titus-common/src/main/java/com/netflix/titus/common/util/limiter/Limiters.java
fabiokung: Yeah, thanks for keeping me honest. I was being lazy here. I will add it

PR #906, opened 2020-08-29
Added a shared dev script to deploy to netflix dev stacks

solarkennedy:
I hear everyone has their own flavor of this.

Instead of documenting, I figured we could just document in terms of code?


solarkennedy:
I've added this to our internal nflx repo instead because it is so nflxy.

PR #907, opened 2020-09-01
Check agent instance relocation attributes in placement

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/scheduler/constraint/AgentManagementConstraint.java
```diff
@@ -50,6 +50,12 @@
 @Singleton
 public class AgentManagementConstraint implements SystemConstraint {
 
+    /**
+     * See RelocationAttributes for full documentation.
+     */
+    public static final String RELOCATION_REQUIRED = "titus.relocation.required";
+    public static final String RELOCATION_REQUIRED_IMMEDIATELY = "titus.relocation.requiredImmediately";
```
fabiokung, line 57: thoughts on extracting these constants (from `RelocationAttributes`) to a common module and centralizing there?

corindwyer: That was the original intent, but it is fine to be extra safe here while we transition to taints.

PR #913, opened 2020-09-03
Add S3 writer role support

tbak:
With this change a decision about which IAM role to use for S3 uploads is made by the control plane, and can be easily changed in the future without requiring agent side updates.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
@@ -161,4 +161,9 @@ public interface DirectKubeConfiguration extends KubeConnectorConfiguration {
      */
     @DefaultValue("false")
     boolean isResourcePoolAffinityEnabled();
+
+    /**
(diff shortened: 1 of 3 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
```diff
@@ -161,4 +161,9 @@
      */
     @DefaultValue("false")
     boolean isResourcePoolAffinityEnabled();
+
+    /**
+     * Default IAM role that should be used to upload log files to S3 bucket.
+     */
+    String getDefaultS3WriterRole();
```
jigish, line 168: how does this get populated?

jigish (approved):

corindwyer (approved):

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
tbak: Fast property

PR #915, opened 2020-09-09
Add ClusterMemberVerifier API to provide extra check for cluster membership

tbak:


titus-common-client/src/main/java/com/netflix/titus/client/clustermembership/resolver/ClusterMemberVerifier.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.netflix.titus.api.clustermembership.model.ClusterMember;
+
+/**
+ * {@link ClusterMemberVerifier} verifies that a discovered cluster member belongs to the desired membership group.
+ * This unlikely event may happen, if an IP address is reused by another stack of the same application before the
+ * the state is cleaned up by the client. The client when reconnecting would connect to a newly created instance of
+ * the same application but in a different stack.
+ */
+public interface ClusterMemberVerifier {
```
andrew-leung, line 27: Will the implementation be part of nftitus and metatron based?

andrew-leung (approved):

titus-common-client/src/main/java/com/netflix/titus/client/clustermembership/resolver/ClusterMemberVerifier.java
tbak: Not metatron based, but checking stack of the member that it matches that of the client. Of course this is just API so many implementations are possible.

PR #926, opened 2020-09-16
Add feature flag to enable/disable S3 writer configuration

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
```diff
@@ -162,6 +162,12 @@
     @DefaultValue("false")
     boolean isResourcePoolAffinityEnabled();
 
+    /**
+     * Set to true to enable resource pool affinity placement constraints.
```
corindwyer, line 166: javadoc looks wrong

corindwyer (approved):

fabiokung (approved):

PR #935, opened 2020-10-09
Set container environment in pod.env

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
```diff
@@ -150,6 +150,9 @@
     @DefaultValue("false")
     boolean isResourcePoolAffinityEnabled();
 
+    @DefaultValue("")
```
corindwyer, line 153: javadoc including expected format

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/env/TitusProvidedContainerEnvFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import com.netflix.titus.api.jobmanager.model.job.Task;
+
+public class TitusProvidedContainerEnvFactory implements ContainerEnvFactory {
+
+    private static final ContainerEnvFactory INSTANCE = new TitusProvidedContainerEnvFactory();
+
+    @Override
+    public Map<String, String> buildContainerEnv(Job<?> job, Task task) {
+        Map<String, String> env = new HashMap<>();
+        env.put("TITUS_JOB_ID", task.getJobId());
```
corindwyer, line 33: are there constants for these somewhere?

corindwyer (approved):

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/env/TitusProvidedContainerEnvFactory.java
tbak: No, this is a block of code that I moved from another place. I can create constants for it.

joshi-keyur (approved):

PR #941, opened 2020-10-15
update the max job dimensions

corindwyer:
### Description of the Change
* Update the max job dimensions.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobConfiguration.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobConfiguration.java
@@ -70,7 +70,7 @@ public interface JobConfiguration {
      * An upper bound on memory (megabytes) a single container may allocate. The actual limit may be lower, as it also depends
      * on instance types available in a tier.
      */
-    @DefaultValue("760000")
+    @DefaultValue("750000")
     int getMemoryMegabytesMax();
 
     @DefaultValue("10000")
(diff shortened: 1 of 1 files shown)


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobConfiguration.java
```diff
@@ -80,7 +80,7 @@
      * An upper bound on disk (megabytes) a single container may allocate. The actual limit may be lower, as it also depends
      * on instance types available in a tier.
      */
-    @DefaultValue("999000")
+    @DefaultValue("1550000")
```
rostyboost, line 83: (just a reminder that the Docker image size is counted in this budget, which is why we currently set the min at 10GB)

tbak (approved):

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobConfiguration.java
```diff
@@ -70,7 +70,7 @@
      * An upper bound on memory (megabytes) a single container may allocate. The actual limit may be lower, as it also depends
      * on instance types available in a tier.
      */
-    @DefaultValue("472000")
+    @DefaultValue("760000")
```
rostyboost, line 73: should we leave some slack for executor operations + spike during coredumps?

corindwyer: I left a couple of G but I could reduce it to be 750000 so there is > 10GB buffer.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobConfiguration.java
corindwyer: yes

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/sanitizer/JobConfiguration.java
rostyboost: Yes, `750000` might be better

rostyboost (approved):

andrew-leung (approved):
It may be worth some comments as bread crumbs for why the default values are what they are (e.g., "based on m5.metal advertised values") for the next person.

PR #950, opened 2020-11-10
Kubernetes Integration Improvements

corindwyer:
### Description of the Change
* Switch to the new node annotation prefix.
* Verify the res node annotation is in a valid format before creating a lease.
* Add the ability to turn off account id verification in the node gc controller.


tbak (approved):

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -498,6 +497,13 @@ private VirtualMachineLease nodeToLease(V1Node node) {
             return Collections.emptyList();
         }
 
+        // Validate the res value is correct otherwise reject the node until it is correct.
+        // TODO This code is temporary and will be changed as the node contract is changing.
+        String resValue = annotations.get(TITUS_NODE_DOMAIN + NODE_ATTRIBUTE_RES);
```
amit-git, line 502: a unit test could help explain the intention behind this change

amit-git (approved):

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
corindwyer: I agree but this class is being rewritten right now so I added a comment instead just to get something out quickly.

joshi-keyur (approved):

PR #952, opened 2020-11-11
Initial support for EBS volume model and validation

andrew-leung:
Partial feature support for EBS volumes. This PR depends on PR https://github.com/Netflix/titus-api-definitions/pull/127

This PR provides:

1. Internal EBS model/data types.
2. API support for EBS volumes specified as job attributes.
3. Validation of user provided EBS volumes.
4. Creation of Kube Persistent Volumes.
5. Pod configurations to use Persistent Volumes.

This PR does NOT provide to following which will be added in follow on PRs:
1. GC of previously created Persistent Volumes.
2. Integration testing with @solarkennedy 's specific CSI driver changes.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolume.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolume.java
@@ -1,0 +1,184 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.api.jobmanager.model.job.ebs;
+
+import java.util.Objects;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+import com.google.common.base.Preconditions;
+
+public class EbsVolume {
+
+    public EbsVolume(String volumeId, String volumeAvailabilityZone, int volumeCapacityGB, String mountPath, MountPerm mountPerm, String fsType) {
+        this.volumeId = volumeId;
+        this.volumeAvailabilityZone = volumeAvailabilityZone;
+        this.volumeCapacityGB = volumeCapacityGB;
+        this.mountPath = mountPath;
+        this.mountPermissions = mountPerm;
+        this.fsType = fsType;
+    }
+
+    public static Builder newBuilder() {
+        return new Builder();
+    }
+
+    public Builder toBuilder() {
+        return new Builder(this);
+    }
+
+    public enum MountPerm {RO, RW}
+
+    @NotNull
+    @Size(min = 1, max = 512, message = "EBS volume ID cannot be empty or greater than 512 bytes")
+    private final String volumeId;
+
+    @Size(min = 1, max = 512, message = "EBS volume AZ cannot be empty or greater than 512 bytes")
+    private final String volumeAvailabilityZone;
+
+    private final int volumeCapacityGB;
+
+    @NotNull
+    @Size(min = 1, max = 1024, message = "EBS volume mount path cannot be empty or greater than 1024 bytes")
+    private final String mountPath;
+
+    @NotNull(message = "'mountPermissions' is null")
+    private final MountPerm mountPermissions;
+
+    @NotNull
+    @Size(min = 1, max = 512, message = "EBS volume FS type cannot be empty or greater than 512 bytes")
+    private final String fsType;
+
+    public String getVolumeId() {
+        return volumeId;
+    }
+
+    public String getVolumeAvailabilityZone() {
+        return volumeAvailabilityZone;
+    }
+
+    public int getVolumeCapacityGB() {
+        return volumeCapacityGB;
+    }
+
+    public String getMountPath() {
+        return mountPath;
+    }
+
+    public MountPerm getMountPermissions() {
+        return mountPermissions;
+    }
+
+    public String getFsType() {
+        return fsType;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        EbsVolume ebsVolume = (EbsVolume) o;
+        return volumeCapacityGB == ebsVolume.volumeCapacityGB &&
+                volumeId.equals(ebsVolume.volumeId) &&
+                Objects.equals(volumeAvailabilityZone, ebsVolume.volumeAvailabilityZone) &&
+                mountPath.equals(ebsVolume.mountPath) &&
+                mountPermissions == ebsVolume.mountPermissions &&
+                fsType.equals(ebsVolume.fsType);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(volumeId, volumeAvailabilityZone, volumeCapacityGB, mountPath, mountPermissions, fsType);
+    }
+
+    @Override
+    public String toString() {
+        return "EbsVolume{" +
+                "volumeId='" + volumeId + '\'' +
+                ", volumeAvailabilityZone='" + volumeAvailabilityZone + '\'' +
+                ", volumeCapacityGB=" + volumeCapacityGB +
+                ", mountPath='" + mountPath + '\'' +
+                ", mountPermissions=" + mountPermissions +
+                ", fsType='" + fsType + '\'' +
+                '}';
+    }
+
+    public static final class Builder {
+        private String volumeId;
+        private String volumeAvailabilityZone;
+        private int volumeCapacityGB;
+        private String mountPath;
+        private MountPerm mountPermissions;
+        private String fsType;
+
+        private Builder() {
+        }
+
+        private Builder(EbsVolume ebsVolume) {
+            this.volumeId = ebsVolume.getVolumeId();
+            this.volumeAvailabilityZone = ebsVolume.getVolumeAvailabilityZone();
+            this.volumeCapacityGB = ebsVolume.getVolumeCapacityGB();
+            this.mountPath = ebsVolume.getMountPath();
+            this.mountPermissions = ebsVolume.getMountPermissions();
+            this.fsType = ebsVolume.getFsType();
+        }
+
+        public Builder withVolumeId(String val) {
+            volumeId = val;
+            return this;
+        }
+
+        public Builder withVolumeAvailabilityZone(String val) {
+            volumeAvailabilityZone = val;
+            return this;
+        }
+
+        public Builder withVolumeCapacityGB(int val) {
+            volumeCapacityGB = val;
+            return this;
+        }
+
+        public Builder withMountPath(String val) {
+            mountPath = val;
+            return this;
+        }
+
+        public Builder withMountPermissions(MountPerm val) {
+            mountPermissions = val;
+            return this;
+        }
+
+        public Builder withFsType(String val) {
+            fsType = val;
+            return this;
+        }
+
+        public EbsVolume build() {
+            Preconditions.checkNotNull(volumeId, "Volume ID is null");
+            Preconditions.checkNotNull(mountPath, "Mount path is null");
+            Preconditions.checkNotNull(mountPermissions, "Mount permission is null");
+            Preconditions.checkNotNull(fsType, "File system type is null");
+            // Volume AZ and capacity may be set after object creation during object sanitization
+            return new EbsVolume(volumeId, volumeAvailabilityZone, volumeCapacityGB, mountPath, mountPermissions, fsType);
+        }
+    }
+}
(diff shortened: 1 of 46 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultDirectKubeApiServerIntegrator.java
```diff
@@ -218,6 +208,58 @@ public String resolveReasonCode(Throwable cause) {
         return podCreateErrorToReasonCodeResolver.resolveReasonCode(cause);
     }
 
+    private static Mono<V1PersistentVolume> launchPersistentVolume(Job<?> job, Task task, V1Pod v1Pod, CoreV1Api coreV1Api) {
+        return Mono.defer(() -> {
+            Optional<V1Volume> optionalV1Volume = v1Pod.getSpec().getVolumes().stream()
```
tbak, line 213: Check getVolumes() != null? Kube objects are not fully initialized.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultDirectKubeApiServerIntegrator.java
```diff
@@ -218,6 +208,58 @@ public String resolveReasonCode(Throwable cause) {
         return podCreateErrorToReasonCodeResolver.resolveReasonCode(cause);
     }
 
+    private static Mono<V1PersistentVolume> launchPersistentVolume(Job<?> job, Task task, V1Pod v1Pod, CoreV1Api coreV1Api) {
+        return Mono.defer(() -> {
```
tbak, line 212: Add stopwatch and metrics like in pod below?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobEbsVolumeValidator.java
```diff
@@ -0,0 +1,70 @@
...
+    private final ValidatorMetrics metrics;
+
+    public JobEbsVolumeValidator(Supplier<ValidationError.Type> validationErrorTypeProvider,
+                                 TitusRuntime titusRuntime) {
+        this.validationErrorTypeProvider = validationErrorTypeProvider;
+        this.metrics = new ValidatorMetrics(JobEbsVolumeValidator.class.getSimpleName(), titusRuntime.getRegistry());
+    }
+
+    @Override
+    public Mono<Set<ValidationError>> validate(JobDescriptor jobDescriptor) {
```
tbak, line 50: This is non blocking operation which could be done in EntityValidator. Is this going to be extended to make an external call?

tbak (changes requested):

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolumeUtils.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+public class EbsVolumeUtils {
+
+    public static <E extends JobDescriptor.JobDescriptorExt> List<EbsVolume> getEbsVolumes(JobDescriptor<E> jobDescriptor) {
+        List<String> volumeIds = getVolumeIds(jobDescriptor);
+        Optional<String> mountPointOptional = getMountPoint(jobDescriptor);
+        Optional<EbsVolume.MountPerm> mountPermOptional = getEbsMountPerm(jobDescriptor);
+        Optional<String> fsTypeOptional = getEbsFsType(jobDescriptor);
+
+        if (!(mountPointOptional.isPresent() && mountPermOptional.isPresent() && fsTypeOptional.isPresent())) {
```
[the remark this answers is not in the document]
andrew-leung: I was leaving that to the `volumeIds.stream()` and following `.collect(Collectors.toList())`. I can add the explicit check to the referenced condition if strongly desired.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolumeUtils.java
```diff
@@ -0,0 +1,91 @@
...
+            return Collections.emptyList();
+        }
+
+        String mountPoint = mountPointOptional.get();
+        EbsVolume.MountPerm mountPerm = mountPermOptional.get();
+        String fsType = fsTypeOptional.get();
+
+        return volumeIds.stream()
+                .map(volumeId -> EbsVolume.newBuilder()
```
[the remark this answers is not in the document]
andrew-leung: The null and 0 values would indicate that those values have not yet been set. Currently, the expectation is that subsequent model validation will catch those not being set. Did you have something in mind for how to mark them?

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/common/DifferenceResolverUtils.java
```diff
@@ -246,12 +248,12 @@ public static int countActiveNotStartedTasks(EntityHolder refJobHolder, EntityHo
     /**
      * Determines unassigned IP allocations based on the current reference view
      */
-    public static Set<String> getUnassignedIpAllocations(JobView refJobView) {
-        // Get all IP allocations from the job
-        Set<String> unassignedIpAddressIds = refJobView.getJob().getJobDescriptor().getContainer().getContainerResources().getSignedIpAddressAllocations()
-                .stream()
+    public static List<String> getUnassignedIpAllocations(JobView refJobView) {
+        // Get a sorted list of all IP allocations from the job
+        List<String> unassignedIpAddressIds = refJobView.getJob().getJobDescriptor().getContainer().getContainerResources().getSignedIpAddressAllocations().stream()
```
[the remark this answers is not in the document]
andrew-leung: There should be no duplicates. I'll add a CollectionInvariant to ensure that is the case.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolumeUtils.java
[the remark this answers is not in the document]
tbak: Just a nit pick, to avoid the stream construction for empty case as we already make some other checks.
No need to change.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolumeUtils.java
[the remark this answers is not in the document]
tbak: Nothing concrete. I was just thinking what would be the best way to deal with partially constructed object.
In general, we do not pass any null values in core model to avoid NPEs.

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobEbsVolumeValidator.java
andrew-leung: I don't expect any major addition extensions. The reason this wasn't an EntitySanitizer is that the JOB_STRICT_SANITIZER EntitySanitizer is used on both Gateway and Coordinator. The object is only ready for validation on the Coordinator, so it seemed cleaner to have a separate way of validating on Coordinator.

tbak (approved):

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolume.java
```diff
@@ -0,0 +1,184 @@
...
+import com.google.common.base.Preconditions;
+
+public class EbsVolume {
+
+    public enum MountPerm {RO, RW}
+
+    public EbsVolume(String volumeId, String volumeAvailabilityZone, int volumeCapacityGB, String mountPath, MountPerm mountPerm, String fsType) {
+        this.volumeId = volumeId;
+        this.volumeAvailabilityZone = volumeAvailabilityZone;
+        this.volumeCapacityGB = volumeCapacityGB;
```
joshi-keyur, line 32: Should we use the same unit (MB) as the `disk` resource?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultDirectKubeApiServerIntegrator.java
```diff
@@ -140,27 +144,19 @@ public boolean isReadyForScheduling() {
...
-                throw new IllegalStateException("Unable to launch a task " + task.getId(), e);
+                logger.error("Unable to convert job {} and task {} to pod: {}", job, task, KubeUtil.toErrorDetails(e), e);
+                throw new IllegalStateException("Unable to convert task to pod " + task.getId(), e);
             }
-        }).subscribeOn(apiClientScheduler).timeout(Duration.ofMillis(configuration.getKubeApiClientTimeoutMs()));
+        })
+                .flatMap(v1Pod -> launchPersistentVolume(job, task, v1Pod, kubeApiFacade.getCoreV1Api())
+                        .then(Mono.just(v1Pod)))
+                .flatMap(v1Pod -> launchPod(task, v1Pod))
+                .subscribeOn(apiClientScheduler).timeout(Duration.ofMillis(configuration.getKubeApiClientTimeoutMs()));
```
joshi-keyur, line 159: Do we need to increase the timeout now that we are adding a call to `launchPersistentVolume`?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultDirectKubeApiServerIntegrator.java
```diff
@@ -218,6 +214,62 @@ public String resolveReasonCode(Throwable cause) {
...
+            try {
+                coreV1Api.createPersistentVolume(v1PersistentVolume, null, null, null);
+                logger.info("Created persistent volume {} in {}", v1PersistentVolume, timer.elapsed(TimeUnit.MILLISECONDS));
+                metrics.persistentVolumeCreateSuccess(timer.elapsed(TimeUnit.MILLISECONDS));
+            } catch (ApiException apiException) {
+                if (isEbsVolumeConflictException(apiException)) {
+                    logger.info("Persistent volume already exists {}", v1PersistentVolume);
+                } else {
+                    logger.error("Unable to create persistent volume {}, error: {}", v1PersistentVolume, apiException);
+                    metrics.persistentVolumeCreateError(apiException, timer.elapsed(TimeUnit.MILLISECONDS));
```
joshi-keyur, line 242: Is there a value in tracking errors by fsType and AZ?

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultTaskToPodConverter.java
```diff
@@ -440,4 +452,28 @@ private void setJobAcceptedTimestamp(TitanProtos.ContainerInfo.Builder container
 
         return Collections.singletonList(constraint);
     }
+
+    /**
+     * Builds the various objects needed to
```
joshi-keyur, line 457: incomplete comment? :)

joshi-keyur (approved):

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolume.java
andrew-leung: The minimum EBS volume size is GB, so I decided to leave it as GB as that is the smallest unit possible unit. Also, this value only translates to the size specified on the EBS PersistentVolume object, so doesn't map or relate to anything in the container's diskMB.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultDirectKubeApiServerIntegrator.java
andrew-leung: That's a good question. My impression is that the default 5 seconds will be sufficient for both create operations (i.e., 2 ApiServer create/write calls) based on the average latency we are seeing in today. Though the timeout is a FP, so we can adjust as collect more experience.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultDirectKubeApiServerIntegrator.java
andrew-leung: That's a good question. What we capture here is the same level of error detail we capture for pods. Specific fields that caused a create failure we buried in the Kube error message string.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolumeUtils.java
```diff
@@ -0,0 +1,90 @@
...
+                        .build())
+                .collect(Collectors.toList());
+    }
+
+    public static <E extends JobDescriptor.JobDescriptorExt> List<String> getVolumeIds(JobDescriptor<E> jobDescriptor) {
+        String volumeIdsStr = StringExt.nonNull(jobDescriptor.getAttributes().get(JobAttributes.JOB_ATTRIBUTES_EBS_VOLUME_IDS));
+        return StringExt.splitByComma(volumeIdsStr);
+    }
+
+    public static <E extends JobDescriptor.JobDescriptorExt> Optional<String> getMountPoint(JobDescriptor<E> jobDescriptor) {
```
solarkennedy, line 66: should this be `getEBSMountPoint`?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ebs/EbsVolumeUtils.java
```diff
@@ -0,0 +1,90 @@
...
+                .map(volumeId -> EbsVolume.newBuilder()
+                        .withVolumeId(volumeId)
+                        .withMountPath(mountPoint)
+                        .withMountPermissions(mountPerm)
+                        .withFsType(fsType)
+                        .build())
+                .collect(Collectors.toList());
+    }
+
+    public static <E extends JobDescriptor.JobDescriptorExt> List<String> getVolumeIds(JobDescriptor<E> jobDescriptor) {
```
solarkennedy, line 61: getEBSVolumeIDs?

solarkennedy (approved):

PR #953, opened 2020-11-12
rechecking agent status if not found in cache

amit-git:
It contains the following updates to the Kubernetes Node GC controller behavior

1. If node not found in AgentManagementService Cache, EC2 service call is made to fetch the instance status. 
2. If node status not found in the Cache or from the EC2 service call, then it's ignored for GC

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/agent/service/AgentManagementService.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/agent/service/AgentManagementService.java
@@ -28,6 +28,7 @@ import com.netflix.titus.api.agent.model.event.AgentEvent;
 import com.netflix.titus.api.model.ResourceDimension;
 import com.netflix.titus.api.model.Tier;
 import com.netflix.titus.common.util.tuple.Either;
+import reactor.core.publisher.Mono;
 import rx.Completable;
 import rx.Observable;
 
@@ -140,6 +141,6 @@ public interface AgentManagementService extends ReadOnlyAgentOperations {
      * @param instanceId of the agent
      * @return Observable of AgentInstance
      */
-    Observable<AgentInstance> getAgentInstanceAsync(String instanceId);
+    Mono<AgentInstance> getAgentInstanceAsync(String instanceId);
 
 }
--- a/titus-api/src/main/java/com/netflix/titus/api/agent/service/AgentManagementService.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/agent/service/AgentManagementService.java
@@ -32,7 +32,6 @@ import rx.Completable;
 import rx.Observable;
 
 public interface AgentManagementService extends ReadOnlyAgentOperations {
-
     /**
      * Returns true if the agent instance group is owned by Fenzo. Only Fenzo owned instance groups are scaled
      * by the capacity management and cluster operations components.
@@ -134,4 +133,13 @@ public interface AgentManagementService extends ReadOnlyAgentOperations {
      * event for each instance group or agent instance change (add/update/remove).
      */
     Observable<AgentEvent> events(boolean includeSnapshot);
+
+    /**
+     * Looks up agent instance by its ID, asynchronously
+     *
+     * @param instanceId of the agent
+     * @return Observable of AgentInstance
+     */
+    Observable<AgentInstance> getAgentInstanceAsync(String instanceId);
+
 }
--- a/titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AwsInstanceCloudConnector.java
+++ b/titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AwsInstanceCloudConnector.java
@@ -51,6 +51,7 @@ import com.amazonaws.services.autoscaling.model.TagDescription;
 import com.amazonaws.services.autoscaling.model.UpdateAutoScalingGroupRequest;
 import com.amazonaws.services.autoscaling.model.UpdateAutoScalingGroupResult;
 import com.amazonaws.services.ec2.AmazonEC2Async;
+import com.amazonaws.services.ec2.model.AmazonEC2Exception;
 import com.amazonaws.services.ec2.model.CreateTagsRequest;
 import com.amazonaws.services.ec2.model.CreateTagsResult;
 import com.amazonaws.services.ec2.model.DescribeInstancesRequest;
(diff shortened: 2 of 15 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -148,9 +147,15 @@ private boolean isAgentInstanceNotAvailable(String nodeName) {
 
         Optional<AgentInstance> agentInstanceOpt = agentManagementService.findAgentInstance(nodeName);
         if (!agentInstanceOpt.isPresent()) {
-            return true;
+            // re-check with agent service to find an instance
+            AgentInstance agentInstance = agentManagementService.getAgentInstanceAsync(nodeName)
+                    .toBlocking()
+                    .firstOrDefault(null);
+            if (agentInstance != null) {
+                return agentInstance.getLifecycleStatus().getState() == InstanceLifecycleState.Stopped;
```
joshi-keyur, line 155: Can LifecycleStatus be `null`?

joshi-keyur, line 155: Should we track how frequently this may be happening using a Counter?

amit-git: That's a good idea. I'll log that for now, can add more metrics if needed.

amit-git: Not sure, but an additional check won't hurt

titus-api/src/main/java/com/netflix/titus/api/agent/service/AgentManagementService.java
```diff
@@ -134,4 +133,13 @@
...
      */
     Observable<AgentEvent> events(boolean includeSnapshot);
+
+    /**
+     * Looks up agent instance by its ID, asynchronously
+     *
+     * @param instanceId of the agent
+     * @return Observable of AgentInstance
+     */
+    Observable<AgentInstance> getAgentInstanceAsync(String instanceId);
```
tbak, line 143: In a new code, let's use spring-reactor API (Flux in this case).

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -148,9 +150,16 @@ private boolean isAgentInstanceNotAvailable(String nodeName) {
 
         Optional<AgentInstance> agentInstanceOpt = agentManagementService.findAgentInstance(nodeName);
         if (!agentInstanceOpt.isPresent()) {
-            return true;
+            // re-check with agent service to find an instance
+            AgentInstance agentInstance = agentManagementService.getAgentInstanceAsync(nodeName)
+                    .toBlocking()
+                    .firstOrDefault(null);
+            if (agentInstance != null && agentInstance.getLifecycleStatus() != null) {
```
tbak, line 157: If after direct query agentInstance is null we should assume that it is not available, so check it first and return `true`?

tbak (changes requested):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
```diff
@@ -148,9 +150,16 @@ private boolean isAgentInstanceNotAvailable(String nodeName) {
 
         Optional<AgentInstance> agentInstanceOpt = agentManagementService.findAgentInstance(nodeName);
         if (!agentInstanceOpt.isPresent()) {
-            return true;
+            // re-check with agent service to find an instance
+            AgentInstance agentInstance = agentManagementService.getAgentInstanceAsync(nodeName)
+                    .toBlocking()
```
corindwyer, line 155: Verify this has a short timeout on it. If not, update code.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/NodeGcController.java
corindwyer: yes, this could should also handle the case of the instance is explicitly gone in AWS. This assumes you can query for the state forever but the EC2 API returns a 404 after the instance has been stopped for an hour IIRC.

corindwyer (changes requested):

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AwsInstanceCloudConnector.java
```diff
@@ -223,6 +227,37 @@ public ResourceDimension getInstanceTypeResourceDimension(String instanceType) {
                 });
     }
 
+    @Override
+    public Mono<Instance> getInstance(String instanceId) {
+        DescribeInstancesRequest req = new DescribeInstancesRequest()
```
tbak, line 232: Wrap it into `Mono.defer` so a new request is create on each subscribe.

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AwsInstanceCloudConnector.java
```diff
@@ -223,6 +227,37 @@ public ResourceDimension getInstanceTypeResourceDimension(String instanceType) {
...
+                        !isr.getReservations().isEmpty() &&
+                        !isr.getReservations().get(0).getInstances().isEmpty())
+                .map(isr -> {
+                    Reservation reservation = isr.getReservations().get(0);
+                    return toInstance(reservation.getInstances().get(0));
+                })
+                .onErrorResume(e -> {
+                    if (e instanceof AmazonEC2Exception) {
+                        AmazonEC2Exception ec2Exception = (AmazonEC2Exception) e;
+                        logger.warn("Exception fetching AWS instance {}", instanceId, ec2Exception);
```
tbak, line 248: Move logger after 400 check, as the latter is ok case? Otherwise it will be a lot of noise in the log file.

tbak (approved):

titus-ext/aws/src/main/java/com/netflix/titus/ext/aws/AwsInstanceCloudConnector.java
corindwyer: It might be useful to have a log line for when instance is not found without stack trace.

corindwyer (approved):

PR #959, opened 2020-11-25
Disable setting env variables in ContainerInfo gated by a fast property

joshi-keyur:
### Description of the Change

This feature, once enabled, will prevent setting env variables in ContainerInfo set as an annotation on the pod resources. 
Env vars are also being set in the pod spec which is now the primary source of this info. Removing the former will reduce redundancy, shrink the pod size and aid pod LIST latency when there are a number of pods in the system.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/config/MasterConfiguration.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/config/MasterConfiguration.java
@@ -92,4 +92,8 @@ public interface MasterConfiguration {
     @PropertyName(name = "mesos.slave.attribute.zone.name")
     @DefaultValue("zone")
     String getHostZoneAttributeName();
+
+    @PropertyName(name = "titus.jobcoordinator.pod.containerInfoEnvEnabled")
+    @DefaultValue("true")
+    boolean isContainerInfoEnvEnabled();
 }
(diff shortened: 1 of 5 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/config/MasterConfiguration.java
```diff
@@ -92,4 +92,8 @@
     @PropertyName(name = "mesos.slave.attribute.zone.name")
     @DefaultValue("zone")
     String getHostZoneAttributeName();
+
+    @PropertyName(name = "titus.jobcoordinator.pod.containerInfoEnvEnabled")
+    @DefaultValue("true")
+    boolean isContainerInfoEnvEnabled();
```
andrew-leung, line 98: Minor: could/should this be a `FeatureRolloutPlans`?

andrew-leung (approved):

tbak (approved):

corindwyer (approved):

titus-server-master/src/main/java/com/netflix/titus/master/config/MasterConfiguration.java
joshi-keyur: Thanks for the feedback. I am hoping to activate this change before the holidays. Hopefully, we won't need feature plan for this short term rollout! Also, it felt the feature plan evaluations are used more for surfacing violation messages to the user? Lmk if I am missing something.

PR #961, opened 2020-12-02
Align KubeScheduler pod -> task state mapping with KubeApiServerIntegrator

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/model/PodWrapper.java
```diff
@@ -41,6 +43,10 @@ public V1Pod getV1Pod() {
         return v1Pod;
     }
 
+    public String getName() {
+        return v1Pod.getMetadata() == null ? "" : StringExt.nonNull(v1Pod.getMetadata().getName());
```
corindwyer, line 47: There is `KubeUtil.getMetadataName`.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/PodToTaskMapper.java
```diff
@@ -0,0 +1,327 @@
...
+
+    /**
+     * Handle pod is assigned to a node but not running yet. The following pre-conditions must exist:
+     * <ul>
+     *     <li>pod condition 'PodScheduled' is set</li>
+     *     <li>nodeName is not empty</li>
+     * </ul>
+     * Based on container state and reason message the state is classified as 'Launched' or 'StartInitiated':
+     * <ul>
+     *     <li>if reason != 'TASK_STARTING' set 'Launched'</li>
```
corindwyer, line 196: I think we can use `V1ContainerStateRunning.getStartedAt` here in the future as these special reasons are going away.

corindwyer (approved):

joshi-keyur (approved):

PR #964, opened 2020-12-18
TITUS-4588 Add support for EBS persistent volume claims

andrew-leung:
This PR changes how EBS persistent volumes are being used. Persistent volumes directly specify the CSI driver rather than the legacy AWS block store and expecting Kube CSI migration to do the mapping. Also, Persistent volume claims are used rather than the specifying the persistent volume directly on the pod.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PersistentVolumeUnassociatedGcController.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PersistentVolumeUnassociatedGcController.java
@@ -41,6 +41,7 @@ import io.kubernetes.client.openapi.models.V1PersistentVolumeStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.netflix.titus.runtime.kubernetes.KubeConstants.DEFAULT_NAMESPACE;
 import static com.netflix.titus.runtime.kubernetes.KubeConstants.NOT_FOUND;
 
 @Singleton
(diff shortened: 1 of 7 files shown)


tbak (approved):

solarkennedy (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PersistentVolumeUnassociatedGcController.java
```diff
@@ -97,13 +98,26 @@ public boolean shouldGc() {
...
         try {
+            // If the PV is deleted while still associated with a PVC (though that is not expected), the PV
+            // will not be removed until it is no longer bound to a PVC.
+            // https://kubernetes.io/docs/concepts/storage/persistent-volumes/#storage-object-in-use-protection
             kubeApiFacade.getCoreV1Api().deletePersistentVolume(
                     volumeName,
                     null,
                     null,
-                    1,
+                    0,
```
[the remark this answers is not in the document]
andrew-leung: It is grace period seconds. I accidentally set it to 1 previously and am fixing here.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PersistentVolumeUnassociatedGcController.java
```diff
@@ -120,6 +134,38 @@ public boolean gcItem(V1PersistentVolume pv) {
         return false;
```
joshi-keyur, line 134: return `true` when it is not found in the `catch(ApiException)` block?

joshi-keyur (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PersistentVolumeUnassociatedGcController.java
andrew-leung: Yes, good catch!

PR #967, opened 2020-12-22
Adding counter to track conflicts between user and Titus provided env

joshi-keyur:
### Description of the Change

This is first in a series of PRs to:
* verify there are currently no conflicting env variables between user provided and Titus provide sets (this PR),
* enforce that user may not provide an env variable that is going to conflict with a Titus provided default, and 
* being able to reliably identify which env variables were provided by the user to recompute signature for verification when launching pods on Titus Agents


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/env/AggregatingContainerEnvFactory.java
```diff
@@ -17,31 +17,17 @@
...
-
-import com.netflix.titus.api.jobmanager.model.job.Job;
-import com.netflix.titus.api.jobmanager.model.job.Task;
 
 /**
  * Aggregate container environment variables from many sources. Evaluation happens from left to right, with the
  * next item overriding entries from previous evaluations if there is a collision.
  */
-public class AggregatingContainerEnvFactory implements ContainerEnvFactory {
+public abstract class AggregatingContainerEnvFactory implements ContainerEnvFactory {
```
tbak, line 26: Is there a value in extracting it as an abstract class?

tbak (approved):

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/env/DefaultAggregatingContainerEnvFactory.java
```diff
@@ -0,0 +1,61 @@
...
+            Map<String, String> envMap = factory.buildContainerEnv(job, task);
+            // Tracking conflicting env var for any two given factories
+            env.keySet().stream().filter(envMap::containsKey).forEach(this::incrementConflictCounter);
+            env.putAll(envMap);
+        }
+        return env;
+    }
+
+    private void incrementConflictCounter(String envVarName) {
+        this.registry.counter(this.conflictId.withTags("env", envVarName)).increment();
```
andrew-leung, line 59: The `this.` prefixes don't seem needed.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/env/DefaultAggregatingContainerEnvFactory.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private static final String CONFLICT_COUNTER = "titus.aggregatingContainerEnv.conflict";
+
+    private final Registry registry;
+
+    private final Id conflictId;
+
+    @Inject
+    public DefaultAggregatingContainerEnvFactory(TitusRuntime titusRuntime) {
+        super(UserProvidedContainerEnvFactory.getInstance(), TitusProvidedContainerEnvFactory.getInstance());
```
andrew-leung, line 41: It seems like we're depending on the order here to make sure that factories with higher precedence override others. If that's the case, maybe we should add a comment here or where we do the put()/overwrite?

andrew-leung (approved):

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/env/DefaultAggregatingContainerEnvFactory.java
joshi-keyur: Good point.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/env/AggregatingContainerEnvFactory.java
joshi-keyur: I will refactor and embed the initialization in the default implementation and remove this class for brevity. Thank you!

PR #969, opened 2021-01-06
ES task environment map sanitization

amit-git:
### Elastic search payload sanitization for environment variables

We have seen cases where ES payload publishing fails if task environment map contains keys with invalid use of period (".") characters. This code changes looks for leading or trailing "." and any instances of multiple "."s
It removes the invalid keys from the map before publishing to elastic search endpoint


tbak (approved):

titus-supplementary-component/tasks-publisher/src/main/java/com/netflix/titus/supplementary/taskspublisher/TaskDocument.java
```diff
@@ -564,6 +568,18 @@ private static TitusTaskState toV2TaskState(TaskStatus taskStatus) {
         }
     }
 
+    @VisibleForTesting
+    static Map<String, String> sanitizeEnvMap(Map<String, String> env) {
+        if (env == null) {
+            return Collections.emptyMap();
+        }
+        return env.keySet().stream().filter(TaskDocument::isSafe).collect(Collectors.toMap(k -> k, env::get));
```
corindwyer, line 576: Might be useful to have a log line or a metric to know how often this is happening.

corindwyer (approved):

amit-git:
Removing invalid ENV vars from the ES task document

PR #975, opened 2021-02-02
Report properly the caller id and reason in the job transaction log

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/common/action/TitusChangeAction.java
```diff
@@ -37,19 +39,25 @@
...
     }
 
-    public TitusChangeAction(Trigger trigger, String id, String name, String summary) {
+    public TitusChangeAction(Trigger trigger, String id, String name, String summary, CallMetadata callMetadata) {
         this.trigger = trigger;
         this.id = id;
         this.name = name;
         this.summary = summary;
+        // We assume that if call metadata is undefined, it comes from reconciler
+        this.callMetadata = callMetadata == null
```
corindwyer, line 58: Can the reconciler set it explicitly such that there is no assumption and empty means a real issue?

corindwyer (approved):

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/common/action/TitusChangeAction.java
tbak: I am doing it as a safety measure. This should never happen, unless there is a bug in the code.

tbak: Actually, maybe instead of assigning reconciler as a default, I should create a different one. Let me do that.

corindwyer (approved):

PR #976, opened 2021-02-05
RFC: Added more verbose and human-friendly reason messages for task updates

solarkennedy:
The real reason I started on this PR was because I couldn't remember how to tell if something was scheduled via kubescheduler.

Then I thought how nice it would be if in the reason message it was more explicit about how thing are going. Like what is the "real" state of things, and what is going to happen next (if applicable).

After all, what does "accepted" really mean? 

I was unable to track down where `created`, `running`, and `<not_given>` (no reason message) came from.

Anyway, seemed like it would be a nice to have, and better messages can assure our users (me) that the thing is working (or not). I'm not super familiar with this codebase or java.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
@@ -170,7 +170,7 @@ public class KubeNotificationProcessor {
 
         TaskStatus newTaskStatus = newTaskStatusOrError.getValue();
         if (TaskStatus.areEquivalent(task.getStatus(), newTaskStatus)) {
-            logger.info("Pod change notification does not change task status: taskId={}, status={}", task.getId(), newTaskStatus);
+            logger.info("Pod change notification does not change task status: taskId={}, status={}", task.getId (), newTaskStatus);
         } else {
             logger.info("Pod notification changes task status: taskId={}, fromStatus={}, toStatus={}", task.getId(),
                     task.getStatus(), newTaskStatus);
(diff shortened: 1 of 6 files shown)


tbak (approved):

tbak:
@solarkennedy Thank you for this PR. Some of the previous messages were very vague.

corindwyer (approved):

andrew-leung (approved):

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
```diff
@@ -182,7 +182,7 @@ public void shutdown() {
                 task.getId(),
                 current -> updateTaskStatus(podWrapper, newTaskStatus, executorDetailsOpt, node, current),
                 V3JobOperations.Trigger.Kube,
-                "Kube pod notification",
+                "Pod status updated from kubernetes node (k8s pod phase is now '" + event.getPod().getStatus().getPhase() + "')",
```
tbak, line 185: `getStatus()` can be null. There are KubeUtils for safe access.

tbak (changes requested):

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
solarkennedy: earlier in the code it checks for `event.getPod().getStatus().getPhase() == null` on line 146, is that enough?

tbak: If it also checks that `event.getPod().getStatus() != null` then yes.

solarkennedy: Yep. if that was null it would return early from line 146

PR #982, opened 2021-02-09
Add capacity group annotation to a pod

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
```diff
@@ -543,8 +548,9 @@ private V1Pod taskInfoToPod(TaskInfoRequest taskInfoRequest) {
         Protos.TaskInfo taskInfo = taskInfoRequest.getTaskInfo();
         String taskId = taskInfo.getName();
         String nodeName = taskInfo.getSlaveId().getValue();
+        String capacityGroup = JobManagerUtil.getCapacityGroupDescriptor(taskInfoRequest.getJob().getJobDescriptor(), capacityGroupManagement).getAppName().toLowerCase();
```
joshi-keyur, line 551: Lower case! nice

joshi-keyur (approved):

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/KubeApiServerIntegrator.java
tbak: Aligned with capacity controller

PR #983, opened 2021-02-09
Make the titus-relocation kill reasons more verbose

solarkennedy:
This does 2 things:

1. Makes the reason front and center in the message.
Previously this kinda looked like a k=v pairs.
2. Makes the messages more verbose.
Hopefully that is a good thing.
My main goal is to always actually say "why", where the
why needs have something about the disruption budget.

I tried to make it clear when we were respecting the jobs budget, and when we can't.
Hopefully that will make things less confusing for our users, and maybe a tad more sympathetic.

### DIFF
--- a/titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/util/RelocationPredicates.java
+++ b/titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/util/RelocationPredicates.java
@@ -57,23 +57,23 @@ public class RelocationPredicates {
         }
 
         if (isRelocationRequired(task)) {
-            return Optional.of("Task tagged for relocation");
+            return Optional.of("Task terminated because it was marked for relocation, and this was a safe time per this job's Disruption Budget");
         }
 
         if (isRelocationRequiredBy(job, task)) {
             long jobTimestamp = getJobTimestamp(job, RelocationAttributes.RELOCATION_REQUIRED_BY);
             long taskTimestamp = getTaskCreateTimestamp(task);
             if (jobTimestamp >= taskTimestamp) {
-                return Optional.of("Job tagged for relocation for tasks created before " + DateTimeExt.toUtcDateTimeString(jobTimestamp));
+                return Optional.of("Task terminated because tasks before " + DateTimeExt.toUtcDateTimeString(jobTimestamp) + " were marked for relocation");
             }
         }
 
         if (instance.isRelocationRequired()) {
-            return Optional.of("Agent instance tagged for eviction");
+            return Optional.of("Task terminated because the underlying Titus Agent was tagged for eviction, and this was a safe time per this job's Disruption Budget");
         }
 
         if (instance.isServerGroupRelocationRequired()) {
-            return Optional.of("Agent instance group tagged for eviction");
+            return Optional.of("Task terminated because the underlying Titus Agent instance group tagged for eviction, and this was a safe time per this job's Disruption Budget");
         }
 
         return Optional.empty();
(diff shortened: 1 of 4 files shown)


andrew-leung (approved):

amitaekbote (approved):

titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/util/RelocationPredicates.java
```diff
@@ -57,60 +57,60 @@
         }
 
         if (isRelocationRequired(task)) {
-            return Optional.of("Task tagged for relocation");
+            return Optional.of("Task killed because it was marked for relocation, and this was a safe time per this job's Disruption Budget");
```
tbak, line 60: I am not sure if this message is used only when the task is actually terminated, or in the termination plan as well. It might be misleading in the latter case. Checking that task must be relocated does not mean that its disruption budget is not violated.

tbak: We decided to stop using the word `kill` and instead use the word `terminate`.

titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/workflow/step/TaskEvictionStep.java
```diff
@@ -87,7 +87,7 @@ public TaskEvictionStep(EvictionServiceClient evictionServiceClient,
                 .collect(Collectors.toMap(
                         TaskRelocationPlan::getTaskId,
                         p -> {
-                            String message = String.format("%s: reasonCode=%s, plannedRelocationTime=%s",
+                            String message = String.format("%s\nreasonCode=%s plannedRelocationTime=%s",
```
tbak, line 90: Is this the new line character on purpose? I think we should keep all messages in a single line.

tbak (changes requested):

titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/util/RelocationPredicates.java
solarkennedy: Oh. Yea if it is only "tagged" (not sure what that means), then that would make sense, because there are other code paths where it is *actually* terminated.

titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/workflow/step/TaskEvictionStep.java
solarkennedy: It was on purpose, I thought it looked better. I'll do one line though.

titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/util/RelocationPredicates.java
solarkennedy: But If I'm reading the way this works, the plan is a kind of queue of reasons (`.right()`) so I think once it is tagged, this reason is shown when it is terminated too.

tbak: Please, have a look at the `TaskRelocationPlan` class. This is persisted in the database for tasks with self managed disruption budgets.

solarkennedy: Hmm. I guess I saw the `Immediate` functions above and assumed that they handled the cases where the Budget could not be respected. Are you sure this is not the part of the code that puts thing into the queue to be relocated that are *not* SelfManaged and *not* immediately relocated (and therefore, relocated due to budget?)?

> This is persisted in the database for tasks with self managed disruption budgets.

But if it is self-managed, this function returns empty right away on line 46?

tbak: The purpose of `RelocationPredicates` is to identify tasks that require relocation. The message should reflect it. We could create a different message when evicting the task in the relocation service.

solarkennedy: Gotcha. I understand now that it isn't as simple as I thought it was.
I'm going to discard this as is, because I agree it will be misleading.

PR #987, opened 2021-02-12
Assign critical tasks to the reserved capacity scheduler

tbak:


joshi-keyur (approved):

corindwyer (approved):

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DefaultTaskToPodConverter.java
```diff
@@ -181,8 +183,14 @@ public V1Pod apply(Job<?> job, Task task) {
                 .env(ContainerEnvs.toV1EnvVar(containerEnvFactory.buildContainerEnv(job, task)))
                 .resources(buildV1ResourceRequirements(job.getJobDescriptor().getContainer().getContainerResources()));
 
+        ApplicationSLA capacityGroupDescriptor = JobManagerUtil.getCapacityGroupDescriptor(job.getJobDescriptor(), capacityGroupManagement);
+        String kubeSchedulerName = configuration.getKubeSchedulerName();
+        if (capacityGroupDescriptor != null && capacityGroupDescriptor.getTier() == Tier.Critical) {
```
joshi-keyur, line 188: Should we also check the `schedulerName` property on ApplicationSLA to route it to `titus-kube-scheduler-reserved`?

tbak: We assume now that all KubeScheduler capacity groups are associated now with `titus-kube-scheduler-reserved` to keep it simple.

PR #990, opened 2021-03-01
Add pod annotations for EBS volumes and disable PV/PVC creation

andrew-leung:
Adds the following pod annotations to a pod for EBS configuration:
```
"ebs.volume.netflix.com/fsType": "xfs",
"ebs.volume.netflix.com/mountPath": "/ebs_mnt",
"ebs.volume.netflix.com/mountPerm": "RW",
"ebs.volume.netflix.com/volumeId": "vol-09a13c88ebc5c3597",
```

And disable creation of EBS PV and PVCs.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
@@ -64,4 +64,10 @@ public interface DirectKubeConfiguration extends KubeConnectorConfiguration {
      */
     @DefaultValue("false")
     boolean isKubeApiServerIntegratorOldPodCreationEnabled();
+
+    /**
+     * Set to true to enable EBS PV and PVC management.
+     */
+    @DefaultValue("false")
+    boolean isEbsVolumePvEnabled();
 }
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/kubernetes/KubeConstants.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/kubernetes/KubeConstants.java
@@ -43,6 +43,8 @@ public final class KubeConstants {
     /**
      * Common prefix for Titus node annotations/labels and taints.
      */
+    public static final String NETFLIX_DOMAIN = "netflix.com/";
+
     public static final String TITUS_DOMAIN = "titus.netflix.com/";
 
     public static final String TITUS_NODE_DOMAIN = "node.titus.netflix.com/";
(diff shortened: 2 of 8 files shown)


andrew-leung:
I added @rgulewich to vet the annotation style.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
```diff
@@ -64,4 +64,10 @@
      */
     @DefaultValue("false")
     boolean isKubeApiServerIntegratorOldPodCreationEnabled();
+
+    /**
+     * Set to true to enable EBS PV and PVC management.
+     */
+    @DefaultValue("false")
+    boolean isEbsVolumePvEnabled();
```
solarkennedy, line 72: Why do we have this still?
Are you thinking we might "someday" make PVs and such?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/kubernetes/KubeConstants.java
```diff
@@ -179,4 +181,20 @@ private KubeConstants() {
      * Reconciler Event Constants
      */
     public static final String NODE_LOST = "NodeLost";
+
+    /**
+     * Volume annotations.
+     */
+    public static final String VOLUME_DOMAIN = "volume." + NETFLIX_DOMAIN;
+    public static final String EBS_DOMAIN = "ebs." + VOLUME_DOMAIN;
+    public static final String EBS_VOLUME_ID = EBS_DOMAIN + "volume-id";
```
solarkennedy, line 190: camelcase on these?
https://kubernetes.io/docs/contribute/style/style-guide/#use-upper-camel-case-for-api-objects

andrew-leung: They were camelcase and @rgulewich mentioned the annotation style being used it lower case, dash separated (https://github.com/Netflix/titus-kube-common/blob/pod-spec/pod/annotations.go).

titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/direct/DirectKubeConfiguration.java
andrew-leung: More of a, "lets actually get this to production" before removing all of the other stuff. Once our path settles, all of that can be removed.

tbak (approved):

joshi-keyur (approved):

titus-server-runtime/src/main/java/com/netflix/titus/runtime/kubernetes/KubeConstants.java
rgulewich: The k8s docs don't explicitly say not to use camelcase, but all of the k8s annotations themselves use lowercase plus dashes, so we should try to be consistent: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

solarkennedy: Heh, got it.

PR #996, opened 2021-03-18
Job attribute for security auth context

andrew-leung:
This PR adds functionality in support of TITUS-4938. The feature uses Metatron auth context information for more secure/granular authz. This PR adds common attribute fields and removes them from created pods to avoid wasting space (the attributes are similarly not in ContainerInfo).

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
@@ -22,11 +22,13 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.InvalidProtocolBufferException;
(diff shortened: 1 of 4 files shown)


tbak (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
```diff
@@ -125,4 +128,13 @@
 
         return annotations;
     }
+
+    /**
+     * Returns a job descriptor with fields unnecessary for inclusion on the pod removed.
+     */
+    @VisibleForTesting
+    static  com.netflix.titus.api.jobmanager.model.job.JobDescriptor<?> filterPodJobDescriptor(com.netflix.titus.api.jobmanager.model.job.JobDescriptor<?> jobDescriptor) {
```
corindwyer, line 136: Shouldn't this just be `filterPodJobDescriptor`? I don't see where a pod comes into the picture.

corindwyer (approved):

PR #999, opened 2021-03-22
Apply AZ constraint to pods with Static IPs

andrew-leung:
Apply AZ node selector to pods with Static IPs for use with Kube Scheduler. Previously Static IP to AZ constraint mapping was handled by Fenzo.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/DefaultPodAffinityFactory.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/DefaultPodAffinityFactory.java
@@ -31,6 +31,8 @@ import com.netflix.titus.api.jobmanager.model.job.JobFunctions;
 import com.netflix.titus.api.jobmanager.model.job.Task;
 import com.netflix.titus.api.jobmanager.model.job.ebs.EbsVolume;
 import com.netflix.titus.api.jobmanager.model.job.ebs.EbsVolumeUtils;
+import com.netflix.titus.api.jobmanager.model.job.vpc.IpAddressAllocationUtils;
+import com.netflix.titus.common.runtime.TitusRuntime;
 import com.netflix.titus.common.util.StringExt;
 import com.netflix.titus.common.util.tuple.Pair;
 import com.netflix.titus.master.kubernetes.pod.resourcepool.PodResourcePoolResolver;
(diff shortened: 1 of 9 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/DefaultPodAffinityFactory.java
```diff
@@ -247,11 +252,26 @@ private void processZoneConstraints() {
...
-                addNodeAffinitySelectorConstraint(KubeConstants.NODE_LABEL_ZONE, optionalEbsVolume.get().getVolumeAvailabilityZone(), true);
-                return;
+            // If there is a Static IP allocation, get its zone
+            Optional<String> optionalIpAllocationZone = IpAddressAllocationUtils.getIpAllocationZoneForTask(job.getJobDescriptor(), task, titusRuntime.getCodeInvariants());
+
+            // If there is an EBS volume, get its zone
+            Optional<String> optionalEbsVolumeZone = EbsVolumeUtils.getEbsVolumeForTask(job, task)
+                    .map(EbsVolume::getVolumeAvailabilityZone);
+
+            if (optionalIpAllocationZone.isPresent() && optionalEbsVolumeZone.isPresent()) {
```
tbak, line 262: We should prevent this case to happen in the job service. Are there some checks there already?

tbak (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/DefaultPodAffinityFactory.java
```diff
@@ -247,11 +252,26 @@ private void processZoneConstraints() {
...
+                String ipAllocationZone = optionalIpAllocationZone.get();
+                String ebsVolumeZone = optionalEbsVolumeZone.get();
+                // If the zones for the two assigned resources do not match we do not assign either.
+                if (ebsVolumeZone.equals(ipAllocationZone)) {
```
joshi-keyur, line 266: Since we are implicitly coupling the zones for both these attributes here, do you think it makes sense to update how we populate these taskContext attributes in the first place so there is a higher likelihood of zone match success between the two?

joshi-keyur (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/DefaultPodAffinityFactory.java
andrew-leung: Discussed offline and will add validation in a subsequent PR.

PR #1000, opened 2021-03-29
De-duplicate the pod informer events in KubeNotificationProcessor

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
```diff
@@ -449,4 +466,27 @@ private static Task addMissingAttributes(PodWrapper podWrapper, Task updatedTask
         updatedAllocation.ifPresent(id -> builder.addToTaskContext(TASK_ATTRIBUTES_OPPORTUNISTIC_CPU_ALLOCATION, id));
         return builder.build();
     }
+
+    /**
+     * 'updatedTask' is a modified copy of 'currentTask' (or exactly the same version of the object if nothing changed).
+     */
+    @VisibleForTesting
+    static boolean areTasksEquivalent(Task currentTask, Task updatedTask) {
+        if (currentTask == updatedTask) {
```
corindwyer, line 475: Guessing this is just defensive programming? I don't think this case is possible though since the objects are created on demand every event.

corindwyer (approved):

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
```diff
@@ -449,4 +466,27 @@ private static Task addMissingAttributes(PodWrapper podWrapper, Task updatedTask
         updatedAllocation.ifPresent(id -> builder.addToTaskContext(TASK_ATTRIBUTES_OPPORTUNISTIC_CPU_ALLOCATION, id));
         return builder.build();
     }
+
+    /**
+     * 'updatedTask' is a modified copy of 'currentTask' (or exactly the same version of the object if nothing changed).
+     */
+    @VisibleForTesting
+    static boolean areTasksEquivalent(Task currentTask, Task updatedTask) {
```
joshi-keyur, line 474: Do you think there is value in tracking the first encountered reason why the tasks may be different? That may provide us insights into further optimizing or potentially extending this check include other parts of the task object.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
tbak: No, it is possible.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/KubeNotificationProcessor.java
tbak: I have been thinking about that as well. I will add additional logging.

joshi-keyur (approved):

PR #1001, opened 2021-03-29
Adding capability to mint new jobIds in federation

joshi-keyur:
### Description of the Change

This work enables the concept of FederationJobs which could span across deployment cells/stacks. As a first step, we need to create the JobIds externally and then pass it the deployment stacks so the job records can be created in external storage specific to the stack.

### DIFF
--- a/titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
+++ b/titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
@@ -138,15 +138,14 @@ public class AggregatingJobServiceGateway implements JobServiceGateway {
         }
         JobManagementServiceStub client = optionalClient.get();
 
-        JobDescriptor jobDescriptorWithStackName = addStackName(jobDescriptor);
-        JobDescriptor enrichedJobDescriptor;
+        JobDescriptor.Builder jobDescriptorBuilder = addStackName(jobDescriptor.toBuilder());
         if (federationConfiguration.isFederationJobIdCreationEnabled()) {
             String federatedJobId = UUID.randomUUID().toString();
-            enrichedJobDescriptor = addFederationAttributes(jobDescriptorWithStackName, federatedJobId, cell.getName());
+            jobDescriptorBuilder = addFederationAttributes(jobDescriptorBuilder, federatedJobId, cell.getName());
         } else {
-            enrichedJobDescriptor = removeFederationAttributes(jobDescriptorWithStackName);
+            jobDescriptorBuilder = removeFederationAttributes(jobDescriptorBuilder);
         }
-
+        JobDescriptor enrichedJobDescriptor = jobDescriptorBuilder.build();
         return createRequestObservable(emitter -> {
             StreamObserver<JobId> streamObserver = GrpcUtil.createClientResponseObserver(
                     emitter,
@@ -460,19 +459,17 @@ public class AggregatingJobServiceGateway implements JobServiceGateway {
         return result.toCompletable();
     }
 
-    private JobDescriptor addFederationAttributes(JobDescriptor jobDescriptor, String federatedJobId, String routingCell) {
-        return jobDescriptor.toBuilder().putAllAttributes(CollectionsExt.<String, String>newHashMap()
+    private JobDescriptor.Builder addFederationAttributes(JobDescriptor.Builder jobDescriptorBuilder, String federatedJobId, String routingCell) {
+        return jobDescriptorBuilder.putAllAttributes(CollectionsExt.<String, String>newHashMap()
                 .entry(JOB_ATTRIBUTES_FEDERATED_JOB_ID, federatedJobId)
                 .entry(JOB_ATTRIBUTE_ROUTING_CELL, routingCell)
-                .toMap())
-                .build();
+                .toMap());
     }
 
-    private JobDescriptor removeFederationAttributes(JobDescriptor jobDescriptor) {
-        return jobDescriptor.toBuilder()
+    private JobDescriptor.Builder removeFederationAttributes(JobDescriptor.Builder jobDescriptorBuilder) {
+        return jobDescriptorBuilder
                 .removeAttributes(JOB_ATTRIBUTES_FEDERATED_JOB_ID)
-                .removeAttributes(JOB_ATTRIBUTE_ROUTING_CELL)
-                .build();
+                .removeAttributes(JOB_ATTRIBUTE_ROUTING_CELL);
     }
 
     private JobQueryResult addStackName(JobQueryResult result) {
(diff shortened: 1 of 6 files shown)


tbak (approved):

titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
```diff
@@ -135,15 +138,24 @@ public AggregatingJobServiceGateway(GrpcConfiguration grpcConfiguration,
         }
         JobManagementServiceStub client = optionalClient.get();
 
-        JobDescriptor withStackName = addStackName(jobDescriptor);
+        JobDescriptor enrichedJobDescriptor;
+        if (federationConfiguration.isFederationJobIdCreationEnabled()) {
+            String federatedJobId = UUID.randomUUID().toString();
+            enrichedJobDescriptor = addJobAttributes(jobDescriptor,
+                    ImmutableMap.of(JOB_ATTRIBUTES_STACK, federationConfiguration.getStack(),
+                            JOB_ATTRIBUTES_FEDERATED_JOB_ID, federatedJobId));
```
gabrielhartmann, line 146: It would be good to also have the "routed cell" from line 128 included in the attributes as well.

gabrielhartmann (approved):

tbak:
@joshi-keyur `DefaultV3JobOperations` junit test would be nice.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
```diff
@@ -570,13 +571,15 @@ public Completable recordTaskPlacement(String taskId, Function<Task, Task> chang
     }
 
     private <E extends JobDescriptor.JobDescriptorExt> Job<E> newJob(JobDescriptor<E> jobDescriptor) {
+        String federatedJobId = jobDescriptor.getAttributes().get(JobAttributes.JOB_ATTRIBUTES_FEDERATED_JOB_ID);
```
corindwyer, line 574: Should this be removed from the `jobDescriptor` once it is read? My concern is that this gets copied over and we have id conflicts.

titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
```diff
@@ -449,6 +461,13 @@ public Completable moveTask(TaskMoveRequest taskMoveRequest, CallMetadata callMe
         return result.toCompletable();
     }
 
+    private JobDescriptor addJobAttributes(JobDescriptor jobDescriptor, Map<String, String> attributes) {
+        if (attributes != null) {
```
fabiokung, line 465: nit: `CollectionsExt.isNullOrEmpty()`

titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
```diff
@@ -135,15 +138,24 @@ public AggregatingJobServiceGateway(GrpcConfiguration grpcConfiguration,
         }
         JobManagementServiceStub client = optionalClient.get();
 
-        JobDescriptor withStackName = addStackName(jobDescriptor);
+        JobDescriptor enrichedJobDescriptor;
```
fabiokung, line 141: the enrichment below is not fully backwards compatible, e.g.: if the `federationJobIdCreation` flag is disabled, jobs that are cloned and already have a `JOB_ATTRIBUTES_FEDERATED_JOB_ID` set will cause jobs with duplicated ids to be created. It also would allow users to directly provide an id for reasons other than cloning, which can be dangerous

fabiokung: I would recommend stripping out (sanitizing) the attribute as the first step, then conditionally add it after if the configuration (feature) flag is enabled

fabiokung (changes requested):

titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
```diff
@@ -135,15 +138,24 @@ public AggregatingJobServiceGateway(GrpcConfiguration grpcConfiguration,
...
+        if (federationConfiguration.isFederationJobIdCreationEnabled()) {
+            String federatedJobId = UUID.randomUUID().toString();
+            enrichedJobDescriptor = addJobAttributes(jobDescriptor,
+                    ImmutableMap.of(JOB_ATTRIBUTES_STACK, federationConfiguration.getStack(),
```
fabiokung, line 145: logic being duplicated from `addStackName()`

titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
gabrielhartmann: No.  `addStackName` is adding the stack from here:
```
federationConfiguration.getStack()
```
which is different than the cell the job is being routed to.

fabiokung: ah, you are right! We fill in cell info later down the road (`JobAttributes.JOB_ATTRIBUTES_CELL`)

titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
joshi-keyur: This is a good point. I think this is in alignment with what @corindwyer was suggesting above. 
If we strip out the attribute from the JobDescriptor once it lands on the job coordinator, the case you are highlighting can also be addressed. However, I do see value in keeping it on the JobDescriptor to validate it was in fact minted by federation. 
I am leaning towards adding code to remove the attribute if found (irrespective of the feature flag value) and then conditionally adding it to the JobDescriptor. Does that sound good, @corindwyer and @fabiokung?

titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
joshi-keyur: We are avoiding calling `build()`. What do you propose?

fabiokung: re-shuffling things a bit to avoid calling `build()` in `addStackName()` too early might be enough. It seems we always want to add the stackName, so my suggestion would be to factor it out of the if/else chain and always add it unconditionally keeping the `Builder` alive by calling `.build()` as late as possible.

fabiokung (approved):

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
joshi-keyur: Thank you. I will create a separate PR for handling this.

PR #1004, opened 2021-04-01
Adding feature flag for routing Jobs with EBS volumes to KubeScheduler

joshi-keyur:

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/KubeSchedulerFeatureConfiguration.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/KubeSchedulerFeatureConfiguration.java
@@ -40,4 +40,10 @@ public interface KubeSchedulerFeatureConfiguration {
      */
     @DefaultValue(".*")
     String getEnabledMachineTypes();
+
+    /**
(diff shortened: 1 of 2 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/KubeSchedulerFeatureConfiguration.java
```diff
@@ -40,4 +40,10 @@
      */
     @DefaultValue(".*")
     String getEnabledMachineTypes();
+
+    /**
+     * Set to true to enable routing jobs with EBS volume configuration to KubeScheduler.
+     */
+    @DefaultValue("false")
```
andrew-leung, line 47: Will there be a corresponding nftitus PR to enable this or will this just be done via FP?

andrew-leung (approved):

tbak (approved):

titus-server-runtime/src/main/java/com/netflix/titus/runtime/KubeSchedulerFeatureConfiguration.java
joshi-keyur: This property should be sufficient.

PR #1005, opened 2021-04-01
Add metrics to the GRPC job service event stream

tbak:
+ more eagerly remove completed subscribers in `EventDistributor`

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
@@ -24,12 +24,17 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.UUID;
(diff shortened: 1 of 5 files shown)


joshi-keyur (approved):

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/endpoint/v3/grpc/DefaultJobManagementServiceGrpc.java
```diff
@@ -602,6 +615,12 @@ public void deleteTaskAttributes(TaskAttributesDeleteRequest request, StreamObse
 
     @Override
     public void observeJobs(ObserveJobsQuery query, StreamObserver<JobChangeNotification> responseObserver) {
+        Stopwatch start = Stopwatch.createStarted();
+
+        String trxId = UUID.randomUUID().toString();
```
joshi-keyur, line 620: Do you see value in using thread ID as `trxId` in case we want to correlated it with observations metrics/logs.

tbak: We do not log this value, only use as a unique id to correlate metrics.

PR #1006, opened 2021-04-06
GRPC findJobs/findTasks optimizations

tbak:
Java stream API and collection sorting is expensive. These changes gave 30% performance gain with 30K running tasks.

### DIFF
--- a/titus-common-api/src/main/java/com/netflix/titus/api/model/IdAndTimestampKey.java
+++ b/titus-common-api/src/main/java/com/netflix/titus/api/model/IdAndTimestampKey.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.api.model;
+
+import com.google.common.base.Preconditions;
+
+public class IdAndTimestampKey<T> implements PaginableEntityKey<T> {
+
+    private final T entity;
+    private final String id;
+    private final long timestamp;
+
+    public IdAndTimestampKey(T entity, String id, long timestamp) {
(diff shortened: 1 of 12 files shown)


titus-common-api/src/main/java/com/netflix/titus/api/model/IdAndTimestampKey.java
```diff
@@ -0,0 +1,60 @@
...
+    public int compareTo(PaginableEntityKey<T> other) {
+        Preconditions.checkArgument(other instanceof IdAndTimestampKey);
+
+        IdAndTimestampKey<T> otherKey = (IdAndTimestampKey<T>) other;
+        int cmp = Long.compare(timestamp, otherKey.timestamp);
+        if (cmp != 0) {
+            return cmp;
+        }
+        return id.compareTo(otherKey.id);
+    }
```
andrew-leung, line 59: Useful to add a toString() method for debugging?

titus-server-runtime/src/main/java/com/netflix/titus/runtime/jobmanager/JobComparators.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+        // Slow path
+        for (int i = 1; i < historySize; i++) {
+            JobStatus next = statusHistory.get(i);
+            if (next.getState() == JobState.Accepted) {
+                return next.getTimestamp();
+            }
+        }
+        // No Accepted state
+        return job.getStatus().getTimestamp();
```
andrew-leung, line 96: This case seems unexpected. Worth logging or otherwise noting it?

andrew-leung (approved):

joshi-keyur:
I am surprised by this discovery. Would you say there was a predominant factor (sorting/temp data structures/pagination) that contributed to the cost before?

joshi-keyur (approved):

titus-server-runtime/src/main/java/com/netflix/titus/runtime/jobmanager/JobComparators.java
tbak: This is unexpected but not strictly required. We just fallback to another state if the `Accepted` state is not found.
Invariant check would be useful, but should be added at the Job/Task creation point.

titus-common-api/src/main/java/com/netflix/titus/api/model/IdAndTimestampKey.java
tbak: Good idea. I will also added equals and hashCode.

tbak:
@joshi-keyur Java stream API is expensive and we use it in many places. I was also surprised how expensive it was to sort a collection of tasks in a paginated request. With 30K it dominated the FlameGraph, most of the time doing task comparison. There are other places in the code which could be optimized further with little effort.

PR #1007, opened 2021-04-08
Grpc object cache

tbak:


titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcObjectsCacheConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+import com.netflix.archaius.api.annotations.DefaultValue;
+
+@Configuration(prefix = "titus.grpc.cache")
+public interface GrpcObjectsCacheConfiguration {
+
+    /**
+     * @return whether or not the cache is enabled
+     */
+    @DefaultValue("true")
+    boolean isGrpcObjectsCacheEnabled();
```
joshi-keyur, line 13: Should we separate into two properties for jobs and tasks in case we want to disable caching for either one for performance reasons?

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/ProtobufCache.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    void shutdown() {
+        scheduleRef.cancel();
+    }
+
+    C get(String id, E entity) {
+        CacheEntry<E, C> entry = cache.get(id);
+        if (entry != null) {
+            if (entry.getCoreEntity() == entity) {
```
joshi-keyur, line 92: Do you think we should use equality check here? I am curious how many times we would hit the negative of this condition but not a complete cache miss.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcObjectsCacheConfiguration.java
tbak: I think we should be able to disable cache as a whole in case there is a problem with it. It is very unlikely we might get odd behavior with one entity, and normal with another.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/ProtobufCache.java
tbak: No, it would be too expensive. It is done on purpose like that. JobService will always return the same, latest value. If that value changes (even if equals is true), we should update the cache so we can depend on fast reference comparison.

andrew-leung (approved):

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/DefaultGrpcObjectsCache.java
```diff
@@ -0,0 +1,128 @@
...
+    public void shutdown() {
+        Evaluators.acceptNotNull(jobCache, ProtobufCache::shutdown);
+        Evaluators.acceptNotNull(taskCache, ProtobufCache::shutdown);
+    }
+
+    public Job getJob(com.netflix.titus.api.jobmanager.model.job.Job<?> coreJob) {
+        if (configuration.isGrpcObjectsCacheEnabled()) {
+            return jobCache.get(coreJob.getId(), coreJob);
+        } else {
+            return toGrpcJob(coreJob);
```
joshi-keyur, line 90: If the object cache is disabled, this method invocation would trigger serialization of the `job` object every time `getJob` is called for the `coreJob`. Should this be optimized?

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/ProtobufCache.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private static final String METRICS_ROOT = "protobufCache.";
+
+    private final Function<E, C> converter;
+    private final Supplier<Predicate<E>> expiredPredicateSupplier;
+    private final GrpcObjectsCacheConfiguration configuration;
+    private final Clock clock;
+
+    @VisibleForTesting
+    final ConcurrentMap<String, CacheEntry<E, C>> cache = new ConcurrentHashMap<>();
```
fabiokung, line 49: did you consider using a Guava cache with capacity-based LRU eviction, instead of relying only on time-based periodic evictions? It would be more resilient to memory blowing up and potentially simpler implementation?

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/ProtobufCache.java
```diff
@@ -0,0 +1,151 @@
...
+                cache.remove(entry.getKey());
+            }
+        }
+
+        // Find expired entries
+        Predicate<E> expiredPredicate = expiredPredicateSupplier.get();
+        Set<String> candidates = CollectionsExt.copyAndRemove(cache.keySet(), toBeRemoved.keySet());
+        candidates.forEach(id -> {
+            CacheEntry<E, C> entry = cache.get(id);
+            if (expiredPredicate.test(entry.getCoreEntity())) {
```
fabiokung, line 121: there's quite a bit of re-work here, since the predicate for each entity will recompute the `all[Job|Task]Ids` `HashMap` for entry. It's probably worth changing the API here to compute `items that should be present` only once to be tested against all entries in the cache.

Alternatively, change to capacity based eviction (instead of periodic time-based cleanup) as I mentioned on a previous comment.

fabiokung (approved):

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/ProtobufCache.java
tbak: I wanted to use Caffeine, but in the end I decided that it will be better to write a custom cache implementation.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/ProtobufCache.java
tbak: We want to minimize calls to JobService API which are relatively expensive. So before checking the expiry of cached entries, we fetch all of them with a single call. Next, we use the cached value (in the clojure), to check every item (this is fast).

fabiokung: Yeah, I get the intent, let me rephrase for clarity. The `expiredPredicate` used here (reproduced below) is evaluated for each candidate (all cache entries), and each time it is evaluated, it recomputes the `known[Job|Task]Ids` Set:

```diff
+                () -> {
+                    List<com.netflix.titus.api.jobmanager.model.job.Job> allJobs = jobOperations.getJobs();
+                    Set<String> knownJobIds = new HashSet<>();
+                    allJobs.forEach(job -> knownJobIds.add(job.getId()));
+                    return job -> !knownJobIds.contains(job.getId());
+                },
+
```

I know `jobOperations.getJobs()` is all in-memory and is a lightweight operation, and building a `HashSet` out of it is not expensive. It's probably not a bottleneck, but removing the O(N^2) here and computing `known[Job|Tasks]Ids` set only once every eviction cycle would be a simple enough optimization that is probably worth it.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/DefaultGrpcObjectsCache.java
```diff
@@ -0,0 +1,128 @@
...
+        if (configuration.isGrpcObjectsCacheEnabled()) {
+            return taskCache.get(coreTask.getId(), coreTask);
+        } else {
+            return toGrpcTask(coreTask);
+        }
+    }
+
+    private Task toGrpcTask(com.netflix.titus.api.jobmanager.model.job.Task coreTask) {
+        Task task = GrpcJobManagementModelConverters.toGrpcTask(coreTask, logStorageInfo);
+        // Need to serialize it once since GeneratedMessageV3 isn't entirely immutable.
```
fabiokung, line 104: minor: move the duplicated comment to javadoc of the method being called

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/ProtobufCache.java
fabiokung: My bad, I didn't realize we were holding a `Supplier<Predicate>` here, and computing the whole `Predicate` only once before checking each item. Nevermind my comments, LGTM 👍

PR #1010, opened 2021-04-13
Add garbage collector to TaskRelocation service

tbak:


titus-ext/jooq/src/main/java/com/netflix/titus/ext/jooq/relocation/JooqTaskRelocationResultStore.java
```diff
@@ -146,6 +148,24 @@ private void createSchemaIfNotExist() {
         });
     }
 
+    /**
+     * Removes all entries older than the given time threshold.
+     */
+    int removeExpiredData(long timeThreshold) {
```
corindwyer, line 154: I think this method should first do a count based on an index to see how many items there are matching the criteria to be deleted and then delete a configured amount of them. That way you can set a metric that shows how many items are eligible for deletion versus are being deleted in the case of large spikes. Also, you can write a better log debug statement saying which items are explicitly being deleted in each batch so it is easier to troubleshoot.

titus-ext/jooq/src/main/java/com/netflix/titus/ext/jooq/relocation/JooqTaskRelocationGC.java
```diff
@@ -0,0 +1,167 @@
...
+                .where(RELOCATION_STATUS.RELOCATION_EXECUTION_TIME.le(lastToRemove))
+                .orderBy(RELOCATION_STATUS.RELOCATION_EXECUTION_TIME.asc())
+                .fetch();
+
+        List<Pair<String, Long>> toRemoveSet = toRemoveRows.stream()
+                .map(r -> Pair.of(r.get(RELOCATION_STATUS.TASK_ID), r.get(RELOCATION_STATUS.RELOCATION_EXECUTION_TIME).getTime()))
+                .collect(Collectors.toList());
+        logger.info("Records to remove: {}", toRemoveSet);
+
+        int removedFromDb = dslContext.delete(RELOCATION_STATUS).where(RELOCATION_STATUS.RELOCATION_EXECUTION_TIME.le(lastToRemove)).execute();
```
corindwyer, line 158: not sure how this is implemented, but it is usually good practice to partition the arrays to a max size as there is a limit to how many ids can be passed in the `delete where clause`. It is probably fine though based on the current configuration values.

corindwyer (approved):

titus-ext/jooq/src/main/java/com/netflix/titus/ext/jooq/relocation/JooqTaskRelocationGC.java
tbak: In this implementation there is no need to send the individual ids. All rows older than the resolved time threshold are removed.

corindwyer: oh, I misread that line. Ignore this comment 😄

titus-ext/jooq/src/main/java/com/netflix/titus/ext/jooq/relocation/JooqTaskRelocationGC.java
```diff
@@ -0,0 +1,167 @@
...
+        int allCount = dslContext.fetchCount(RELOCATION_STATUS);
+        logger.info("All rows in 'relocation_status' table: {}", allCount);
+        allRowsGauge.set(allCount);
+
+        int expiredCount = dslContext.fetchCount(
+                RELOCATION_STATUS,
+                RELOCATION_STATUS.RELOCATION_EXECUTION_TIME.lt(Timestamp.from(Instant.ofEpochMilli(timeThreshold)))
+        );
+        logger.info("Expired rows in 'relocation_status' table: {}", expiredCount);
+        expiredRowsGauge.set(allCount);
```
fabiokung, line 124: copy-and-pasta error? I belive the argument here should be `expiredCount`: `expiredRowsGauge.set(expiredCount)`

titus-ext/jooq/src/main/java/com/netflix/titus/ext/jooq/relocation/JooqTaskRelocationGC.java
```diff
@@ -0,0 +1,167 @@
...
+
+        List<Pair<String, Long>> toRemoveSet = toRemoveRows.stream()
+                .map(r -> Pair.of(r.get(RELOCATION_STATUS.TASK_ID), r.get(RELOCATION_STATUS.RELOCATION_EXECUTION_TIME).getTime()))
+                .collect(Collectors.toList());
+        logger.info("Records to remove: {}", toRemoveSet);
+
+        int removedFromDb = dslContext.delete(RELOCATION_STATUS).where(RELOCATION_STATUS.RELOCATION_EXECUTION_TIME.le(lastToRemove)).execute();
+
+        logger.info("Removed expired rows from 'relocation_status' table: {}", removedFromDb);
+        gcCounter.increment(allCount);
```
fabiokung, line 161: copy-and-pasta? `s/allCount/removedFromDb/`

titus-ext/jooq/src/main/java/com/netflix/titus/ext/jooq/relocation/JooqTaskRelocationGC.java
```diff
@@ -0,0 +1,167 @@
...
+
+        relocationResultStore.removeFromCache(timeThreshold);
```
fabiokung, line 163: minor (feel free to ignore): we have all the ids in `toRemoveSet` above, could invalidate only those to avoid unnecessary cache misses

fabiokung (changes requested):

PR #1011, opened 2021-04-13
Add GRPC fallback test

gabrielhartmann:
### Description of the Change

In order to slowly and safely introduce redirecting calls to the new federation service, a first step should be to dual write a subset of requests to the current and new services simultaneously.  In the initial commit presented here we see a couple tests illustrating the proposed approach.  Thanks to @corindwyer for originally suggesting this "fallback" style.  The core rx.java primitive used to implement the fallback is the `onErrorResumeNext` method.

Today the new service returns an `UNIMPLEMENTED` status code for all JobService methods and will continue to do so until it takes over responsibility for portions of the interface.  So the `unimplementedFallback` test demonstrates that a federation service returning `UNIMPLEMENTED` falls back to the current cellular interface.

The `timeoutFallback` test illustrates the case in which a new federation service instance becomes either slow or entirely unresponsive to API calls.  Again we demonstrate fallback to the cellular interface.

Essentially the proposal is that for `create`, `update`, and `delete` calls on Jobs, dual writes will occur, with fallback to the cellular interface being the primary execution path.  The new federation service will perform operations on FederatedJobDescriptor CRDs, but will always return `UNIMPLEMENTED` to ensure that the current code path is executed.  

In this way we can start to observe the performance characteristics of the new federation service as well as determine what it _would have done_ to ensure functional correctness in comparison to the current system.

Before I make the rest of the changes to plumb in communication with the new federation service, I wanted to make sure we're aligned on the core "dual write" primitive.

### DIFF
--- a/titus-server-federation/src/test/java/com/netflix/titus/federation/service/GrpcFallbackTest.java
+++ b/titus-server-federation/src/test/java/com/netflix/titus/federation/service/GrpcFallbackTest.java
@@ -1,0 +1,115 @@
+package com.netflix.titus.federation.service;
+
+import com.netflix.titus.grpc.protogen.JobDescriptor;
+import com.netflix.titus.grpc.protogen.JobId;
+import com.netflix.titus.grpc.protogen.JobManagementServiceGrpc;
+import com.netflix.titus.runtime.endpoint.common.grpc.GrpcUtil;
+import io.grpc.stub.StreamObserver;
+import io.grpc.testing.GrpcServerRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import rx.Observable;
+
+import java.util.Optional;
+
+import static com.netflix.titus.runtime.endpoint.common.grpc.GrpcUtil.createRequestObservable;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class JobManagementServiceWithUnimplementedInterface extends JobManagementServiceGrpc.JobManagementServiceImplBase {
+    // All interface methods return UNIMPLEMENTED status.
+}
+
+class JobManagementServiceWithSlowMethods extends JobManagementServiceGrpc.JobManagementServiceImplBase {
+    public long createCount = 0;
+
+    public void createJob(com.netflix.titus.grpc.protogen.JobDescriptor request,
+                          io.grpc.stub.StreamObserver<com.netflix.titus.grpc.protogen.JobId> responseObserver) {
+        // Increment call count, but never respond to the responseObserver
+        createCount++;
+    }
+}
+
+public class GrpcFallbackTest {
+    private static final String appName = "app";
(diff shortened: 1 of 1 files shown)


titus-server-federation/src/test/java/com/netflix/titus/federation/service/GrpcFallbackTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public void createJob(com.netflix.titus.grpc.protogen.JobDescriptor request,
+                          io.grpc.stub.StreamObserver<com.netflix.titus.grpc.protogen.JobId> responseObserver) {
+        // Increment call count, but never respond to the responseObserver
+        createCount++;
+    }
+}
+
+public class GrpcFallbackTest {
+    private static final String appName = "app";
```
fabiokung, line 34: minor nit: code conventions for these types of constants is `CAPITAL_UNDERSCORED`

titus-server-federation/src/test/java/com/netflix/titus/federation/service/GrpcFallbackTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+import static com.netflix.titus.runtime.endpoint.common.grpc.GrpcUtil.createRequestObservable;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class JobManagementServiceWithUnimplementedInterface extends JobManagementServiceGrpc.JobManagementServiceImplBase {
+    // All interface methods return UNIMPLEMENTED status.
+}
+
+class JobManagementServiceWithSlowMethods extends JobManagementServiceGrpc.JobManagementServiceImplBase {
+    public long createCount = 0;
```
fabiokung, line 24: minor: set the field as `volatile` to avoid it biting us in the future in case we stop using `directExecutor()`

fabiokung (approved):
LGTM, I think that while we experiment with the new service it would be better to hedge the request (concurrently to both implementations), instead of serializing it through them to avoid the latency penalty to users while we get more data on the new service.

This is fine though, my only request is that you add a feature toggle to quickly disable the double write.

One thing that is not clear is if you intend to merge this test, or if it was only used to illustrate your proposal.

gabrielhartmann:
> LGTM, I think that while we experiment with the new service it would be better to hedge the request (concurrently to both implementations), instead of serializing it through them to avoid the latency penalty to users while we get more data on the new service.
> 
> This is fine though, my only request is that you add a feature toggle to quickly disable the double write.
> 
> One thing that is not clear is if you intend to merge this test, or if it was only used to illustrate your proposal.

Thank you for having a look and giving some feedback @fabiokung.

* I will add a feature toggle to enable/disable
* I don't intend this PR as something to merge.  I'll be putting out a real PR incorporating the method presented here.  Was just hoping for early feedback on something small before a bigger PR.

PR #1015, opened 2021-04-14
Move jooq common code to its own subproject

tbak:


titus-ext/jooq-common/src/main/java/com/netflix/titus/ext/jooq/JooqConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+package com.netflix.titus.ext.jooq;
+
+import com.netflix.archaius.api.annotations.DefaultValue;
+
+public interface JooqConfiguration {
+
+    @DefaultValue("jdbc://localhost")
+    String getDatabaseUrl();
+
+    @DefaultValue("false")
```
amitaekbote, line 26: this might need to be true for CI unit tests to pass

amitaekbote (approved):

PR #1028, opened 2021-05-04
Align ContainerResources signedIpAddressAllocations field/getter name

tbak:


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/ContainerResources.java
```diff
@@ -29,15 +29,16 @@
...
 
 import static com.netflix.titus.common.util.CollectionsExt.nonNull;
 
 /**
+ *
  */
 @ClassInvariant.List({
-        // TODO: This should be enabled as a ClassInvariant rather than a {@link ExtendedJobSanitizer} with TITUS-5370.
-        // @ClassInvariant(expr = "@asserts.matchingEbsAndIpZones(ebsVolumes, ipSignedAddressAllocations)", mode = VerifierMode.Strict),
+        @ClassInvariant(expr = "@asserts.matchingEbsAndIpZones(ebsVolumes, signedIpAddressAllocations)", mode = VerifierMode.Strict),
```
andrew-leung, line 41: I think we should leave this commented out for now. Since this will get also get checked during Gateway model validation, a change is needed to matchingEbsAndIpZones() to allow it to handle EbsVolume objects before TJC sanitization.

PR #1029, opened 2021-05-04
Add resourcePool to ApplicationSLA

joshi-keyur:
Add resourcePool to ApplicationSLA

- Set default value for resourcePool to reserved/elastic for non-GPU critical/flex tier capacity groups when scheduler is set to kubeScheduler
- In additional to creating a mechanism to separate capacity management at ApplicationSLA (Capacity Group)m, this change enables GPU capacity management where there may be multiple pools of GPU instances

### DIFF
--- a/titus-server-master/src/test/java/com/netflix/titus/master/endpoint/v2/rest/ApplicationSlaManagementResourceTest.java
+++ b/titus-server-master/src/test/java/com/netflix/titus/master/endpoint/v2/rest/ApplicationSlaManagementResourceTest.java
@@ -28,9 +28,9 @@ import com.netflix.titus.api.model.ApplicationSLA;
 import com.netflix.titus.common.util.archaius2.Archaius2Ext;
 import com.netflix.titus.master.config.MasterConfiguration;
 import com.netflix.titus.master.service.management.ApplicationSlaManagementService;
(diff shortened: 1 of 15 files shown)


titus-server-master/src/test/java/com/netflix/titus/master/endpoint/v2/rest/ApplicationSlaManagementResourceTest.java
```diff
@@ -147,13 +147,13 @@ public void getAllApplications() {
 
     @Test
     public void getApplicationSlaBySchedulerName() {
-        when(capacityManagementService.getApplicationSLAsForScheduler("kubescheduler")).thenReturn(Arrays.asList(SAMPLE_SLA_MANAGED_BY_KUBESCHEDULER));
```
tbak, line 150: Is there a constant for that?

tbak (approved):

titus-server-master/src/test/java/com/netflix/titus/master/endpoint/v2/rest/ApplicationSlaManagementResourceTest.java
joshi-keyur: I will create a constant in the next PR.

PR #1037, opened 2021-05-13
ManyReconciler API extension

tbak:


titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/ReconcilerActionProviderPolicy.java
```diff
@@ -0,0 +1,150 @@
...
+     */
+    private final int priority;
+
+    /**
+     * Evaluation frequency. It is guaranteed the time between a subsequent provider evaluations is at least the
+     * configured interval value. But the actual time might be longer if a processing takes a lot of time or there
+     * are other action providers with a higher priority. The time interval is measured between the last action
+     * completion time and the next action start time.
+     */
+    private final Duration executionInterval;
```
andrew-leung, line 53: Should this be suffixed with `Ms` for millis?

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/ReconcilerActionProviderPolicy.java
```diff
@@ -0,0 +1,150 @@
...
+            .withExecutionInterval(Duration.ZERO)
+            .build();
+
+    /**
+     * Execution priority. The lower the value, the higher the priority.
+     */
+    private final int priority;
+
+    /**
+     * Evaluation frequency. It is guaranteed the time between a subsequent provider evaluations is at least the
```
andrew-leung, line 48: I'm not sure I follow exactly what the `executionInterval`  is supposed to be doing. Is it saying there should be a delay of `executionInterval` between evaluations for a specific actions provider so that it doesn't create too many actions? If so, it seems like this would potentially create delay in cases where there isn't pressure and we could process these actions faster.

andrew-leung (approved):

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/ReconcilerActionProviderPolicy.java
tbak: This is `Duration` type, so we do not need this suffix.

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/ReconcilerActionProviderPolicy.java
tbak: The primary motivation is that sometimes there are evaluation rules that are expensive to compute and running them too tightly is too expensive. One is example is a full reconciliation cycle that should be run every so often (let's say 60sec). With this API you would configure two action providers: one for slow reconciliation cycle and one for the regular processing with no delay.

amit-git (commented):
LGTM

PR #1039, opened 2021-05-19
Better event processing in ManyReconciler

tbak:


titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/internal/EventDistributor.java
```diff
@@ -0,0 +1,270 @@
...
+                        fromQueue.getData(),
+                        fromQueue.getTransactionId()
+                ));
+            }
+        }
+        return mergeResult;
+    }
+
+    private void processEvents(List<SimpleReconcilerEvent<DATA>> events) {
+        events.forEach(event -> {
```
andrew-leung, line 205: I'm a little confused about why we iterate over each individual event here. If `events` has 5 `SimpleReconcilerEvent`s won't we add this same `events` to the `holder.onNext(events)` 5 times?

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/internal/EventDistributor.java
```diff
@@ -0,0 +1,270 @@
...
+            return cancelled;
+        }
+
+        private boolean onNext(List<SimpleReconcilerEvent<DATA>> event) {
+            if (cancelled) {
+                return false;
+            }
+
+            try {
+                emitter.next(event);
```
andrew-leung, line 252: If one of the `activeSink`/subscribers is slow would we expect the `emitter.next(event)` to be impacted? I suppose at least the reconciler `doProcess()` loop is decoupled from this so won't be impacted.

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/internal/EventDistributor.java
```diff
@@ -0,0 +1,270 @@
...
+                    }
+                }
+            } finally {
+                eventLoopThread = null;
+            }
+        }
+    }
+
+    void connectSink(FluxSink<List<SimpleReconcilerEvent<DATA>>> sink) {
+        String id = UUID.randomUUID().toString();
```
andrew-leung, line 108: It would be nice if we had a way to add some caller details here for better operational visibility. Then we'd know which caller stream may be having issues. Though I don't see how to do that without plumbing context down through many methods.

andrew-leung (approved):

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/internal/EventDistributor.java
tbak: Copy-paste. This class is based on the old reconciler code. Thanks for catching this

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/internal/EventDistributor.java
tbak: Yes, slow consumers will have an impact here. That is why all slow consumers should subscribe on their own thread.

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/internal/EventDistributor.java
tbak: Yes, I wanted to do exactly that, but could not find any reasonable way to pass this information. We could make it explicit in the `Flux<List<SimpleReconcilerEvent<DATA>>> changes()` method signature (or have a second method with consumer name parameter).

PR #1040, opened 2021-05-24
Do not use Java Stream API in JobSnapshot

tbak:


titus-client/src/main/java/com/netflix/titus/runtime/connector/jobmanager/JobSnapshot.java
```diff
@@ -209,7 +207,10 @@ public JobSnapshot build() {
             List<Task> allTasks = new ArrayList<>();
             tasksByJobId.values().forEach(allTasks::addAll);
 
-            Map<String, Task> taskById = allTasks.stream().collect(Collectors.toMap(Task::getId, Function.identity()));
+            Map<String, Task> taskById = new HashMap<>();
+            for (Task task : allTasks) {
```
andrew-leung, line 211: Could we add to the map in the `tasksByJobId.values().forEach()` line above it?

andrew-leung (approved):

PR #1044, opened 2021-06-01
tolerate decommissioning nodes by default

fabiokung:
unless jobs have the `activehost` hard constraint


alekar (approved):
Toleration looks right.

joshi-keyur (approved):

corindwyer (approved):

PR #1051, opened 2021-06-15
Set resourcePool task attribute and default ApplicationSLA schedulerName to kubeScheduler

joshi-keyur:


corindwyer (approved):

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/service/ServiceDifferenceResolver.java
```diff
@@ -285,8 +285,12 @@ public ServiceDifferenceResolver(
 
         JobDescriptor jobDescriptor = refJobView.getJob().getJobDescriptor();
         ApplicationSLA capacityGroupDescriptor = JobManagerUtil.getCapacityGroupDescriptor(jobDescriptor, capacityGroupService);
+        String resourcePool = capacityGroupDescriptor.getResourcePool();
```
tbak, line 288: Why if this is null value? Or it is always set to something?

PR #1056, opened 2021-06-22
Bump titus api definitions to latest

solarkennedy:
This is just step 0 to be able to use the new `NetworkMode` part of the api.


amitaekbote (approved):
looks good! builds with dependencies works on my dev setup.

PR #1057, opened 2021-06-25
Refactor remote federation / fallback

gabrielhartmann:
### Commit message
This simplifies introduction of remote federation, avoiding
the necessity of custom modification of every method in the
AggregatingJobServiceGateway.

### Description of the Change
The motivation for this change was the start of work for adding UpdateJobCapacity to fedv2.  It became clear that modifying the behavior of each API call in the AggregatingJobServiceGateway was going to get messy.  The approach presented here seems cleaner to me.

The FallBackJobServiceGateway has a primary and a secondary: 
```
primary:   RemoteJobServiceGateway      (aka fedv2)
secondary: AggregatingJobServiceGateway (aka fedv1)
```

If remote federation is enabled then the primary is attempted for all JobServiceGateway calls.  If an error occurs, fallback to the secondary occurs.  If remote federation is disabled the FallbackJobServiceGateway does not attempt the primary and calls the secondary in all cases.  So when remote federation is enabled the FallbackJobServiceGateway attempts to use fedv2 and falls back to fedv1 when an error is encountered.  When remote federation is disabled all calls are sent directly to fedv1.

Metrics look like this on fallback depending on GRPC status codes:
```
titus.fallbackJobServiceGateway.fallbackCount.createJob UNAVAILABLE
titus.fallbackJobServiceGateway.fallbackCount.findJobs UNIMPLEMENTED
```


gabrielhartmann:
I've tested this all on my dev stack.  I'll look into the unit test failures and fix the dependency injection stuff its complaining about.

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
```diff
@@ -1,64 +1,254 @@
...
-import static com.netflix.titus.runtime.endpoint.common.grpc.GrpcUtil.createRequestObservable;
-import static com.netflix.titus.runtime.endpoint.common.grpc.GrpcUtil.createWrappedStub;
+import javax.inject.Inject;
 
-public class FallbackJobServiceGateway {
-    private final JobManagementServiceGrpc.JobManagementServiceStub primaryClient;
-    private final GrpcConfiguration grpcConfiguration;
+public class FallbackJobServiceGateway implements JobServiceGateway {
+    public static String METRIC_ROOT = "titus.fallbackJobServiceGateway.";
+    private static String METRIC_FALLBACK_ROOT= METRIC_ROOT + "fallbackCount.";
```
fabiokung, line 38: nit: constants are `final`

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
```diff
@@ -1,64 +1,254 @@
...
+            });
+        } else {
+            return secondary;
+        }
+    }
+
+    private Id getMetricId(String metricName, Throwable t) {
+        String reason = t.getClass().getSimpleName();
+
+        if (t instanceof StatusRuntimeException) {
```
fabiokung, line 246: I recommend using `Status.fromThrowable(t).getCode()` instead

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
```diff
@@ -1,64 +1,254 @@
...
+        return getFallbackObservable(methodName, primaryObservable, secondaryObservable);
+    }
+
+    @Override
+    public Observable<JobChangeNotification> observeJobs(ObserveJobsQuery query, CallMetadata callMetadata) {
+        final String methodName = "observeJobs";
+        Observable<JobChangeNotification> primaryObservable = primary.observeJobs(query, callMetadata);
+        Observable<JobChangeNotification> secondaryObservable = secondary.observeJobs(query, callMetadata);
+        return getFallbackObservable(methodName, primaryObservable, secondaryObservable);
+    }
```
fabiokung, line 157: This strategy will not work well with stream APIs. If there is an error after the stream is established and some items were already sent, we want to propagate the error, not redo the stream on the secondaryObservables

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
```diff
@@ -1,64 +1,254 @@
...
+    public Completable moveTask(TaskMoveRequest taskMoveRequest, CallMetadata callMetadata) {
+        final String methodName = "moveTask";
+        Completable primaryCompletable = primary.moveTask(taskMoveRequest, callMetadata);
+        Completable secondaryObservable = secondary.moveTask(taskMoveRequest, callMetadata);
+        return getFallbackCompletable(methodName, primaryCompletable, secondaryObservable);
+    }
+
+    private <T> Observable<T> getFallbackObservable(String methodName, Observable<T> primary, Observable<T> secondary) {
+        if (federationConfiguration.isRemoteFederationEnabled()) {
+            return primary.onErrorResumeNext( (t) -> {
```
fabiokung, line 217: For operations with side effects, we run the risk of retrying un-retryable errors here, i.e. on partial failures where the remote service partially succeeded but the call returns an error - e.g. a timeout.

This is OK for a first iteration, but did you consider limiting what types of errors cause a fallback to `secondary`? Perhaps only `UNIMPLEMENTED` should trigger a fallback?

fabiokung (changes requested):
LGTM overall, but I'd leave the streaming APIs alone, or you need some more smartness there to know to fallback only when the `primary` stream wasn't estabilished

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
gabrielhartmann: This is a good idea `UNIMPLEMENTED` and `UNAVAILABLE` would probably be good choices.  Let me think about it a little.

gabrielhartmann:
@fabiokung Thanks for the review.  I believe I've addressed all your comments.  Are there other streaming APIs apart from `observeJob` and `observeJobs` which I should also avoid falling back on?

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
```diff
@@ -1,64 +1,272 @@
...
+    public Observable<JobQueryResult> findJobs(JobQuery jobQuery, CallMetadata callMetadata) {
+        final String methodName = "findJobs";
+        Observable<JobQueryResult> primaryObservable = primary.findJobs(jobQuery, callMetadata);
+        Observable<JobQueryResult> secondaryObservable = secondary.findJobs(jobQuery, callMetadata);
+        return getFallbackObservable(methodName, primaryObservable, secondaryObservable);
+    }
+
+    @Override
+    public Observable<JobChangeNotification> observeJob(String jobId, CallMetadata callMetadata) {
+        // TODO: Direct stream APIs to the remote (primary) JobServiceGateway.
```
fabiokung, line 148: fix the comment here? We are sending to secondary

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
```diff
@@ -1,64 +1,272 @@
...
+        // A failure in a stream which is in progress on the primary falling back to the secondary
+        // is not the right behavior.
+        return secondary.observeJob(jobId, callMetadata);
+    }
+
+    @Override
+    public Observable<JobChangeNotification> observeJobs(ObserveJobsQuery query, CallMetadata callMetadata) {
+        // TODO: Direct stream APIs to the remote (primary) JobServiceGateway.
```
fabiokung, line 156: same as above

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
fabiokung: sgtm, I'd caution against `UNAVAILABLE` as well, I'm not sure it could be thrown for in-flight requests. Perhaps I'm being a bit too paranoid.

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
```diff
@@ -1,64 +1,272 @@
...
+
+    private void incrementFallbackCounter(String methodName, Throwable t) {
+        Id metricId = this.getMetricId(METRIC_FALLBACK_ROOT + methodName, t);
+        this.titusRuntime.getRegistry().counter(metricId).increment();
+    }
+
+    private Id getMetricId(String metricName, Throwable t) {
+        String reason = t.getClass().getSimpleName();
+
+        if (t instanceof StatusRuntimeException) {
```
fabiokung, line 264: Remove this type check, `Status.fromThrowable()` already does it for you with the correct types. You might need to refactor this code a bit to recover the original message if `code == UNKNOWN`

fabiokung (approved):
LGTM with some minor nits

fabiokung:
> Are there other streaming APIs apart from `observeJob` and `observeJobs` which I should also avoid falling back on?

No, I believe these are the two gRPC streaming APIs we currently serve on the JobManagement service

titus-server-federation/src/main/java/com/netflix/titus/federation/service/FallbackJobServiceGateway.java
gabrielhartmann: The TODO is to direct to the primary instead of the secondary.

PR #1059, opened 2021-06-28
Ignore the string version of network_configuration

solarkennedy:
getNetworkModeName() is somehow leaking into the json serialization
of the NetworkConfiguration object, leading to:

```
"networkConfiguration":{"networkMode":0,"networkModeName":"UnknownNetworkMode"}
```

I don't want this, I only want the enum int in there.

Additionally, if we do make mistakes in the future with new properties,
I don't want the API to panic, so I added

    @JsonIgnoreProperties(ignoreUnknown = true)

Which seems to be common practice in this codebase.

I'm not sure how to test this end-to-end.


solarkennedy:
Confirmed on my dev stack that with this patch, the cassandra form of tasks has only the int:
```
...,"networkConfiguration":{"networkMode":0},
```

amitaekbote (approved):

PR #1066, opened 2021-07-19
Add last update timestamp to Job/Task objects and GRPC event notifications

tbak:


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/Job.java
```diff
@@ -45,11 +46,15 @@
...
 
-    public Job(String id, JobDescriptor<E> jobDescriptor, JobStatus status, List<JobStatus> statusHistory) {
+    private final Version version;
+
+    public Job(String id, JobDescriptor<E> jobDescriptor, JobStatus status, List<JobStatus> statusHistory, Version version) {
         this.id = id;
         this.jobDescriptor = jobDescriptor;
         this.status = status;
         this.statusHistory = CollectionsExt.nullableImmutableCopyOf(statusHistory);
+        // TODO Remove me. Version is a newly introduced field. If not set explicitly, assign a default value
```
andrew-leung, line 56: Just to clarify, the "remove me" is for the handling of null values not the version field itself?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/Version.java
```diff
@@ -0,0 +1,89 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.api.jobmanager.model.job;
+
+import java.util.Objects;
+
+/**
+ * TODO Add monotonically increasing `revision` number unique within a cell
```
andrew-leung, line 22: Would this revision number replace the timestamp or be in addition to? Seems like they would be very similar.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/VersionSuppliers.java
```diff
@@ -0,0 +1,54 @@
...
+
+        private final Clock clock;
+
+        private SimpleVersionSupplier(Clock clock) {
+            this.clock = clock;
+        }
+
+        @Override
+        public Version nextVersion() {
+            return Version.newBuilder().withTimestamp(clock.wallTime()).build();
```
andrew-leung, line 51: Currently if this gets called within the same millisecond is there any issue/side effect (e.g., 2 objects with the same version)?

andrew-leung (approved):

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/Job.java
tbak: Yes, so we do not accidentally create an object that has null Version value.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/Version.java
tbak: It would be in addition to it. Timestamp has a meaning and allows us to measure processing latencies. Revision id is needed for strict ordering and implementation of the consistent caching in gateway.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/VersionSuppliers.java
tbak: No, as we only use this to measure time, not order events/objects. Ordering will be supported when we introduce revision id.

PR #1067, opened 2021-07-20
Record job event stream propagation latencies

tbak:


titus-common/src/main/java/com/netflix/titus/common/util/event/EventPropagationUtil.java
```diff
@@ -0,0 +1,125 @@
...
+ * For events generated before the snapshot marker, we cannot use lastUpdateTimestamp as a starting point. These
+ * events will miss the TJC processing latency.
+ */
+public class EventPropagationUtil {
+
+    public static final String EVENT_ATTRIBUTE_PROPAGATION_STAGES = "event.propagation.stages";
+
+    public static String buildNextStage(String id, Map<String, String> attributes, long beforeTimestamp, long afterTimestamp) {
+        String stages = attributes.get(EVENT_ATTRIBUTE_PROPAGATION_STAGES);
+        long delaysMs = afterTimestamp - beforeTimestamp;
```
andrew-leung, line 53: It is possible this to be negative, correct? Are we then leaving the attribute as a negative value and letting `parseTrace()` adjust the metric to be >=0?

titus-client/src/main/java/com/netflix/titus/runtime/connector/jobmanager/JobEventPropagationUtil.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Help utils for measuring job event notification propagation delays.
+ */
+public class JobEventPropagationUtil {
+
+    public static final String CHECKPOINT_TJC_INTERNAL = "tjcInternal";
+    public static final String CHECKPOINT_GATEWAY_CLIENT = "gwClient";
+    public static final String CHECKPOINT_GATEWAY_INTERNAL = "gwInternal";
+    public static final String CHECKPOINT_FED_CLIENT = "fedClient";
+    public static final String CHECKPOINT_FED_INTERNAL = "federationInternal";
```
andrew-leung, line 40: I don't see this being used at all. Is this only intended for cases where more was being done in the Federation?

titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
```diff
@@ -493,10 +499,12 @@ private JobChangeNotification addStackName(JobChangeNotification notification) {
         switch (notification.getNotificationCase()) {
             case JOBUPDATE:
                 Job job = addStackName(notification.getJobUpdate().getJob());
+                job = JobEventPropagationUtil.recordChannelLatency(CHECKPOINT_FED_CLIENT, job, notification.getTimestamp(), titusRuntime.getClock());
```
andrew-leung, line 502: Presumably this is something we should add to FedV2 as well so that we don't lose it later in the quarter when we switch to that?

andrew-leung (approved):

titus-common/src/main/java/com/netflix/titus/common/util/event/EventPropagationUtil.java
tbak: Yes, this is just for the visual inspection. I think it makes sense to see that there is a time shift between servers.

titus-client/src/main/java/com/netflix/titus/runtime/connector/jobmanager/JobEventPropagationUtil.java
tbak: It is used in `JobEventPropagationMetrics` to create a tag value. It is passed via `FEDERATION_LABELS`

titus-server-federation/src/main/java/com/netflix/titus/federation/service/AggregatingJobServiceGateway.java
tbak: Yes. Once we get this working and cleaned up, I would like to add it to FedV2 as well.

PR #1070, opened 2021-07-23
Enrich v0 pod objects with image,cmd,args

solarkennedy:
I'm trying to help remove the JobDescriptor object from our pods.

To do that, on those tools that need the jobdescriptor, we can fake
it by sourcing other pieces of data from the pod object itself.

For https://github.com/Netflix-Skunkworks/titus-isolate/pull/341
specifically, in this PR, I've added in Image, Cmd, Args into the
pod object directly instead of the cinfo placeholders.

Sure, this data isn't actually consumed by VK (on v0 pods), but it
is useful for other surrounding tools, annd heck just running kubectl
to see what the heck is going on?


solarkennedy:
Closing because the v1 pod migration will happen soon, and instead of populating both here, I'll adjust the applications that use jobdescriptor and have them account for both v0 pods and v1 pods.

PR #1072, opened 2021-07-28
Validator for AutoScalingPolicy model object

joshi-keyur:
### Description of the Change

Basic check to prevent invalid input with null AlarmConfiguration

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/appscale/model/PolicyConfiguration.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/appscale/model/PolicyConfiguration.java
@@ -16,8 +16,6 @@
 
 package com.netflix.titus.api.appscale.model;
 
-import com.google.common.base.Preconditions;
-
 public class PolicyConfiguration {
 
     private final String name;
@@ -106,7 +104,6 @@ public class PolicyConfiguration {
         }
 
         public PolicyConfiguration build() {
-            Preconditions.checkNotNull(alarmConfiguration, "AlarmConfiguration is null");
             return new PolicyConfiguration(name, policyType, stepScalingPolicyConfiguration, alarmConfiguration, targetTrackingPolicy);
         }
     }
--- a/titus-api/src/main/java/com/netflix/titus/api/appscale/model/PolicyConfiguration.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/appscale/model/PolicyConfiguration.java
@@ -16,6 +16,8 @@
 
 package com.netflix.titus.api.appscale.model;
 
+import com.google.common.base.Preconditions;
+
 public class PolicyConfiguration {
 
     private final String name;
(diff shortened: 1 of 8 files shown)


titus-api/src/main/java/com/netflix/titus/api/appscale/model/PolicyConfiguration.java
```diff
@@ -104,6 +106,7 @@ public Builder withTargetTrackingPolicy(TargetTrackingPolicy targetTrackingPolic
         }
 
         public PolicyConfiguration build() {
```
andrew-leung, line 108: Should we also check that both step and target aren't null (one, and only one) must be set too.

titus-api/src/main/java/com/netflix/titus/api/appscale/model/PolicyConfiguration.java
```diff
@@ -104,6 +106,7 @@ public Builder withTargetTrackingPolicy(TargetTrackingPolicy targetTrackingPolic
...
+            Preconditions.checkNotNull(alarmConfiguration, "AlarmConfiguration is null");
```
andrew-leung, line 109: Does this result in a proper INVALID_ARGUMENT return type to users? It would be good to know that this doesn't give some obscure error.

andrew-leung: Also, are we still planning on adding Java Bean validation or just rely on this?

andrew-leung (approved):

tbak (approved):

titus-api/src/main/java/com/netflix/titus/api/appscale/model/PolicyConfiguration.java
joshi-keyur: This was a quick fix to prevent persisting an invalid value and cause GetAllScalingPolicies API to fail.
Since we have temporary remediation steps, I can look into adding some more validation and tests to confirm the new behavior.

titus-api/src/main/java/com/netflix/titus/api/appscale/model/PolicyConfiguration.java
joshi-keyur: Sounds good.

tbak (approved):

corindwyer (approved):

titus-api/src/main/java/com/netflix/titus/api/appscale/model/sanitizer/ScalingPolicyAssertions.java
```diff
@@ -0,0 +1,47 @@
...
+import com.netflix.titus.api.appscale.model.PolicyConfiguration;
+
+public class ScalingPolicyAssertions {
+
+    public Map<String, String> validateScalingPolicy(AutoScalingPolicy autoScalingPolicy) {
+        if (autoScalingPolicy == null || autoScalingPolicy.getPolicyConfiguration() == null) {
+            return Collections.emptyMap();
+        }
+        PolicyConfiguration policyConfiguration = autoScalingPolicy.getPolicyConfiguration();
+        if (policyConfiguration == null) {
```
andrew-leung, line 33: Isn't this the same check as above?

joshi-keyur: aha! Thanks

PR #1080, opened 2021-08-06
Hide Kube Java lib invocations behind internal API

tbak:
Hide Kube Java lib invocations behind internal API. We need a proper interface between Kube and Titus so we can easily inject stubbed version of Kube Cluster in the integration tests.

### DIFF
--- a/titus-common-runtime/src/main/java/com/netflix/titus/runtime/connector/kubernetes/KubeApiFacade.java
+++ b/titus-common-runtime/src/main/java/com/netflix/titus/runtime/connector/kubernetes/KubeApiFacade.java
@@ -18,7 +18,6 @@ package com.netflix.titus.runtime.connector.kubernetes;
 
 import com.netflix.titus.runtime.connector.kubernetes.v1.V1OpportunisticResource;
 import io.kubernetes.client.informer.SharedIndexInformer;
-import io.kubernetes.client.openapi.ApiException;
 import io.kubernetes.client.openapi.models.V1Node;
 import io.kubernetes.client.openapi.models.V1PersistentVolume;
 import io.kubernetes.client.openapi.models.V1PersistentVolumeClaim;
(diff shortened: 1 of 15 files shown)


titus-common-runtime/src/main/java/com/netflix/titus/runtime/connector/kubernetes/KubeApiFacade.java
```diff
@@ -18,26 +18,49 @@
...
+ */
 public interface KubeApiFacade {
 
-    ApiClient getApiClient();
-
-    CoreV1Api getCoreV1Api();
+    // Nodes
 
-    CustomObjectsApi getCustomObjectsApi();
+    void deleteNode(String nodeName) throws ApiException;
```
corindwyer, line 36: Maybe we shouldn't leak the underlying client's ApiException? Thoughts?

corindwyer (approved):

titus-common-runtime/src/main/java/com/netflix/titus/runtime/connector/kubernetes/KubeApiFacade.java
tbak: Good idea. I will create a wrapper

corindwyer (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PersistentVolumeClaimGcController.java
```diff
@@ -107,20 +106,11 @@ private boolean gcPersistentVolumeClaim(V1PersistentVolumeClaim pvc) {
...
-                    null,
-                    null
-            );
+            kubeApiFacade.deleteNamespacedPersistentVolumeClaim(DEFAULT_NAMESPACE, volumeClaimName);
             logger.info("Successfully deleted persistent volume claim {}", formatPvcEssentials(pvc));
             return true;
-        } catch (ApiException e) {
-            if (!e.getMessage().equalsIgnoreCase(NOT_FOUND)) {
+        } catch (KubeApiException e) {
+            if (e.getErrorCode() != KubeApiException.ErrorCode.NOT_FOUND) {
```
fabiokung, line 113: if I read this code correctly, shouldn't this be an `==` comparison? If I'm right, this bug was already present before your change.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PersistentVolumeUnassociatedGcController.java
```diff
@@ -113,19 +112,11 @@ private boolean gcPersistentVolume(V1PersistentVolume pv) {
...
-                    null,
-                    null
-            );
+            kubeApiFacade.deletePersistentVolume(volumeName);
             logger.info("Successfully deleted persistent volume {}", formatPvEssentials(pv));
             return true;
-        } catch (ApiException e) {
-            if (!e.getMessage().equalsIgnoreCase(NOT_FOUND)) {
+        } catch (KubeApiException e) {
+            if (e.getErrorCode() != KubeApiException.ErrorCode.NOT_FOUND) {
```
fabiokung, line 119: same from before, this looks like it should be an `==`

fabiokung (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/controller/PersistentVolumeClaimGcController.java
tbak: Good catch. I kept the original logic which was also incorrect.

PR #1084, opened 2021-08-17
Fix shard id generator (hashcode may be negative)

tbak:


amit-git (approved):

andrew-leung (approved):

titus-common/src/main/java/com/netflix/titus/common/framework/simplereconciler/ManyReconciler.java
```diff
@@ -190,7 +190,7 @@ private Builder() {
                     ? CloseableReference.referenceOf(Schedulers.newSingle("reconciler-notification-" + name, true), Scheduler::dispose)
                     : CloseableReference.referenceOf(notificationScheduler);
 
-            Function<String, Integer> shardIndexSupplier = id -> id.hashCode() % shardCount;
+            Function<String, Integer> shardIndexSupplier = id -> Math.abs(id.hashCode()) % shardCount;
```
andrew-leung, line 193: I assume we elsewhere validate that shardCount is > 0?

PR #1086, opened 2021-08-18
Only set the ipv4 container attribute if we have one

solarkennedy:
We are moving to a new world where IPv6 is the main thing,
and IPv4 is extra.

This change makes the ipv4/ipv6 task attribute setter "NetworkMode"-aware,
and only sets the IPv4 when it makes sense.

Only a small number of experimental tasks are using this mode,
so we need to use those as canaries to see what else will break
when we stop publishing a v4 address in this way.

----

This change will also be paired with a titus-executor change that add networkMode
to the details that come back for the control-plane to use.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/TaskAttributes.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/TaskAttributes.java
@@ -58,6 +58,7 @@ public final class TaskAttributes {
     public static final String TASK_ATTRIBUTES_CONTAINER_IP = "task.containerIp";
     public static final String TASK_ATTRIBUTES_CONTAINER_IPV4 = "task.containerIPv4";
     public static final String TASK_ATTRIBUTES_CONTAINER_IPV6 = "task.containerIPv6";
+    public static final String TASK_ATTRIBUTES_TRANSITION_IPV4 = "task.transitionIPv4";
     public static final String TASK_ATTRIBUTES_NETWORK_INTERFACE_ID = "task.networkInterfaceId";
     public static final String TASK_ATTRIBUTES_NETWORK_INTERFACE_INDEX = "task.networkInterfaceIndex";
     public static final String TASK_ATTRIBUTES_EXECUTOR_URI_OVERRIDE = "task.executorUriOverride";
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/TaskAttributes.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/TaskAttributes.java
@@ -58,6 +58,12 @@ public final class TaskAttributes {
     public static final String TASK_ATTRIBUTES_CONTAINER_IP = "task.containerIp";
     public static final String TASK_ATTRIBUTES_CONTAINER_IPV4 = "task.containerIPv4";
     public static final String TASK_ATTRIBUTES_CONTAINER_IPV6 = "task.containerIPv6";
+    /*
+     * TASK_ATTRIBUTES_TRANSITION_IPV4 is a special IP that represents the IP that
+     * Tasks in the special Ipv6AndIpv4Fallback network Mode use.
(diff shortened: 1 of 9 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
```diff
@@ -133,21 +133,31 @@ public static String getCapacityGroupDescriptorName(JobDescriptor<?> jobDescript
...
+                String ipv4 = networkConfiguration.getIpAddress();
+                String ipv6 = networkConfiguration.getIpV6Address();
+                String primaryIP = networkConfiguration.getPrimaryIpAddress();
                 Map<String, String> newContext = new HashMap<>(task.getTaskContext());
                 BiConsumer<String, String> contextSetter = (key, value) -> StringExt.applyIfNonEmpty(value, v -> newContext.put(key, v));
 
-                contextSetter.accept(TaskAttributes.TASK_ATTRIBUTES_CONTAINER_IP, networkConfiguration.getIpAddress());
-                contextSetter.accept(TaskAttributes.TASK_ATTRIBUTES_CONTAINER_IPV4, networkConfiguration.getIpAddress());
-                Evaluators.acceptNotNull(networkConfiguration.getIpV6Address(), ipv6 -> newContext.put(TaskAttributes.TASK_ATTRIBUTES_CONTAINER_IPV6, ipv6));
+                if (networkConfiguration.getNetworkMode() == "Ipv6AndIpv4Fallback") {
```
tbak, line 151: `equals`?
Use constant for `Ipv6AndIpv4Fallback`?

tbak (changes requested):

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/JobManagerUtil.java
solarkennedy: Got it.
I think I've imported the constant correctly, please double check that importing the enum like I have is appropriate.

titus-server-master/src/main/java/com/netflix/titus/master/mesos/TitusExecutorDetails.java
```diff
@@ -165,9 +175,24 @@ public String toString() {
...
+                    ", networkMode='" + networkMode + '\'' +
                     ", eniID='" + eniID + '\'' +
                     ", resourceID='" + resourceID + '\'' +
                     '}';
         }
+
+        public String getPrimaryIpAddress() {
+            // For backward-compatibility reasons, the "primary ip" is considered the v4 address
+            // if we have one (if we are not in the transition mode), otherwise ipv6
+            if (ipAddress != null && !ipAddress.equals("")) {
```
corindwyer, line 187: you can use `StringExt.isNotEmpty` to make these kind of statements cleaner

titus-api/src/main/java/com/netflix/titus/api/jobmanager/TaskAttributes.java
```diff
@@ -58,6 +58,7 @@
     public static final String TASK_ATTRIBUTES_CONTAINER_IP = "task.containerIp";
     public static final String TASK_ATTRIBUTES_CONTAINER_IPV4 = "task.containerIPv4";
     public static final String TASK_ATTRIBUTES_CONTAINER_IPV6 = "task.containerIPv6";
+    public static final String TASK_ATTRIBUTES_TRANSITION_IPV4 = "task.transitionIPv4";
```
corindwyer, line 61: Can you add a javadoc for what this is? The others are self explanatory but this one could be interpreted in different ways.

corindwyer (approved):

tbak (approved):

solarkennedy:
Now, I'm not sure if this *must* be merged before I get https://github.com/Netflix/titus-executor/pull/714 in.
I don't know who the code will behave if the network mode isn't present in the first place.

I'll know right away on staging I suppose.

PR #1092, opened 2021-08-24
Improve a final reason message for tasks terminated by users

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/mesos/kubeapiserver/PodToTaskMapper.java
```diff
@@ -139,6 +140,9 @@ public PodToTaskMapper(PodWrapper podWrapper,
             reasonCode = resolveFinalTaskState(REASON_NORMAL);
         } else if (podWrapper.getPodPhase() == PodPhase.FAILED) {
             reasonCode = resolveFinalTaskState(REASON_FAILED);
+            if(taskState == KillInitiated) {
+                message = "Titus task terminated by a user";
```
solarkennedy, line 144: Now that I see this code, I didn't realize that the update came *from* the pod status.
I don't think we should be overwriting what the original message was here on this line.

Let me make a different PR that actually sets that message to something more sane in the first place.

https://github.com/Netflix/titus-executor/pull/720

tbak:
Cancelling this as we will fix it in titus-executor

PR #1098, opened 2021-09-01
Binpack self-managed tasks to optimize their relocation

fabiokung:
The new behavior can be disabled by the `titus.feature.relocationBinpackingEnabled` flag.

Binpacking is accomplished by:

* a new label added to tasks that should opt-into binpacking (currently only tasks from jobs with SelfManaged disruption policies)
* Pod affinity and anti-affinity rules against the label above

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/affinity/DefaultPodAffinityFactory.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/affinity/DefaultPodAffinityFactory.java
@@ -25,10 +25,12 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import com.netflix.titus.api.FeatureActivationConfiguration;
 import com.netflix.titus.api.jobmanager.JobConstraints;
 import com.netflix.titus.api.jobmanager.model.job.Job;
 import com.netflix.titus.api.jobmanager.model.job.JobFunctions;
 import com.netflix.titus.api.jobmanager.model.job.Task;
+import com.netflix.titus.api.jobmanager.model.job.disruptionbudget.SelfManagedDisruptionBudgetPolicy;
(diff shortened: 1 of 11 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/affinity/DefaultPodAffinityFactory.java
```diff
@@ -285,13 +293,52 @@ private void processZoneConstraints() {
...
+                            .weight(RELOCATION_AFFINITY_WEIGHT)
+                            .podAffinityTerm(new V1PodAffinityTerm()
+                                    .labelSelector(new V1LabelSelector()
+                                            .addMatchExpressionsItem(new V1LabelSelectorRequirement()
+                                                    .key(KubeConstants.POD_LABEL_RELOCATION_BINPACK)
+                                                    .operator(KubeConstants.SELECTOR_OPERATOR_EXISTS)
+                                            )
+                                    )
+                                    .topologyKey(KubeConstants.NODE_LABEL_MACHINE_ID)
+                            )
```
tbak, line 311: This does not distinguish between different jobs, does it?
We must ensure that job spreading is disabled. This is controlled independently in `DefaultTopologyFactory`. Maybe we should change that implementation as well and check `isRelocationBinpackingEnabled` there.
Or the goal here is to limit number of machines used, but preserve the original spreading behavior?

tbak (approved):

corindwyer (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/affinity/DefaultPodAffinityFactory.java
fabiokung: I wanted to preserve the original spreading behavior, but favor binpacking when possible (when multiple choices given the current spreading behavior is maintained). Let me at least review the weights in both cases

PR #1102, opened 2021-09-06
code hygiene: remove unused code

fabiokung:
Removed:

* granular kube-scheduler assignment (routing) configuration: only integration tests are relying on assigning tasks to Fenzo
* job activity history publisher configuration


titus-api/src/main/java/com/netflix/titus/api/FeatureRolloutPlans.java
```diff
@@ -63,25 +63,11 @@
     )
     String JOB_AUTHORIZATION_FEATURE = "jobAuthorizationFeature";
 
-    @FeatureRollout(
```
corindwyer, line 66: I didn't see a removal of this in the code. Was it ever used?

corindwyer (approved):

tbak (approved):

titus-api/src/main/java/com/netflix/titus/api/FeatureRolloutPlans.java
fabiokung: I think the use got removed on a previous PR that removed other JAH pieces

PR #1107, opened 2021-09-10
Expanded the error message on kubeapi failures

solarkennedy:
For Titus developers, and heck for titus users, it is
a better experience for all when we get the right error message
we need when it is "low level". The types of errors one gets
when k8s throws a non-200 at us are hard to get at currently.

This change makes it so we get the response body with it.
Per https://github.com/kubernetes-client/java/issues/695#issuecomment-530139220

Before:

```
Failed to create pod: (IllegalStateException) Unable to launch a task 6325b498-3132-496e-969b-ae6308b43871 -CAUSED BY-> (KubeApiException) Bad Request -CAUSED BY-> (ApiException) Bad Request
```

After:
```
Failed to create pod: (IllegalStateException) Unable to launch a task 8cdf15d2-8b79-4c09-a8c7-9d7544a6bcbe -CAUSED BY-> (KubeApiException) Bad Request: httpStatus=400, body={"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"admission webhook \"agent-webhook.netflix.svc\" denied the request: failed to apply mutator for sidecar \"sleepy\" on channel \"default\": unable to process add_containers: add_containers.0.env.0.value: undefined field: \"sleep-time\"","code":400} -CAUSED BY-> (ApiException) Bad Request
```

### DIFF
--- a/titus-common-runtime/src/main/java/com/netflix/titus/runtime/connector/kubernetes/KubeApiException.java
+++ b/titus-common-runtime/src/main/java/com/netflix/titus/runtime/connector/kubernetes/KubeApiException.java
@@ -36,7 +36,7 @@ public class KubeApiException extends RuntimeException {
     }
 
     public KubeApiException(ApiException cause) {
-        this(cause.getMessage(), cause);
+        this(String.format("%s: httpStatus=%s, body=%s", cause.getMessage(), cause.getCode(), cause.getResponseBody()), cause);
     }
 
     public ErrorCode getErrorCode() {


titus-common-runtime/src/main/java/com/netflix/titus/runtime/connector/kubernetes/KubeApiException.java
```diff
@@ -36,7 +36,7 @@ public KubeApiException(String message, Throwable cause) {
     }
 
     public KubeApiException(ApiException cause) {
-        this(cause.getMessage(), cause);
+        this(String.format("%s: httpStatus=%s, body=%s", cause.getMessage(), cause.getCode(), cause.getResponseBody()), cause);
```
slushie, line 39: Nice! 🥳 Does the Kube API always return `metav1.Status` in this context? It'd be especially slick if we could parse the `Message` field from [this struct](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L650-L680). 😎

solarkennedy: We are talking http here I  think, so I don't think there is a guarantee we have that (could be just plain text that says "503" or whatever?)

We *could* do a try/except, and if we couldn't read the status, then we just print the full json.
But I think printing the full json is probably ok in this case

slushie: Yep, good point. LGTM :+1:

corindwyer (approved):

tbak (approved):

PR #1108, opened 2021-09-11
Implemented pod Volume support

solarkennedy:
This takes in the new Volumes field in the API and turns them into
real V1Pod Volumes on the pod object, ready for virtual-kubelet to
consume.

I've only implemented one kind of volume so far, the SharedContainerVolume,
which is the first one we will need to make it so containers can
share files and mounts (fuse sidecar, shared things in /etc, etc)

This only implements the Volume part, not the VolumeMount part (which
slots into the containers definitions). That will happen next.


titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/Volume.java
```diff
@@ -0,0 +1,78 @@
...
+package com.netflix.titus.api.jobmanager.model.job;
+
+import javax.validation.Valid;
+
+public class Volume {
+
+    @Valid
+    private final String name;
+
+    private final SharedContainerVolumeSource sharedContainerVolumeSource;
```
tbak, line 26: Is this value relevant for all volume types?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/store/mixin/VolumeMixin.java
```diff
@@ -0,0 +1,21 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.netflix.titus.api.jobmanager.model.job.SharedContainerVolumeSource;
+
+public abstract class VolumeMixin {
+    @JsonCreator
+    public VolumeMixin(
+            @JsonProperty("name")
```
tbak, line 12: Formatting?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/store/mixin/VolumeMixin.java
```diff
@@ -0,0 +1,21 @@
...
+                    String name,
+
+            @JsonInclude(Include.NON_NULL)
```
tbak, line 15: Why is this annotation here?

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -371,6 +374,24 @@ private static BasicContainer toCoreBasicContainer(com.netflix.titus.grpc.protog
                 .build();
     }
 
+    private static List<Volume> toCoreVolumes(List<com.netflix.titus.grpc.protogen.Volume> volumes) {
+        return volumes.stream().map(GrpcJobManagementModelConverters::toCoreVolume).collect(Collectors.toList());
```
tbak, line 378: We should avoid Java stream API usage, as it is very slow and produces a lot of garbage in the heap.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v0/V0SpecPodFactory.java
```diff
@@ -177,6 +182,29 @@ public V1Pod buildV1Pod(Job<?> job, Task task, boolean useKubeScheduler, boolean
         return new V1Pod().metadata(metadata).spec(spec);
     }
 
+    private List<V1Volume> buildV1Volumes(List<Volume> volumes) {
+        if (volumes == null) { return null;}
```
tbak, line 186: It would be safer to return `Collections.emtpyList()`

titus-api/src/main/java/com/netflix/titus/api/jobmanager/store/mixin/VolumeMixin.java
solarkennedy: Great question.

My reasoning here is that as we add all sorts of different configs for volumes for new types (ebs, efs, etc):
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#volume-v1-core

I didn't want to polute the json with all sorts of non-related configs.

So if figured for this case, we only want to serialize configs that are related to the volume in question (ie if they are non-null.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: Can you recommend a preferred pattern for `map` over an array like this?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v0/V0SpecPodFactory.java
solarkennedy: Oh right

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/Volume.java
solarkennedy: It is not, but I'm not sure how to make it optional except to just set it to `null` when not related.

Do you think this would be better implemented as.. inheritance?

tbak: Yes, we use inheritance in such cases. Please, have a look at the disruption budget model.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
tbak: We should use `for` loop in such case.

corindwyer:
Are you using an IDE for these updates? Formatting looks different than I would expect from `codequality/Intellij_code_style.xml`.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
```diff
@@ -272,6 +277,30 @@ private V1Container buildV1ExtraContainers(BasicContainer extraContainer) {
                 .imagePullPolicy(DEFAULT_IMAGE_PULL_POLICY);
     }
 
+
+    private List<V1Volume> buildV1Volumes(List<Volume> volumes) {
+        if (volumes == null) { return null;}
```
corindwyer, line 282: `Collections.emptyList`

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
```diff
@@ -272,6 +277,30 @@ private V1Container buildV1ExtraContainers(BasicContainer extraContainer) {
...
+    private V1Volume buildV1Volume(Volume volume) {
+        V1FlexVolumeSource flexVolume = getV1FlexVolume(volume);
+        return new V1Volume()
+                .name(volume.getName())
+                .flexVolume(flexVolume);
+    }
+
+    private V1FlexVolumeSource getV1FlexVolume(Volume volume) {
+        SharedContainerVolumeSource sharedContainerVolumeSource = volume.getSharedContainerVolumeSource();
+        if (sharedContainerVolumeSource == null) { return null; }
```
corindwyer, line 295: Do you know what happens if this returns null and it is set to null higher up?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/store/mixin/VolumeMixin.java
solarkennedy: I agree it looks weird, but doing the automatic reformatting seems to do nothing for the arguments to the function like this. 

I've manually tried to make it look more like the other files by putting the decorator right next to the argument name, and the reformatter still doesn't care.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: Ok easy. Let me know if you would like me to make a second PR on this file to replace all the stream() usage to for loops too.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
solarkennedy: Yes. On these Volume types, there a ton of BLASource fields. One for every kind of volume you can think of. 99% of them are null.

When these fields are null, all the storage drivers just know not to worry about them.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -362,7 +362,11 @@ public static Container toCoreContainer(com.netflix.titus.grpc.protogen.Containe
     }
 
     private static List<BasicContainer> toCoreBasicContainers(List<com.netflix.titus.grpc.protogen.BasicContainer> extraContainersList) {
-        return extraContainersList.stream().map(GrpcJobManagementModelConverters::toCoreBasicContainer).collect(Collectors.toList());
+        List<BasicContainer> basicContainers = Collections.emptyList();
```
corindwyer, line 365: IIRC, this won't work as `emptyList` is a specific implementation that must be empty. You can change it to `new ArrayList()`

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -375,7 +379,11 @@ private static BasicContainer toCoreBasicContainer(com.netflix.titus.grpc.protog
     }
 
     private static List<Volume> toCoreVolumes(List<com.netflix.titus.grpc.protogen.Volume> volumes) {
-        return volumes.stream().map(GrpcJobManagementModelConverters::toCoreVolume).collect(Collectors.toList());
+        List<Volume> coreVolumes = Collections.emptyList();
```
corindwyer, line 382: same as above

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: Ah you are right, they are immutable

solarkennedy:
> Are you using an IDE for these updates? Formatting looks different than I would expect from `codequality/Intellij_code_style.xml`.

I am, but I hadn't configured it to format-on-save. This is set now.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/Volume.java
solarkennedy: ty. I've done this, please give this extra scrutiny because this is my first time interacting with inheritance in java.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/Volume.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.titus.api.jobmanager.model.job.volume;
+
+import javax.validation.Valid;
+
+public class Volume {
+    @Valid
+    private static String name;
```
tbak, line 23: It should not be static.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/Volume.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public Builder toBuilder() {
+        return newBuilder();
+    }
+
+    public static Builder newBuilder() {
+        return new Builder(name);
+    }
+
+    public static class Builder {
```
tbak, line 41: To avoid casting when using this builder it should include template parameter.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -388,20 +389,27 @@ private static BasicContainer toCoreBasicContainer(com.netflix.titus.grpc.protog
     }
 
     private static Volume toCoreVolume(com.netflix.titus.grpc.protogen.Volume grpcVolume) {
-        return Volume.newBuilder()
-                .withName(grpcVolume.getName())
-                .withSharedContainerVolumeSource(toCoreSharedVolumeSource(grpcVolume.getSharedContainerVolumeSource()))
-                .build();
+        if (grpcVolume.getSharedContainerVolumeSource() != null) {
```
tbak, line 392: I think this should be oneof in protobuf model. I assume that each volume will be of exactly one type?

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -388,20 +389,27 @@ private static BasicContainer toCoreBasicContainer(com.netflix.titus.grpc.protog
...
+            return SharedContainerVolume.newBuilder()
+                    .withName(grpcVolume.getName())
+                    .withSharedContainerVolumeSource(toCoreSharedVolumeSource(grpcVolume.getSharedContainerVolumeSource()))
+                    .build();
+        } else {
+            // Currently the shared volume source is the only kind of volume we support
+            return null;
```
tbak, line 399: This is dangerous. We should either do invariant check and throw exception or return `Optional<Volume>`

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -500,29 +509,34 @@ public static Day toCoreDay(com.netflix.titus.grpc.protogen.Day grpcDay) {
...
+            (com.netflix.titus.grpc.protogen.ContainerHealthProvider grpcContainerHealthProvider) {
         return ContainerHealthProvider.newBuilder()
                 .withName(grpcContainerHealthProvider.getName())
                 .withAttributes(grpcContainerHealthProvider.getAttributesMap())
                 .build();
     }
 
-    public static JobDescriptor.JobDescriptorExt toCoreJobExtensions(com.netflix.titus.grpc.protogen.JobDescriptor grpcJobDescriptor) {
+    public static JobDescriptor.JobDescriptorExt toCoreJobExtensions(com.netflix.titus.grpc.protogen.JobDescriptor
+                                                                             grpcJobDescriptor) {
```
tbak, line 539: This formatting looks strange. Is it auto-formatter? We should use common template.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/SharedContainerVolume.java
```diff
@@ -0,0 +1,51 @@
+package com.netflix.titus.api.jobmanager.model.job.volume;
+
+import com.netflix.titus.api.jobmanager.model.job.SharedContainerVolumeSource;
+
+public class SharedContainerVolume extends Volume {
+
+    private final SharedContainerVolumeSource sharedContainerVolumeSource;
```
tbak, line 7: Isn't `SharedContainerVolumeSource` and `SharedContainerVolume` the same? Why not just `SharedContainerVolume` with fields from `SharedContainerVolumeSource`?

tbak (changes requested):

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: Added https://github.com/Netflix/titus-api-definitions/pull/150

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: Afaik I'm using the shared codestyle template.
Not sure, the autoformat doesn't seem to care about the line len or anything. I'll remove this noise.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/SharedContainerVolume.java
solarkennedy: Yes, you are right. I didn't really understand "What" we were inheriting till now.
Next pass will just be the SharedContainerVolumeSource with the right fields.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: Let me figure out how to do one of those. I agree I don't like having the null here, but on the other hand it doesn't sound too crazy to have the control plane tolerate (but just not understand) new volumes?

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/Volume.java
solarkennedy: whoops

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/Volume.java
solarkennedy: I don't understand the java type system to know exactly what those words mean :(
Let me know on the next review if it is still relevant

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v0/V0SpecPodFactory.java
```diff
@@ -177,6 +183,39 @@ public V1Pod buildV1Pod(Job<?> job, Task task, boolean useKubeScheduler, boolean
...
+
+    private V1Volume buildV1Volume(Volume volume) {
+        if (volume.getVolumeSource() instanceof SharedContainerVolumeSource) {
+            V1FlexVolumeSource flexVolume = getV1FlexVolumeForSharedContainerVolumeSource(volume);
+            return new V1Volume()
+                    .name(volume.getName())
+                    .flexVolume(flexVolume);
+        } else {
+            // SharedVolumeSource is currently the only supported volume type
+            return null;
```
tbak, line 205: I think this should never be null. Either we have `Volume` with valid data or the `Volume` list should be empty.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
```diff
@@ -272,6 +281,40 @@ private V1Container buildV1ExtraContainers(BasicContainer extraContainer) {
...
+
+    private V1Volume buildV1Volume(Volume volume) {
+        if (volume.getVolumeSource() instanceof SharedContainerVolumeSource) {
+            V1FlexVolumeSource flexVolume = getV1FlexVolumeForSharedContainerVolumeSource(volume);
+            return new V1Volume()
+                    .name(volume.getName())
+                    .flexVolume(flexVolume);
+        } else {
+            // SharedVolumeSource is currently the only supported volume type
+            return null;
```
tbak, line 304: I think this should never be null. Either we have Volume with valid data or the Volume list should be empty.

tbak (changes requested):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v0/V0SpecPodFactory.java
solarkennedy: I think I (finally) understand what you are saying. 
You are (rightfully) concerned we might accidentally have this in a pod spec:

```
volumes: [ {volume1}, null ],
```

I think the Optional thing sounds like a good choice here, and simply not return anything and then have the caller not add anything to `volumes` in that case.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/SharedContainerVolumeSource.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public String getSourceContainer() {
+        return sourceContainer;
+    }
+
+    public String getSourcePath() {
+        return sourcePath;
+    }
+
+    public SharedContainerVolumeSource(
```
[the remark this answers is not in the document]
solarkennedy: Added

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/Volume.java
solarkennedy: Ok I think I understand, but was this comment with regards to VolumeSource and not Volume? 
Because I think the VolumeSource stuff is where I have to do the casting.

solarkennedy:
Alright I think I've address the review comments (thank you @tbak).
Still not 100% sure I have the builder format correct.

corindwyer (approved):

solarkennedy:
Finally worked through some build issues with @joshi-keyur and got this deployed to my devstack. Added one more commit where I forgot to include VolumeSourceMixing where it couldn't deserialize correctly.

tbak (approved):

PR #1109, opened 2021-09-13
Upgrade netflixoss to fix CNAME issues

rpalcolea:
### Description of the Change

**FIXME**: please add at least one sentence explaining why changes here are being made.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ buildscript {
 }
 
 plugins {
-    id 'nebula.netflixoss' version '9.1.0'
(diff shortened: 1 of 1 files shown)


solarkennedy:
Cross linking (cc @tbak ) to https://github.com/Netflix/titus-control-plane/pull/1110

I don't think we are ready to upgrade this far, because it means upgrading so many other things.

Can we get by on #1110 instead and just point to the new url/cname?

solarkennedy:
We are archiving this repo.

PR #1114, opened 2021-09-16
Disable CircleCI

tbak:


.circleci/config.yml_disabled
```diff
@@ -1,3 +1,5 @@
+REMOVE ME
```
solarkennedy, line 1: Why not just `git rm` ?

solarkennedy (approved):

.circleci/config.yml_disabled
tbak: I am keeping it for reference. I still have to move docker-compose section to Github actions

PR #1115, opened 2021-09-20
Take incoming platform Sidecars and make annotations for them

solarkennedy:
This new feature takes the new PlatformSidecar objects from the
Titus API proto definition and populates the appropriate
annotations on the pod. No additional processing is required by
the control-plane, because we want to stay pretty close to
the "the pod spec is the api" vision.

Two big things we need to iron out:

1. Annotation format. It would be nice to get some good conventions,
and hopefully not change them over time. Luckily in the webhook
we can parse multiple annotations and merge them together as we change.

2. This whole `Struct` stuff. The reasoning behind using a protobuf
[Struct](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Struct)
field was to preserve the structured data that would come in
from the API, and not treat it like a "string" of escaped json.

I think this is still probably a good idea, and only stringify
it out at the last second sounds also like a good idea?
But I've never done something like this in java, so it needs
extra scrutiny. Also I don't know if it is kosher for the actual
POJO to "just" be a struct, but it sure simplifies things!

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/PlatformSidecar.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/PlatformSidecar.java
@@ -8,7 +8,6 @@ import javax.validation.Valid;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonSerializer;
@@ -123,7 +122,7 @@ class ProtoStructSerializer extends JsonSerializer<Struct> {
 
 class ProtoStructDeserializer extends JsonDeserializer<Struct> {
     @Override
-    public Struct deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+    public Struct deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
         Struct.Builder argsBuilder = Struct.newBuilder();
         Map m = p.readValueAs(Map.class);
         JsonFormat.parser().merge(m.toString(), argsBuilder);
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/PlatformSidecar.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/PlatformSidecar.java
@@ -1,21 +1,8 @@
 package com.netflix.titus.api.jobmanager.model.job;
 
-import java.io.IOException;
-import java.util.Map;
 import java.util.Objects;
 import javax.validation.Valid;
 
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.core.JsonGenerator;
-import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.databind.DeserializationContext;
-import com.fasterxml.jackson.databind.JsonDeserializer;
-import com.fasterxml.jackson.databind.JsonSerializer;
-import com.fasterxml.jackson.databind.SerializerProvider;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import com.google.protobuf.Struct;
-import com.google.protobuf.util.JsonFormat;
 
 public class PlatformSidecar {
 
@@ -26,16 +13,9 @@ public class PlatformSidecar {
     private final String channel;
 
     @Valid
-    @JsonSerialize(using = ProtoStructSerializer.class)
-    @JsonDeserialize(using = ProtoStructDeserializer.class)
-//    @JsonSerialize(using = MyMessageSerializer.class)
-//    @JsonDeserialize(using = MyMessageDeserializer.class, as = Struct.class, keyAs = String.class, contentAs = Value.class)
-//    @JsonSerialize(as = Struct.class, keyAs = String.class, contentAs = Value.class)
-//    @JsonDeserialize(as = Struct.class, keyAs = String.class, contentAs = Value.class)
-    @JsonIgnoreProperties(ignoreUnknown = true)
-    public final Struct arguments;
+    public final String arguments;
 
-    public PlatformSidecar(String name, String channel, Struct arguments) {
+    public PlatformSidecar(String name, String channel, String arguments) {
         this.name = name;
         this.channel = channel;
         this.arguments = arguments;
@@ -49,7 +29,7 @@ public class PlatformSidecar {
         return channel;
     }
 
-    public Struct getArguments() {
+    public String getArguments() {
         return arguments;
     }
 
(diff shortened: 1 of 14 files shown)


solarkennedy:
I think one more thing I need feedback on is how to best handle the json serialization exceptions.
I'm very afraid of NPE'ing the process, so I would like to fail as safely as possible.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/PlatformSidecar.java
```diff
@@ -0,0 +1,131 @@
...
+    @Valid
+    private final String channel;
+
+    @Valid
+    @JsonSerialize(using = ProtoStructSerializer.class)
+    @JsonDeserialize(using = ProtoStructDeserializer.class)
+//    @JsonSerialize(using = MyMessageSerializer.class)
+//    @JsonDeserialize(using = MyMessageDeserializer.class, as = Struct.class, keyAs = String.class, contentAs = Value.class)
+//    @JsonSerialize(as = Struct.class, keyAs = String.class, contentAs = Value.class)
+//    @JsonDeserialize(as = Struct.class, keyAs = String.class, contentAs = Value.class)
```
fabiokung, line 34: experiments/testing leftovers?

solarkennedy: Yea :(
Disregard for now because I'm going to redo how we serialize after that meeting.

solarkennedy:
This is ready for the next phase of review. All tests pass, and I've confirmed on my dev stack that it correctly takes structure arguments, and those arguments are easily visible in the UI as true json, but under the hood in cassandra they are stringified json, and that same stringified json is what ends up on the pod annotation.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -406,7 +411,32 @@ private static SharedContainerVolumeSource toCoreSharedVolumeSource(com.netflix.
...
+                .withChannel(ps.getChannel())
+                .withArguments(structToJSONString(ps.getArguments(), ps.getName()))
+                .build();
+    }
+
+    private static String structToJSONString(Struct arguments, String sidecarName) {
+        try {
+            return JsonFormat.printer().omittingInsignificantWhitespace().print(arguments);
+        } catch (InvalidProtocolBufferException e) {
+            throw new IllegalArgumentException("Unable to serialize arguments for the " + sidecarName + " sidecar: " + e.getMessage());
```
tbak, line 434: We must handle this error somehow in GRPC server and return validation error to the caller just like we do it for other fields. We cannot use here `EntityValidator`, so this must be done explicitly in the GRPC call handler. We do it already for some other fields.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -1071,6 +1102,32 @@ public static Capacity toGrpcCapacity(com.netflix.titus.api.jobmanager.model.job
...
+                .setArguments(jsonStringToStruct(ps.getArguments(), ps.getName()))
+                .build();
+    }
+
+    private static Struct jsonStringToStruct(String arguments, String sidecarName) {
+        Struct.Builder argsBuilder = Struct.newBuilder();
+        try {
+            JsonFormat.parser().merge(arguments, argsBuilder);
+        } catch (InvalidProtocolBufferException e) {
+            throw new IllegalArgumentException("Could not serialize platform sidecar arguments for " + sidecarName + ": " + e.getMessage());
```
tbak, line 1126: This will prevent any external access to this job. I think we should return here `Optional<String>`, and log information about mapping error.

tbak (changes requested):

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: Did you mean `IllegalArgumentException` not `EntityValidator`?

This looked like the standard way to throw errors in invalid grpc proto input in this class, so I figured `IllegalArgumentException` would be handled similarly in the upper layers.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: I see what you mean. 
I hope that never happens, it would be a little hard to debug.
I hope logging that error yet still returning empty Struct isn't going to have any other weird repercussions I can't think about...

solarkennedy: Can you confirm, you would like this method to return `Pair<Struct, Optional<String>>` and have the caller deal with the String if present? (kinda golang style)

And in the `toGrpcPlatformSidecar` method, if it sees that optional string, just log instead?

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
tbak: I meant that ideally validation would be handled by `EntityValidator`, but it works on the internal Job object (not protobuf), so it is not possible to use it that early.
We map validation errors to GRPC model. Please, have a look at the `com.netflix.titus.runtime.endpoint.v3.grp.ErrorResponses` class.

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
tbak: Yes, that would be even better. You can use `Either` type for that (from titus-common).

solarkennedy: Used `Either`

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: Hmm. I'll create circular imports if I try to use that here.
Are you sure I should be using it directly?

What if instead I use the `TitusServiceException` thing? (which in turn uses ErrorResponses to return a sane GRPC error code).
If I use that, the http/grpc error I get looks like:


```
E           tests.titus_client.titus_client_exceptions.HTTPClientException: 400: Error Bad Request for url: https://titusapi-test-devkyleacell001.vip.us-east-1.test.cloud.netflix.net:7004/api/v3/jobs: {'message': 'Error creating core job descriptor: Unable to '
E                       'serialize arguments for the sleepy sidecar: HI',
E            'statusCode': 400}
```

Which I think is what you wanted to accomplish, to fail way early?

solarkennedy:
(Rebased on master to account for the volumeMount changes)

tbak (approved):

PR #1116, opened 2021-09-21
Added VolumeMount support

solarkennedy:
This adds VolumeMount support to the main container and extraContainers,
so that they can actually use the Volumes defined in the previous PR.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/BasicContainer.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/BasicContainer.java
@@ -54,18 +54,22 @@ public class BasicContainer {
     )
     private final Map<String, String> env;
 
+    private final List<VolumeMount> volumeMounts;
+
     public BasicContainer(
             String name,
             Image image,
             List<String> entryPoint,
             List<String> command,
-            Map<String, String> env
+            Map<String, String> env,
+            List<VolumeMount> volumeMounts
     ) {
         this.name = name;
         this.image = image;
         this.entryPoint = CollectionsExt.nullableImmutableCopyOf(entryPoint);
         this.command = CollectionsExt.nullableImmutableCopyOf(command);
         this.env = CollectionsExt.nullableImmutableCopyOf(env);
+        this.volumeMounts = CollectionsExt.nullableImmutableCopyOf(volumeMounts);
     }
 
     public String getName() {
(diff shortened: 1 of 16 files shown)


solarkennedy:
This one is ready for review too, but I'm a bit concerned.

For this one, it seems to complained that all existing jobs on my dev stack 

```2021-09-25 01:03:30,520 ERROR [cass_titus_dev-nio-worker-3] service.JobReconciliationFrameworkFactory:? - Bad job record found: jobId=83482f6f-338f-40b8-9383-4def4709699f, violations={jobDescriptor.container='null' in fields: volumeMounts}```

Which makes me think I *must* relax the nonNull constraint?

File: titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/BasicContainer.java [hunk 1]
```diff
@@ -54,18 +54,22 @@
     )
     private final Map<String, String> env;
 
+    private final List<VolumeMount> volumeMounts;
```
tbak, line 57: This class misses validation annotations.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/BasicContainer.java
```diff
@@ -77,20 +81,30 @@ public Image getImage() {
...
-        if (command == null) { return Collections.emptyList();}
+        if (command == null) {
+            return Collections.emptyList();
+        }
         return command;
     }
 
     public Map<String, String> getEnv() {
-        if (env == null) { return Collections.emptyMap();}
+        if (env == null) {
```
tbak, line 98: Validation annotations should ensure that all values are non-null.

tbak (changes requested):

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/BasicContainer.java
solarkennedy: Gotcha, I've removed these from command/env/entrypoint.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/BasicContainer.java
solarkennedy: Added `CollectionInvariants`

solarkennedy:
I can confirm now that this branch correctly creates volumeMounts on the pod object, and is roll-backsafe (jobs with volumemounts still can be read by old code, of course you can't see the volumeMounts in the json, but the code still works).

tbak (approved):

PR #1121, opened 2021-09-28
Use Fabric8IO in the cluster membership management

tbak:
as an option to the standard Kube client


titus-common-ext/kube/src/main/java/com/netflix/titus/ext/kube/clustermembership/connector/transport/fabric8io/Fabric8IOKubeMembershipExecutor.java
```diff
@@ -0,0 +1,155 @@
...
+                                case ADDED:
+                                    sink.next(ClusterMembershipEvent.memberAddedEvent(revision));
+                                    break;
+                                case MODIFIED:
+                                    sink.next(ClusterMembershipEvent.memberUpdatedEvent(revision));
+                                    break;
+                                case DELETED:
+                                    sink.next(ClusterMembershipEvent.memberRemovedEvent(revision));
+                                    break;
+                                case ERROR:
```
fabiokung, line 129: do you know if the fabric8 client guarantees the watch stream will be closed after all `ERROR` actions?

titus-common-ext/kube/src/main/java/com/netflix/titus/ext/kube/clustermembership/connector/transport/fabric8io/crd/FClusterMember.java
```diff
@@ -0,0 +1,88 @@
...
+package com.netflix.titus.ext.kube.clustermembership.connector.transport.fabric8io.crd;
+
+import java.util.List;
+import java.util.Map;
+
+import io.fabric8.kubernetes.model.annotation.Plural;
+import io.fabric8.kubernetes.model.annotation.Singular;
+
+@Singular("clusterMember")
+@Plural("clusterMember")
```
fabiokung, line 26: clusterMembers?

Also, arent CRD names always lowercase by convention? E.g.: `clustermember` and `clustermembers`?

titus-common-ext/kube/src/main/java/com/netflix/titus/ext/kube/clustermembership/connector/transport/fabric8io/crd/FClusterMember.java
```diff
@@ -0,0 +1,88 @@
...
+public class FClusterMember {
+
+    private String memberId;
+
+    private boolean enabled;
+
+    private boolean active;
+
+    private List<FClusterMemberAddress> clusterMemberAddress;
```
fabiokung, line 35: nit: pluralize - `clusterMemberAddresses`

fabiokung (approved):
could we use https://www.javadoc.io/static/io.fabric8/kubernetes-client/5.8.0/io/fabric8/kubernetes/client/KubernetesClient.html#leaderElector-- instead of a CRD?

If you want to go the CRD route, code LGTM

amit-git (approved):

titus-common-ext/kube/src/main/java/com/netflix/titus/ext/kube/clustermembership/connector/transport/fabric8io/Fabric8IOKubeMembershipExecutor.java
tbak: No, it will not be closed. As I understand this is potentially recoverable error. But the callback does not provide any information on the nature of the error, so there is nothing to be logged here.

Here you have code that handles it:
```java
  protected boolean onStatus(Status status) {
    // The resource version no longer exists - this has to be handled by the caller.
    if (status.getCode() == HTTP_GONE) {
      close(new WatcherException(status.getMessage(), new KubernetesClientException(status)));
      return true;
    }

    eventReceived(Action.ERROR, null);
    logger.error("Error received: {}", status);
    return false;
  }
```

`close` is call if the error is not recoverable, which happens only if the resource itself is gone.

tbak:
>> could we use https://www.javadoc.io/static/io.fabric8/kubernetes-?? 
>> client/5.8.0/io/fabric8/kubernetes/client/KubernetesClient.html#leaderElector-- instead of a CRD?

We use CRDs for cluster membership management, not leader election. The API above does not provide equivalent functionality. The question is if we should simplify this implementation or provide an alternative implementation which is a thin wrapper around this one. I think it is something we might want to consider in the future.

PR #1135, opened 2021-10-13
Removed the useBytes toggle, always set bytes on pods

solarkennedy:
Now that we have 100% rollout of this feature, we are
ready to remove the toggle.

This change must be deployed *after* the titus-executor
change to always use bytes


solarkennedy:
The newer version of titus-executor that always assumes bytes is everywhere.
Should be safe to merge and deploy this now.

PR #1141, opened 2021-10-14
Update integration tests partitions to improve the build execution time

tbak:


tbak:
The build time reduced from ~18min to ~10min.

tbak:
@solarkennedy Good question. I verified it locally. The partitioning itself was applied previously. This is only a rearrangement, as some tasks were running short (2min) and one very long (16min).

PR #1155, opened 2021-10-27
Support consistent reads from the TitusGateway job cache

tbak:
Consistent reads are implemented by two mechanisms:
* TitusGateway cache connector sends periodically keep alive requests to TJC. A keep alive response when received is guaranteed to come after all events proceeding its creation.
* The internal TJC job service emits checkpoint events that are used to decide when a keep alive response can be sent

### DIFF
--- a/titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/LocalCacheQueryProcessor.java
+++ b/titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/LocalCacheQueryProcessor.java
@@ -25,11 +25,14 @@ import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import javax.annotation.PreDestroy;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import com.netflix.spectator.api.Registry;
 import com.netflix.titus.api.jobmanager.model.job.Job;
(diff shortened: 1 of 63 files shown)


titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/LocalCacheQueryProcessor.java
```diff
@@ -209,46 +249,73 @@ private boolean isJobFinishedEvent(JobChangeNotification event) {
...
+                                    "Downstream event stream reconnected."
+                            )));
+                        }
+
+                        // Job data replicator emits keep alive events periodically if there is nothing in the stream. We have
+                        // to filter them out here.
+                        if (jobManagerEvent instanceof JobKeepAliveEvent) {
+                            // Check if staleness is not too high.
+                            if (jobDataReplicator.getStalenessMs() > configuration.getObserveJobsStalenessDisconnectMs()) {
+                                return Mono.error(new StatusRuntimeException(Status.ABORTED.augmentDescription(
```
joshi-keyur, line 282: Log the observed value of the staleness from the received keep alive event?

joshi-keyur (approved):

titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/LocalCacheQueryProcessor.java
tbak: I added metric to track this event.

PR #1157, opened 2021-10-29
relocation de-scheduler fix

amit-git:
### Runtime type error fix

Relocation service runs into the runtime type error while de-scheduling eligible tasks from marked or unhealthy agents. This change is similar in nature to the bug-fix done to Metrics step code in order avoid incorrect casting of Map to a List from PCollectionJobSnapsho

### DIFF
--- a/titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/descheduler/DefaultDeschedulerService.java
+++ b/titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/descheduler/DefaultDeschedulerService.java
@@ -77,11 +77,8 @@ public class DefaultDeschedulerService implements DeschedulerService {
 
     @Override
     public List<DeschedulingResult> deschedule(Map<String, TaskRelocationPlan> plannedAheadTaskRelocationPlans) {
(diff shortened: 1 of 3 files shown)


titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/descheduler/DefaultDeschedulerService.java
```diff
@@ -77,11 +77,8 @@ public DefaultDeschedulerService(ReadOnlyJobOperations jobOperations,
 
     @Override
     public List<DeschedulingResult> deschedule(Map<String, TaskRelocationPlan> plannedAheadTaskRelocationPlans) {
-        List<Pair<Job, List<Task>>> allJobsAndTasks = jobOperations.getJobsAndTasks();
```
tbak, line 80: The reason why we should use this method is to have a consistent snapshot of the data. If you ask for the jobs first, and next for tasks, it is possible that you get tasks for which there are no jobs in the`jobs` collection.
I think we should make a fix but handling properly the result which is `List<Pair<Job, Map<String,Task>>>`.

tbak (approved):

PR #1159, opened 2021-10-29
Added basic platformSidecar job query search support

solarkennedy:
This adds the most basic job query search based on platform
sidecars.

It only adds a single param `platformSidecar` so one can get all
the jobs/tasks that have use that platformSidecar.

I expect all queries around platform sidecars like this to be centered
around a specific sidecar, not multiple sidecars, so I didn't think we
needed additional complexity around N sidecars.

### DIFF
--- a/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/query/V3AbstractQueryCriteriaEvaluator.java
+++ b/titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/query/V3AbstractQueryCriteriaEvaluator.java
@@ -27,6 +27,7 @@ import java.util.function.Predicate;
 import com.netflix.titus.api.jobmanager.model.job.Job;
 import com.netflix.titus.api.jobmanager.model.job.JobGroupInfo;
 import com.netflix.titus.api.jobmanager.model.job.JobState;
+import com.netflix.titus.api.jobmanager.model.job.PlatformSidecar;
 import com.netflix.titus.api.jobmanager.model.job.ext.BatchJobExt;
 import com.netflix.titus.common.util.tuple.Pair;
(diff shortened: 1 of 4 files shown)


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/query/V3AbstractQueryCriteriaEvaluator.java
```diff
@@ -155,6 +157,19 @@ public boolean test(Pair<Job<?>, TASK_OR_SET> jobListPair) {
         return apply(imageTag, job -> job.getJobDescriptor().getContainer().getImage().getTag());
     }
 
+    private Optional<Predicate<Pair<Job<?>, TASK_OR_SET>>> applyPlatformSidecar(Optional<String> platformSidecarNameOpt) {
+        return platformSidecarNameOpt.map(platformSidecarName ->
+                jobTaskPair -> {
+                    List<PlatformSidecar> platformSidecars = jobTaskPair.getLeft().getJobDescriptor().getPlatformSidecars();
+                    List<String> psNames = new ArrayList<>();
+                    for (PlatformSidecar ps : platformSidecars) {
+                        psNames.add(ps.getName());
```
tbak, line 166: You could make the comparison here, and return immediately if `ps.geName().equals(platformSidecarName)`

aaronlehmann:
Wow, that was fast. Thanks!

titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/query/V3AbstractQueryCriteriaEvaluator.java
solarkennedy: Oh yea of course!

tbak (approved):

PR #1165, opened 2021-11-02
Adding kube shared informer to gateway

amitaekbote:
We need to get container status from pods. The pods information is not available at the gateway layer until we add shared informer. This PR adds shared informer and code to get the container status. The shared informer will be behind a feature flag.

### DIFF
--- a/titus-server-gateway/src/main/java/com/netflix/titus/gateway/kubernetes/F8KubeObjectFormatter.java
+++ b/titus-server-gateway/src/main/java/com/netflix/titus/gateway/kubernetes/F8KubeObjectFormatter.java
@@ -1,0 +1,50 @@
+package com.netflix.titus.gateway.kubernetes;
+
+import io.fabric8.kubernetes.api.model.ObjectMeta;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.PodSpec;
+import io.fabric8.kubernetes.api.model.PodStatus;
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
+import io.kubernetes.client.openapi.models.V1Pod;
+import io.kubernetes.client.openapi.models.V1PodSpec;
+import io.kubernetes.client.openapi.models.V1PodStatus;
+
+public class F8KubeObjectFormatter {
+
+    public static String formatPodEssentials(Pod pod) {
+        try {
+            return formatPodEssentialsInternal(pod);
+        } catch (Exception e) {
+            return "pod formatting error: " + e.getMessage();
+        }
+    }
+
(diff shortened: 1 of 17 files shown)


titus-server-gateway/src/main/java/com/netflix/titus/gateway/kubernetes/F8KubeObjectFormatter.java
```diff
@@ -0,0 +1,50 @@
+package com.netflix.titus.gateway.kubernetes;
+
+import io.fabric8.kubernetes.api.model.ObjectMeta;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.PodSpec;
+import io.fabric8.kubernetes.api.model.PodStatus;
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
```
tbak, line 7: These are imports for the other Kube client. They should not be needed.

titus-server-gateway/src/main/java/com/netflix/titus/gateway/startup/TitusGateway.java
```diff
@@ -46,6 +50,14 @@ public static void main(String[] args) throws Exception {
             System.exit(1);
         }
 
+        try {
+            KubeApiConnector kubeApiConnector = new KubeApiConnector(Fabric8IOClients.createFabric8IOClient(), TitusRuntimes.internal());
```
tbak, line 54: Do not use `TitusRuntimes.internal()`, but the `TitusRuntime` instance provided by Guice.

tbak (changes requested):

titus-server-gateway/src/main/java/com/netflix/titus/gateway/startup/TitusGateway.java
```diff
@@ -26,9 +26,14 @@
...
 import com.netflix.governator.LifecycleInjector;
 import com.netflix.governator.guice.jetty.Archaius2JettyModule;
+import com.netflix.titus.common.runtime.TitusRuntime;
+import com.netflix.titus.common.runtime.TitusRuntimes;
 import com.netflix.titus.common.util.guice.ContainerEventBus;
+import com.netflix.titus.gateway.kubernetes.KubeApiConnector;
+import com.netflix.titus.runtime.connector.kubernetes.Fabric8IOClients;
 import com.sampullara.cli.Args;
 import com.sampullara.cli.Argument;
+import io.fabric8.kubernetes.client.NamespacedKubernetesClient;
```
tbak, line 36: A lot of new imports. Are they needed?

tbak (approved):

PR #1166, opened 2021-11-03
Emit job/task archive events in the GRPC job event stream

tbak:
This change extends the job GRPC event stream by the job/task archive events. These events are only emitted by `ObserveJobsWithKeepAlive` endpoint for backward compatibility (and because they are meant for the internal consumption only).


tbak:
Merging it now. I will address code review requests in a separate PR.

PR #1171, opened 2021-11-12
Add fields filtering in the job event stream

tbak:


titus-client/src/main/java/com/netflix/titus/runtime/connector/jobmanager/replicator/GrpcJobReplicatorEventStream.java
```diff
@@ -242,7 +242,9 @@ public GrpcJobReplicatorEventStream(JobManagementClient client,
                             }
                         }
                     } else {
-                        titusRuntime.getCodeInvariants().inconsistent("Job record not found: jobId=%s, taskId=%s", task.getJobId(), task.getId());
+                        if (!jobsToRemove.contains(task.getJobId())) {
+                            titusRuntime.getCodeInvariants().inconsistent("Job record not found: jobId=%s, taskId=%s", task.getJobId(), task.getId());
```
andrew-leung, line 246: Is there a way to log this at a different level (e.g., just info level)? This is by far the most frequently occurring RADAR alert and it's not really actionable.

andrew-leung (approved):

titus-client/src/main/java/com/netflix/titus/runtime/connector/jobmanager/replicator/GrpcJobReplicatorEventStream.java
tbak: This check should eliminate all these alerts. If this happens again, there is a bug that must be fixed.

PR #1176, opened 2021-11-17
Inject container states into task

amitaekbote:
Injecting container states into tasks at gateway

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/FeatureActivationConfiguration.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/FeatureActivationConfiguration.java
@@ -54,4 +54,10 @@ public interface FeatureActivationConfiguration {
      */
     @DefaultValue("true")
     boolean isRelocationBinpackingEnabled();
+
+    /**
+     * Enable shared informer on Titus gateway
+     */
+    @DefaultValue("false")
+    boolean isKubeSharedInformerEnabled();
 }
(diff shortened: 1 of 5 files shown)


titus-api/src/main/java/com/netflix/titus/api/FeatureActivationConfiguration.java
```diff
@@ -54,4 +54,10 @@
      */
     @DefaultValue("true")
     boolean isRelocationBinpackingEnabled();
+
+    /**
+     * Enable shared informer on Titus gateway
+     */
+    @DefaultValue("false")
+    boolean isKubeSharedInformerEnabled();
```
tbak, line 62: I believe the shared informer is always enabled. Feature flag is for injecting container status. I think we should change the name of this method accordingly.

tbak (approved):

titus-server-gateway/src/main/java/com/netflix/titus/gateway/kubernetes/KubeApiConnector.java
```diff
@@ -93,10 +93,12 @@ public KubeApiConnector(NamespacedKubernetesClient kubernetesClient, TitusRuntim
                 ContainerStatus containerStatus = iterator.next();
                 ContainerState status = containerStatus.getState();
                 TaskStatus.ContainerState.ContainerHealth containerHealth = TaskStatus.ContainerState.ContainerHealth.Unset;
-                if(status.toString().equals("running")) {
+                if(status.toString().equalsIgnoreCase("running")) {
```
solarkennedy, line 96: Can you take advantage of the client libs enums like `V1ContainerState.SERIALIZED_NAME_RUNNING` ?

amitaekbote: the fabric 8 kube client does not have that

PR #1177, opened 2021-11-17
Added SaaSVolume support

solarkennedy:
This adds support for a new SaaS Volume type.
It only has the 1 string to worry about, the ID.

On the backend (TJC), pods are created using the Ceph
volume type, with the ID as the path.

### DIFF
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/SaaSVolumeSource.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/SaaSVolumeSource.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.titus.api.jobmanager.model.job.volume;
+
+import java.util.Objects;
+import javax.validation.Valid;
+
+public class SaaSVolumeSource extends VolumeSource {
+
+    @Valid
+    private final String saaSVolumeID;
+
+    public SaaSVolumeSource(String saaSVolumeID) {
+        this.saaSVolumeID = saaSVolumeID;
+    }
+
+    public static SaaSVolumeSource.Builder newBuilder() {
+        return new SaaSVolumeSource.Builder();
+    }
+
+    public static SaaSVolumeSource.Builder newBuilder(SaaSVolumeSource saaSVolumeSource) {
+        return new SaaSVolumeSource.Builder()
+                .withSaaSVolumeID(saaSVolumeSource.saaSVolumeID);
+    }
+
+    public String getSaaSVolumeID() {
+        return saaSVolumeID;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        SaaSVolumeSource that = (SaaSVolumeSource) o;
+        return Objects.equals(saaSVolumeID, that.saaSVolumeID);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(saaSVolumeID);
+    }
+
+    @Override
+    public String toString() {
+        return "SaaSVolumeSource{" +
+                "SaaSVolumeID='" + saaSVolumeID + '\'' +
+                '}';
+    }
+
+    public static final class Builder<E extends VolumeSource> {
+        private String saaSVolumeID;
+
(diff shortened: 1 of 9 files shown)


titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
```diff
@@ -1159,12 +1171,22 @@ public static Capacity toGrpcCapacity(com.netflix.titus.api.jobmanager.model.job
                     .setName(volume.getName())
                     .mergeSharedContainerVolumeSource(toGrpcSharedVolumeSource((SharedContainerVolumeSource) source))
                     .build();
+        } else if (source instanceof SaaSVolumeSource) {
+            return com.netflix.titus.grpc.protogen.Volume.newBuilder()
+                    .setName(volume.getName())
+                    .mergeSaaSVolumeSource(toGrpcSaaSVolumeSource((SaaSVolumeSource) source))
```
tbak, line 1177: Why not `set`?

tbak (approved):

titus-client/src/main/java/com/netflix/titus/runtime/endpoint/v3/grpc/GrpcJobManagementModelConverters.java
solarkennedy: Whoops.
No reason. It should be set.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/SaaSVolumeSource.java
```diff
@@ -0,0 +1,66 @@
+package com.netflix.titus.api.jobmanager.model.job.volume;
```
andrew-leung, line 1: Missing OSS header. True for other new files as well.

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/SaaSVolumeSource.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.Objects;
+import javax.validation.Valid;
+
+public class SaaSVolumeSource extends VolumeSource {
+
+    @Valid
```
andrew-leung, line 8: I don't think the validation does anything since there isn't a validator to delegate to. My `@NotNull` or `@MinSize` makes more sense?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
```diff
@@ -247,6 +252,55 @@ private static String buildV1VolumeMountPropagation(String mountPropagation) {
...
+    }
+
+    private static Optional<V1Volume> buildV1Volume(Volume volume) {
+        if (volume.getVolumeSource() instanceof SharedContainerVolumeSource) {
+            V1FlexVolumeSource flexVolume = getV1FlexVolumeForSharedContainerVolumeSource(volume);
+            return Optional.ofNullable(new V1Volume()
+                    .name(volume.getName())
+                    .flexVolume(flexVolume));
+        } else if (volume.getVolumeSource() instanceof SaaSVolumeSource) {
+            V1CephFSVolumeSource cephVolume = getV1CephVolumeSource(volume);
```
andrew-leung, line 273: In the future, if the SaaS volume ID can be backed by multiple storage providers, how does TJC know which one to configure here? Would it need to call the SaaS control plane to get its type, have the user provide it, or TBD?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
```diff
@@ -247,6 +252,55 @@ private static String buildV1VolumeMountPropagation(String mountPropagation) {
...
+        options.put("sourceContainer", sharedContainerVolumeSource.getSourceContainer());
+        options.put("sourcePath", sharedContainerVolumeSource.getSourcePath());
+        return new V1FlexVolumeSource()
+                .driver("SharedContainerVolumeSource")
+                .options(options);
+    }
+
+    private static V1CephFSVolumeSource getV1CephVolumeSource(Volume volume) {
+        SaaSVolumeSource saaSVolumeSource = (SaaSVolumeSource) volume.getVolumeSource();
+        // This CEPH volume source is not "truly" representative of all the data we will
```
andrew-leung, line 295: FYI, Ceph isn't an acronym.

andrew-leung (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
solarkennedy: That is a titus-executor decision.
Ceph will probably still be the transport, but the backend store of the *ceph storage nodes* may be like, s3, ebs, ssds, who knows. 

If some day we decide to change the *transport*, TBD, but still will be an internal TJC decision.
Or heck maybe we'll just use/abuse the Ceph slot in the pod spec :)

titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/volume/SaaSVolumeSource.java
solarkennedy: Oh I thought it was just a handy way to make sure it was non-null too? (I thought it was a superset of `@NotNull`)

PR #1181, opened 2021-11-19
Added container image tag to v1 pod schema pods

solarkennedy:
We need to preserve this Image Tag somehow into
the v1 pod so that we can retrieve it on the backend
and use it for Metatron to validate.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
@@ -129,6 +129,7 @@ import static com.netflix.titus.master.kubernetes.pod.KubePodConstants.NEVER_RES
 import static com.netflix.titus.master.kubernetes.pod.KubePodConstants.POD_CPU_BURSTING_ENABLED;
 import static com.netflix.titus.master.kubernetes.pod.KubePodConstants.POD_FUSE_ENABLED;
 import static com.netflix.titus.master.kubernetes.pod.KubePodConstants.POD_HOSTNAME_STYLE;
+import static com.netflix.titus.master.kubernetes.pod.KubePodConstants.POD_IMAGE_TAG_PREFIX;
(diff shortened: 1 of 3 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
```diff
@@ -207,6 +208,10 @@ public V1Pod buildV1Pod(Job<?> job, Task task, boolean useKubeScheduler, boolean
         labels.put("v3.job.titus.netflix.com/job-id", job.getId());
         labels.put("v3.job.titus.netflix.com/task-id", taskId);
 
+        // A V1Container has no room to store the original tag that the Image came from, so we store it as an
+        // annotation. Only saving the 'main' one for now.
+        annotations.put(POD_IMAGE_TAG_PREFIX + "main", job.getJobDescriptor().getContainer().getImage().getTag());
```
JackOfMostTrades, line 213: Possibly a dumb question, but jobs can be submitted with a digest instead of a tag, right? Does this behave the right way if a job is submitted with a digest?

solarkennedy: Great question.

I guess... this would be NPE? 😬 
Or the string `null` ? I'm not super sure.

solarkennedy: Ok confirmed this is `""` in the case where it is submitted with *only* a digest and no tag.

I'm pretty sure this is what we want and should be considered correct behavior?

I don't want it to *not* be set and assume.

JackOfMostTrades: Ok, I think that should plumb an empty string all the way through then which should be fine.

JackOfMostTrades (approved):

tbak (approved):

PR #1182, opened 2021-11-19
Change the job topology spreading rules

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/topology/DefaultTopologyFactory.java
```diff
@@ -146,15 +149,20 @@ private static int getJobMaxSkew(Job<?> job) {
...
 
         DisruptionBudgetPolicy policy = job.getJobDescriptor().getDisruptionBudget().getDisruptionBudgetPolicy();
+        // Job spreading is only relevant for jobs that care about the availability.
         if (!(policy instanceof AvailabilityPercentageLimitDisruptionBudgetPolicy)) {
-            return 1;
+            return -1;
         }
         int jobSize = getJobDesiredSize(job);
         if (jobSize <= 1) {
             return 1;
```
joshi-keyur, line 158: Why not just disable the topology spread by returning -1?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/topology/DefaultTopologyFactory.java
```diff
@@ -146,15 +149,20 @@ private static int getJobMaxSkew(Job<?> job) {
         }
 
         DisruptionBudgetPolicy policy = job.getJobDescriptor().getDisruptionBudget().getDisruptionBudgetPolicy();
+        // Job spreading is only relevant for jobs that care about the availability.
```
rostyboost, line 152: 🎆

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodConfiguration.java
```diff
@@ -135,6 +135,19 @@
     @DefaultValue("NONE")
     String getDisabledJobSpreadingPattern();
 
+    /**
+     * Job spreading skew adjustment factor, with the skew computed as job_size/alpha. Setting value >= 1 effectively
```
gabrielhartmann, line 139: Do you mean `<` 1?  We expect to set this value greater than 1.  In fact the default is `3`.  I prefer naming this variable as `desiredNodeCount` or something like that, or at least comment this as the motivation for the value, so in 6 months we don't have to figure out what "alpha" means again.

gabrielhartmann (approved):

rostyboost (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/topology/DefaultTopologyFactory.java
tbak: Good point. Will do that.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodConfiguration.java
tbak: Yes, I meant `<`. Will fix it.
I did not want to use the term `desiredNodeCount` as this is true only for small jobs. It is more like a number of job partitions, where each partition may be placed on one node if it would fit into it.

gabrielhartmann: Ok.  Totally fine.  Maybe a little commentary in the comment for future developers would be helpful, maybe not.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodConfiguration.java
```diff
@@ -135,6 +135,19 @@
     @DefaultValue("NONE")
     String getDisabledJobSpreadingPattern();
 
+    /**
+     * Job spreading skew adjustment factor, with the skew computed as job_size/alpha. Setting value <= 1 effectively
```
fabiokung, line 139: minor: `s/with the skew/with the max skew/`

fabiokung: also perhaps a good idea to mention this is job spreading on Nodes (not AZs, or some other topology dimension)

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/topology/DefaultTopologyFactory.java
```diff
@@ -146,15 +149,20 @@ private static int getJobMaxSkew(Job<?> job) {
...
         if (!(policy instanceof AvailabilityPercentageLimitDisruptionBudgetPolicy)) {
```
fabiokung, line 153: minor, can we make this part of the `DisruptionBudgetPolicy` contract? It will make this behavior more robust against new disruption policies being added in the future (e.g. I will soon add another `GangMigration` policy and will most definitely not remember to come check this logic):

```diff
diff --git a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/disruptionbudget/DisruptionBudgetPolicy.java b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/disruptionbudget/DisruptionBudgetPolicy.java
index d7acaa646..fb5513bc9 100644
--- a/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/disruptionbudget/DisruptionBudgetPolicy.java
+++ b/titus-api/src/main/java/com/netflix/titus/api/jobmanager/model/job/disruptionbudget/DisruptionBudgetPolicy.java
@@ -17,4 +17,5 @@
 package com.netflix.titus.api.jobmanager.model.job.disruptionbudget;

 public abstract class DisruptionBudgetPolicy {
+    public abstract boolean isJobSpreadingRequired();
 }
```

fabiokung (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/topology/DefaultTopologyFactory.java
tbak: We could, but we agreed that the spreading is a concept that has nothing to do with the disruption budget.
I added this check here now, but it could change to something else in the future.
I am not totally opposed, but let's think a bit more about that.

joshi-keyur (approved):

PR #1184, opened 2021-11-19
Add print/remove unschedulable job commands

tbak:
Add print/remove unschedulable job commands

### DIFF
--- a/titus-cli/src/main/java/com/netflix/titus/cli/command/job/unschedulable/PrintUnschedulableJobsCommand.java
+++ b/titus-cli/src/main/java/com/netflix/titus/cli/command/job/unschedulable/PrintUnschedulableJobsCommand.java
@@ -59,8 +59,11 @@ public class PrintUnschedulableJobsCommand implements CliCommand {
                 .orElseThrow(() -> new IllegalArgumentException("Wrong expiry threshold"));
 
         Pair<Map<String, Job>, Map<String, Map<String, Task>>> all = loadActiveJobsAndTasks(context);
-        Map<String, UnschedulableJob> unschedulable = UnschedulableFinder.findUnschedulableJobs(all.getLeft(), all.getRight(), stuckInAcceptedThresholdMs);
+        Map<String, UnschedulableJob> unschedulable = UnschedulableFinder.findUnschedulableJobs(context, all.getLeft(), all.getRight(), stuckInAcceptedThresholdMs);
         logger.info("Found {} unschedulable jobs", unschedulable.size());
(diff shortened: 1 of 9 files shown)


titus-cli/src/main/java/com/netflix/titus/cli/command/job/unschedulable/UnschedulableFinder.java
```diff
@@ -0,0 +1,64 @@
...
+            long waitTimeMs = System.currentTimeMillis() - acceptedTimestamp;
+            if (waitTimeMs < stuckInAcceptedThresholdMs) {
+                return Optional.empty();
+            }
+        }
+
+        // All tasks are in the Accepted state for more than stuckInAcceptedThreshold.
+        return Optional.of(new UnschedulableJob(
+                job.getId(),
+                "All tasks are stuck in the 'Accepted' state for too long. " +
```
solarkennedy, line 60: Can you print `stuckInAcceptedThresholdMs` instead of `too long` ?

titus-cli/src/main/java/com/netflix/titus/cli/command/job/unschedulable/PrintUnschedulableJobsCommand.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public void execute(CommandContext context) throws Exception {
+        long stuckInAcceptedThresholdMs = TimeUnitExt.toMillis(context.getCLI().getOptionValue('e'))
+                .orElseThrow(() -> new IllegalArgumentException("Wrong expiry threshold"));
+
+        Pair<Map<String, Job>, Map<String, Map<String, Task>>> all = loadActiveJobsAndTasks(context);
+        Map<String, UnschedulableJob> unschedulable = UnschedulableFinder.findUnschedulableJobs(all.getLeft(), all.getRight(), stuckInAcceptedThresholdMs);
+        logger.info("Found {} unschedulable jobs", unschedulable.size());
+        unschedulable.forEach((jobId, info) -> logger.info("    {}: {}", jobId, info.getReason()));
+    }
```
solarkennedy, line 65: I don't see what `getReason()` will output (I don't understand how `reason` is set in the `UnschedulableJob`.
Is it going to be last scheduling result? (that would be cool)

solarkennedy (approved):

titus-cli/src/main/java/com/netflix/titus/cli/command/job/unschedulable/PrintUnschedulableJobsCommand.java
tbak: I like the idea with the last scheduling result. Let me do that.

PR #1185, opened 2021-11-19
Inject container status and task relocation data in the job event stream

tbak:


amitaekbote (approved):

titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/GatewayJobServiceGateway.java
```diff
@@ -294,17 +293,21 @@ public GatewayJobServiceGateway(GrpcRequestConfiguration tunablesConfiguration,
     @Override
     public Observable<JobChangeNotification> observeJob(String jobId, CallMetadata callMetadata) {
         if (localCacheQueryProcessor.canUseCache(Collections.emptyMap(), "observeJob", callMetadata)) {
-            return localCacheQueryProcessor.syncCache("observeJob", JobChangeNotification.class).concatWith(localCacheQueryProcessor.observeJob(jobId));
+            return localCacheQueryProcessor.syncCache("observeJob", JobChangeNotification.class)
+                    .concatWith(localCacheQueryProcessor.observeJob(jobId))
+                    .map(taskRelocationDataInjector::injectIntoTaskUpdateEvent);
         }
-        return super.observeJob(jobId, callMetadata);
+        return super.observeJob(jobId, callMetadata).map(taskRelocationDataInjector::injectIntoTaskUpdateEvent);
```
solarkennedy, line 300: I've not seen this code before, but before, this localCacheQueryProcessor thing was doing the heavy lifting, regardless of the number of observeJobs streamers.

But now with the `.map()`, will we be calling (or really, looking at the shared informer) for each job and for each streamer? If so, any worries about scaling here?

solarkennedy (approved):

titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/GatewayJobServiceGateway.java
tbak: `localCacheQueryProcessor` provides the raw data. The data injector decorates them with additional information.
The shared informer calls should be inexpensive, just like setting the data in the protobuf messages.

titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/TaskDataInjector.java
```diff
@@ -182,11 +205,11 @@ private long getTaskRelocationTimeout() {
 
     private Task newTaskWithContainerState(Task task) {
         return task.toBuilder().setStatus(TaskStatus.newBuilder()
-                    .addAllContainerState(kubeApiConnector.getContainerState(task.getId()))).build();
+                .addAllContainerState(kubeApiConnector.getContainerState(task.getId()))).build();
```
solarkennedy, line 208: I think this line nukes out whatever original status stuff we had when we do `setStatus` to this new status we build! (which only has ContainerState, and none of whatever we originally had)

solarkennedy: Should be `task.toBuilder().setStatus(task.getStatus().toBuilder()` I think

amitaekbote: https://github.com/Netflix/titus-control-plane/pull/1189

PR #1194, opened 2021-12-09
Provide more information about the relocation trigger

tbak:


titus-supplementary-component/task-relocation/src/main/java/com/netflix/titus/supplementary/relocation/workflow/step/TaskEvictionStep.java
```diff
@@ -90,9 +90,20 @@ public TaskEvictionStep(EvictionServiceClient evictionServiceClient,
                 .collect(Collectors.toMap(
                         TaskRelocationPlan::getTaskId,
                         p -> {
-                            String message = String.format("%s: reasonCode=%s, plannedRelocationTime=%s",
-                                    p.getReasonMessage(), p.getReason(), DateTimeExt.toUtcDateTimeString(p.getRelocationTime())
-                            );
+                            String message;
+                            switch (p.getReason()) {
```
joshi-keyur, line 94: Do we need to handle `null` reason value?

tbak: Good catch. I will make sure in the builder that this value is always set to something.

joshi-keyur (approved):

PR #1196, opened 2021-12-17
Use EBS annotations for v0, and vol/volmount for v1 pods

solarkennedy:
Previously we were always creating annotations *and* vol/volmount
for EBS on v0 *and* v1 pods. I found this confusing, because
annotations are used on v0 and vols are used on v1.

I didn't want to perpetuate the annotations (wasting space)
for no reason on v1 pods.

This change makes it more clear that on v0 we use annotations
only, and on v1 we use pod constructs only.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
@@ -146,8 +146,7 @@ import static com.netflix.titus.master.kubernetes.pod.KubePodConstants.WORKLOAD_
 import static com.netflix.titus.master.kubernetes.pod.KubePodConstants.WORKLOAD_OWNER_EMAIL;
 import static com.netflix.titus.master.kubernetes.pod.KubePodConstants.WORKLOAD_SEQUENCE;
 import static com.netflix.titus.master.kubernetes.pod.KubePodConstants.WORKLOAD_STACK;
-import static com.netflix.titus.master.kubernetes.pod.KubePodUtil.buildV1VolumeInfo;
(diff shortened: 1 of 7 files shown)


tbak (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
```diff
@@ -460,7 +458,6 @@ V1ResourceRequirements buildV1ResourceRequirements(ContainerResources containerR
         });
 
         appendS3WriterRole(annotations, job, task);
-        annotations.putAll(createEbsPodAnnotations(job, task));
```
andrew-leung, line 463: I believe these annotations are still going to used by our kube exclusive resource scheduler plugin (https://github.com/Netflix/titus-kube-common/blob/master/pod/annotations.go#L104). Is there something else that we are expecting will enforce exclusive access to the V1Volume/PVC?

solarkennedy: Oh. Hmm.
I suppose it would be nice if that plugin respected V1 volumes and didn't have to rely on a custom annotation?

If you point me to the code, maybe I can add in that capability first?

andrew-leung: Here is the plugin code (https://stash.corp.netflix.com/projects/TN/repos/titus-kube-scheduler/browse/plugins/exclusive/ebsvolid/ebsvolid_plugin.go) and CC @joshi-keyur who originally wrote it.

solarkennedy: Hmm. Looks very annotation-specific.

I'll discard for now

PR #1199, opened 2022-01-05
Use awsElasticBlockStore instead of PVS pods

solarkennedy:
This change makes it so we always use the awsElasticBlockStore
volumeSource on a volume, instead of using PV/PVCs.

This does leave a lot of leftover unused code that I will clean up on
the next PR, only after verifying there are no unintended side-effects
from configuring pods this way.

On v0 pods, nothing should change, because we still just use the
annotations as the source of truth.

For v1 pods, tests should go green (they fail currently), because the
titus-executor expects awsElasticBlockStore to be filled in, but is
currently not.

### DIFF
--- a/titus-server-master/src/test/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactoryTest.java
+++ b/titus-server-master/src/test/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactoryTest.java
@@ -42,7 +42,6 @@ import com.netflix.titus.api.jobmanager.model.job.volume.Volume;
 import com.netflix.titus.api.model.EfsMount;
 import com.netflix.titus.common.util.tuple.Pair;
 import com.netflix.titus.master.kubernetes.pod.KubePodConfiguration;
-import com.netflix.titus.master.kubernetes.pod.KubePodUtil;
 import com.netflix.titus.master.kubernetes.pod.affinity.PodAffinityFactory;
 import com.netflix.titus.master.kubernetes.pod.env.DefaultPodEnvFactory;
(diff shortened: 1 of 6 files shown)


titus-server-master/src/test/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactoryTest.java
```diff
@@ -52,12 +52,7 @@
 import com.netflix.titus.master.service.management.ApplicationSlaManagementService;
 import com.netflix.titus.runtime.kubernetes.KubeConstants;
 import com.netflix.titus.testkit.model.job.JobGenerator;
-import io.kubernetes.client.openapi.models.V1Affinity;
-import io.kubernetes.client.openapi.models.V1Container;
-import io.kubernetes.client.openapi.models.V1FlexVolumeSource;
-import io.kubernetes.client.openapi.models.V1Pod;
-import io.kubernetes.client.openapi.models.V1Volume;
-import io.kubernetes.client.openapi.models.V1VolumeMount;
+import io.kubernetes.client.openapi.models.*;
```
tbak, line 55: We do not use `*`, only direct imports as a convention.

tbak (approved):

titus-server-master/src/test/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactoryTest.java
solarkennedy: Oh sorry, I reloaded my ide and lost the imported code style xml.

andrew-leung (approved):

PR #1210, opened 2022-01-19
Remove the opportunistic resource CRD support

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodConstants.java
```diff
@@ -113,19 +111,6 @@ private KubePodConstants() {
...
-    public static final String OPPORTUNISTIC_RESOURCE_ID = "opportunistic.scheduler.titus.netflix.com/id";
-
-    public static final String PREDICTION_RUNTIME = "predictions.scheduler.titus.netflix.com/runtime";
-    public static final String PREDICTION_CONFIDENCE = "predictions.scheduler.titus.netflix.com/confidence";
-    public static final String PREDICTION_MODEL_ID = "predictions.scheduler.titus.netflix.com/model-id";
-    public static final String PREDICTION_MODEL_VERSION = "predictions.scheduler.titus.netflix.com/version";
-
-    public static final String PREDICTION_AB_TEST_CELL = "predictions.scheduler.titus.netflix.com/ab-test";
-    public static final String PREDICTION_AVAILABLE = "predictions.scheduler.titus.netflix.com/available";
-    public static final String PREDICTION_SELECTOR_INFO = "predictions.scheduler.titus.netflix.com/selector-info";
```
fabiokung, line 127: we should keep job runtime predictions, as they have uses beyond scheduling on opportunistic CPUs

fabiokung (changes requested):
other than removing attributes for runtime predictions, LGTM

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodConstants.java
tbak: At least some of these constants are defined in another class (`com.netflix.titus.runtime.kubernetes.KubeConstants`). There were not used.

fabiokung (approved):

PR #1214, opened 2022-01-20
Use the Fabric8IO Kube client in TaskRelocation

tbak:


amit-git (approved):
May be KubeApiFacade interface should be marked deprecated after this change

PR #1215, opened 2022-01-21
Use the existing containerEnv factory with V1 pods

solarkennedy:
I want v1 pods to use the existing business logic as v0 pods when it
comes to generating env variables (at least, as much as possible).

This change allows V1 pods to get the existing env logic, which includes
setting private NETFLIX env variables.

To do this, I had to extend the API of the buildContainerEnv to export
the pair thing, even if in v0 it doesn't do anything with the left hand
side of the pair.


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/legacy/DefaultAggregatingContainerEnvFactory.java
```diff
@@ -55,18 +57,22 @@ public DefaultAggregatingContainerEnvFactory(TitusRuntime titusRuntime, Containe
...
+            List<String> incomingSystemEnv = incomingContainerEnv.getLeft();
             // Tracking conflicting env var for any two given factories
             env.keySet()
                     .stream()
-                    .filter(envMap::containsKey)
+                    .filter(incomingEnv::containsKey)
                     .forEach(envVarName -> incrementConflictCounter(envVarName, job.getId(), job.getJobDescriptor().getApplicationName()));
-            env.putAll(envMap);
+            env.putAll(incomingEnv);
+            systemEnvNames.addAll(incomingSystemEnv);
```
tbak, line 73: This will not work well with the conflicts, unless the assumption is that we always start with the user env, and system overrides come later.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/legacy/TitusProvidedContainerEnvFactory.java
```diff
@@ -16,32 +16,45 @@
...
     private static final ContainerEnvFactory INSTANCE = new TitusProvidedContainerEnvFactory();
 
     @Override
-    public Map<String, String> buildContainerEnv(Job<?> job, Task task) {
+    public Pair<List<String>, Map<String, String>> buildContainerEnv(Job<?> job, Task task) {
+        // In TitusProvidedContainerEnvFactory, systemEnvNames ends up having everything we add
+        List<String> systemEnvNames = new ArrayList<>();
         Map<String, String> env = new HashMap<>();
         env.put(KubeConstants.POD_ENV_TITUS_JOB_ID, task.getJobId());
+        systemEnvNames.add(KubeConstants.POD_ENV_TITUS_JOB_ID);
```
tbak, line 41: Move these to immutable `List` constant, which is copied only for batch jobs to add the extra entry.

tbak (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/legacy/DefaultAggregatingContainerEnvFactory.java
solarkennedy: It will work in the sense that `systemEnvNames` will have a list of all the keys, *even* if the values are not what we expected.

PR #1219, opened 2022-02-04
Added image data and platform sidecar data to containerStatus

solarkennedy:
Now that we have containerStatus plumbed all the way through, this is
the next pass which adds more data like image stuff and platform sidecar
stuff.

The reason this is difficult is because with platform sidecars, we need
to look at annotations to verify what the exact tag/digest/name that an
image comes from. The stock k8s v1 container status has no room for this
of course. And it doesn't have room for tag/digest, which we really
want.

This adds code for parsing and extracting annotations to be able to get
back the exact container image and (optionally) the platform sidecar
that added it.

I consider this code to be "best effort", as this is a new feature and
there are a number of ways that the data can simply not show up.

Once this data is in our API, I can make it visible to the Titus UI and
spinnaker to make it clear to our users exactly what sidecars are
running, what version, and why (like what platform sidecar mutator added
it).

### DIFF
--- a/titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/TaskDataInjector.java
+++ b/titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/TaskDataInjector.java
@@ -33,6 +33,7 @@ import com.netflix.titus.api.relocation.model.TaskRelocationPlan;
 import com.netflix.titus.common.util.CollectionsExt;
 import com.netflix.titus.common.util.ExceptionExt;
 import com.netflix.titus.common.util.rx.ReactorExt;
+import com.netflix.titus.grpc.protogen.BasicImage;
 import com.netflix.titus.grpc.protogen.JobChangeNotification;
 import com.netflix.titus.grpc.protogen.MigrationDetails;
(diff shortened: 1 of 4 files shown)


titus-server-gateway/src/main/java/com/netflix/titus/gateway/service/v3/internal/TaskDataInjector.java
```diff
@@ -223,26 +228,107 @@ private Task newTaskWithContainerState(Task task) {
...
+        BasicImage.Builder bi = BasicImage.newBuilder();
+        String imageWithoutRegistry = stripRegistryFromImage(image);
+        String name = getNameFromImageString(imageWithoutRegistry);
+        bi.setName(name);
+        bi.setDigest(getDigestFromImageString(imageWithoutRegistry));
+        bi.setTag(getTagFromImageString(imageWithoutRegistry, containerName, pod));
+        return bi.build();
+    }
+
+    private static String stripRegistryFromImage(String image) {
```
tbak, line 274: All these methods might be useful in other places as well. We should have them in `KubeUtil` or something like that.

tbak (approved):

PR #1230, opened 2022-03-11
image sanitizer fix

amit-git:
### Sanitizer behavior change

With this change, we fail early on resolving image digest. For an unknown error, it'd fail job creation instead of letting the job scheduling progress without image digest only to fail during the container launch time.


titus-server-runtime/src/main/java/com/netflix/titus/runtime/endpoint/admission/JobImageSanitizer.java
```diff
@@ -130,7 +130,8 @@ private boolean isAllowedException(Throwable throwable, Image image) {
                 imageResource,
```
gabrielhartmann, line 130: Is this metric getting incremented on the recent registry failures we've seen?

gabrielhartmann (approved):

tbak (approved):

PR #1235, opened 2022-03-22
Upped the DELETE_GRACE_PERIOD_SECONDS to something higher than 600

solarkennedy:
We are deleting our pods a little too early, and there are pods that are
not actually finished.

This confuses the VPC service because the pod is gone, which is a signal
to VPC service that the IP is free to use.

The current grace period seconds is 600, so this is 200 seconds more.


joshi-keyur:
Should we make this a dynamic property?

solarkennedy:
Gunna let @fabiokung take over with another PR

PR #1236, opened 2022-03-22
Configurable deleteGracePeriod for Pods

fabiokung:
Supersedes #1235

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/client/DefaultDirectKubeApiServerIntegrator.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/client/DefaultDirectKubeApiServerIntegrator.java
@@ -68,7 +68,6 @@ public class DefaultDirectKubeApiServerIntegrator implements DirectKubeApiServer
     private static final Logger logger = LoggerFactory.getLogger(DefaultDirectKubeApiServerIntegrator.class);
 
     private static final String KUBERNETES_NAMESPACE = "default";
-    private static final int DELETE_GRACE_PERIOD_SECONDS = 300;
 
     private static final String NOT_FOUND = "Not Found";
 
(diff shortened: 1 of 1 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/client/DefaultDirectKubeApiServerIntegrator.java
```diff
@@ -172,7 +169,7 @@ public boolean isReadyForScheduling() {
             Stopwatch timer = Stopwatch.createStarted();
             try {
                 logger.info("Deleting pod: {}", taskId);
-                kubeApiFacade.deleteNamespacedPod(KUBERNETES_NAMESPACE, taskId, DELETE_GRACE_PERIOD_SECONDS);
+                kubeApiFacade.deleteNamespacedPod(KUBERNETES_NAMESPACE, taskId, configuration.getDeleteGracePeriodSeconds());
```
solarkennedy, line 172: I think @sargun wants this to be a little more than the actual pod DeleteGracePeriodSeconds, to give time for the surrounding systems to react.

But IMO I think having them match (==) is the most correct?

solarkennedy (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/client/DefaultDirectKubeApiServerIntegrator.java
fabiokung: no strong opinion either way, but leaving the pod on apiserver a little longer than `DeleteGracePeriodSeconds` makes sense given there are propagation delays across multiple components beyond the kubelet/executor sleeping on `DeleteGracePeriodSeconds`

tbak (approved):

PR #1240, opened 2022-03-25
refactoring job submit limiter

amit-git:
Currently two error conditions are clubbed together in the job submit limiter that makes it hard to distinguish between 
1.  Duplicate job sequence ID submission
2. Max active jobs limit reached.

This code change is aimed at separating the two error conditions into their own JobManagerException error codes. It enables the API to return `RESOURCE_EXHAUSTED` when max active job limit is reached.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
@@ -245,6 +245,7 @@ public class DefaultV3JobOperations implements V3JobOperations {
                     }
                     Optional<JobManagerException> jobLimitError = jobSubmitLimiter.checkIfAllowed(jobDescriptorWithCallerId);
                     if (jobLimitError.isPresent()) {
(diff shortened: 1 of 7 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/DefaultV3JobOperations.java
```diff
@@ -243,10 +243,9 @@ public void shutdown() {
                     if (reservationFailure.isPresent()) {
                         return Observable.error(JobManagerException.jobCreateLimited(reservationFailure.get()));
                     }
-                    Optional<String> limited = jobSubmitLimiter.checkIfAllowed(jobDescriptorWithCallerId);
-                    if (limited.isPresent()) {
-                        jobSubmitLimiter.releaseId(jobDescriptorWithCallerId);
```
tbak, line 248: Is it safe to remove this line?

amit-git: Good catch. Didn't mean to remove it from here

tbak (approved):

PR #1253, opened 2022-04-07
Respect killWaitSeconds job attribute

solarkennedy:
I believe this got lost in the v0->v1 transition.

Do I need to try/catch number format exceptions? or would it be better
ot just let it raise so the user can see it somehow?


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
```diff
@@ -291,6 +293,14 @@ private V1Container buildV1ExtraContainers(BasicContainer extraContainer) {
                 .volumeMounts(KubePodUtil.buildV1VolumeMounts(extraContainer.getVolumeMounts()));
     }
 
+    private Long getTerminationGracePeriodSeconds(Job<?> job) {
+        String jobKillWaitSeconds = job.getJobDescriptor().getAttributes().get(JOB_PARAMETER_ATTRIBUTES_KILL_WAIT_SECONDS);
+        if (jobKillWaitSeconds != null) {
+            return Long.valueOf(jobKillWaitSeconds);
```
tbak, line 299: Try/catch in case a user sets invalid value. In the latter case, set default.

tbak:
@solarkennedy This error would not be propagated to the client directly as pods are created by the reconciler. If you want to do that, you would have to verify how the exception is mapped to the task state, and possibly do some adjustments.

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/v1/V1SpecPodFactory.java
solarkennedy: added

tbak (approved):

PR #1254, opened 2022-04-08
RFC: Added generated Annotations.java for kube-common constants

solarkennedy:
I'm a little worried about copy/pasting constants between golang and
java.

It would be nice if we really could have these in 1 place
(titus-kube-common in this PR) and then never worry about them getting
out of sync again.

If you agree, then I'll do the additional work of removing the
duplicated consts of KubePodConstants and KubeConstants.

Also if you like this approach I can extend it to include labels and
node stuff (just, one PR at a time).


solarkennedy:
Whoops, I made an RFC but didn't add reviewers to comment! Added.

tbak (approved):

solarkennedy:
I've rebased and added a commit where I've replaced all the network-related kube constants with autogen ones.

solarkennedy:
Ok I've migrated a couple classes of annotations in this PR so you can see how it looks.

As expected, it is just "lots of red" as the constants files are reduced.

I don't want to move 100% of the constants in 1 pr. I would like to do this a little incrementally.

joshi-keyur (approved):

gabrielhartmann (approved):

PR #1256, opened 2022-04-08
Favor bin-packing when terminating tasks for service jobs

joshi-keyur:
### Description of the Change

When a service job scales down, we terminate `Started` tasks on a service job to match the new lower job size.
Tasks for termination are identified in a descending order of number of tasks per agent first and then by number of tasks per zone to zone-balance the terminations when possible. The unit test included in this PR provides a good visualization of this. This approach by design is likely to leave the remaining tasks of the job in a same or more spread state across the fleet. 

The alternate approach presented here (controlled by a feature flag) provides a bin-packing friendly way for task terminations with the same zone-balancing treatment in the former approach. Under this, tasks are identified in the ascending order of task count per agent first. This increases the likelihood that agents with lower number of running tasks can become empty and naturally scale down.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/service/ScaleDownEvaluator.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/service/ScaleDownEvaluator.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import com.netflix.titus.api.FeatureActivationConfiguration;
 import com.netflix.titus.api.jobmanager.TaskAttributes;
 import com.netflix.titus.api.jobmanager.model.job.ServiceJobTask;
 import com.netflix.titus.api.jobmanager.model.job.Task;
@@ -79,7 +80,7 @@ import com.netflix.titus.common.util.tuple.Pair;
  */
 class ScaleDownEvaluator {
 
-    static List<ServiceJobTask> selectTasksToTerminate(List<ServiceJobTask> allTasks, int expectedSize, TitusRuntime titusRuntime) {
+    static List<ServiceJobTask> selectTasksToTerminate(FeatureActivationConfiguration configuration, List<ServiceJobTask> allTasks, int expectedSize, TitusRuntime titusRuntime) {
         int targetTerminateCount = allTasks.size() - expectedSize;
         if (targetTerminateCount <= 0) {
             return Collections.emptyList();
@@ -110,7 +111,12 @@ class ScaleDownEvaluator {
         List<ServiceJobTask> startedTasks = notStartedAndStartedTaskGroups.getRight();
 
         // Step 4: scale down tasks in Started state
-        List<ServiceJobTask> startedToRemove = selectTasksToTerminateInEquivalenceGroup(startedTasks, targetTerminateCount - tasksToKill.size(), titusRuntime);
+        List<ServiceJobTask> startedToRemove;
+        if (configuration.isServiceJobTaskTerminationFavorBinPackingEnabled()) {
+            startedToRemove = selectTasksToTerminateInEquivalenceGroupFavorBinPacking(startedTasks, targetTerminateCount - tasksToKill.size(), titusRuntime);
+        } else {
+            startedToRemove = selectTasksToTerminateInEquivalenceGroup(startedTasks, targetTerminateCount - tasksToKill.size(), titusRuntime);
+        }
         appendCandidatesToTerminate(tasksToKill, startedToRemove, targetTerminateCount);
 
         // Extra check in case we messed up somewhere.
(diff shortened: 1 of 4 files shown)


titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/service/ScaleDownEvaluator.java
```diff
@@ -177,6 +183,36 @@ private static List<ServiceJobTask> selectTasksToTerminateInEquivalenceGroup(Lis
...
+            // smallest number of tasks on an agent in any zone
+            int smallestGroup = region.getSmallestTaskGroup();
+
+            // Step 2: kill tasks from the smallest groups, trying to maintain zone balancing
+            // There is increased likelihood that this way we will give agents with a smaller number of tasks
+            // a chance of being scaled-down.
+            boolean hasMore = true;
+            while (tasksToKill.size() < targetTerminateCount && hasMore) {
+                Optional<Zone> zoneOpt = region.getSmallestZoneWithTaskGroupSize(smallestGroup);
+                if (hasMore = zoneOpt.isPresent()) {
```
solarkennedy, line 201: This this `=` correct? The if statement just checks assignment? or is it checking the .isPresent?

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/service/ScaleDownEvaluator.java
```diff
@@ -177,6 +183,36 @@ private static List<ServiceJobTask> selectTasksToTerminateInEquivalenceGroup(Lis
...
+                    // Step 3: remove oldest task
```
solarkennedy, line 202: This comment is no longer correct, right?

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/service/ScaleDownEvaluator.java
```diff
@@ -177,6 +183,36 @@ private static List<ServiceJobTask> selectTasksToTerminateInEquivalenceGroup(Lis
...
+            // Step 1: select smallest groups of tasks to scale down
+            // smallest number of tasks on an agent in any zone
+            int smallestGroup = region.getSmallestTaskGroup();
+
+            // Step 2: kill tasks from the smallest groups, trying to maintain zone balancing
+            // There is increased likelihood that this way we will give agents with a smaller number of tasks
+            // a chance of being scaled-down.
+            boolean hasMore = true;
+            while (tasksToKill.size() < targetTerminateCount && hasMore) {
+                Optional<Zone> zoneOpt = region.getSmallestZoneWithTaskGroupSize(smallestGroup);
```
solarkennedy, line 200: How does this maintain zone balancing? 
If we really want to maintain zone balancing (more than binbacking) then I think this should reuse `getLargestZoneWithTaskGroupSize`.

joshi-keyur: I think the name of this function needs to be improved. 
Given two zones with multiple agents where both the zones have an agent with the smallest number of tasks/agent, we'd prefer to terminate tasks on the agent in a zone that has more overall tasks so we end up reducing the disparity between zones.

gabrielhartmann: I don't think zone balance should be a responsibility or consideration of this component at all.  Zone balance seems like a responsibility of something in fleet management land.  If, due to binpacking, we are always removing machines from a particular zone, generating zone imbalance, then fleet management should prefer to add machines in the same zone.

gabrielhartmann:
I don't think this change is necessary or desirable for a few reasons.
1. bin-packing hasn't been turned on during scheduling in the first place, so doing it on descheduling seems premature
1. turning on bin-packing plus the behavior of the current defragger will already achieve the same result as this PR
1. this is a heuristic based approach to descheduling

Regarding the last point there is already some work done to determine which machines are most likely to empty the soonest.  This ranking, if insufficient, should be extended to consistently identify machines which machines are closest to retirement.  Then systems both on scheduling and descheduling sides can consult a single source of truth.  Attempts to make scheduling and descheduling sides agree using separate sets of heuristics is undesirable.

gabrielhartmann (commented):
.

gabrielhartmann (changes requested):
I don't think this change is necessary or desirable for a few reasons.

1. bin-packing hasn't been turned on during scheduling in the first place, so doing it on descheduling seems premature
1. turning on bin-packing plus the behavior of the current defragger will already achieve the same result as this PR
1. this is a heuristic based approach to descheduling

Regarding the last point, there is already some work done to determine which machines are most likely to empty the soonest. This ranking, if insufficient, should be extended to consistently identify machines which machines are closest to retirement. Then systems both on scheduling and descheduling sides can consult a single source of truth. Attempts to make scheduling and descheduling sides agree using separate sets of heuristics is undesirable.

titus-server-master/src/main/java/com/netflix/titus/master/jobmanager/service/service/ScaleDownEvaluator.java
gabrielhartmann: Think about it this way.  We can't address every optimization at every layer so we have to make clear decisions about component responsibility.  Why are we tackling zone balancing here?  Why not node health (prefer descheduling on nodes which are frequently unhealthy)?  Why not resource utilization (prefer descheduling on nodes which are excessively oversubscribed)?  Why not deployment plans (prefer descheduling on nodes part of an immutable deploy)?

All these optimizations could help bin-packing or any of a variety of fleetwide goals.  But we can't responsibly embed these heuristics into a single component, and then try to embed them again in a different component.  Scheduling and descheduling sides need to have a shared view of what nodes are to be avoided/preferred.

A single place needs to synthesize all our fleet management policy goals so that scheduling and descheduling sides can cooperate in a debuggable and maintainable fashion.

joshi-keyur:
Thanks @gabrielhartmann for the comments. I am going to close this PR. 
We can revisit the descheduling topic once we get a clearer picture after rolling out mixed-scheduling.

PR #1260, opened 2022-04-19
network mode added to ES model

amit-git:
Network mode configuration is added to the published task document


andrew-leung (approved):

solarkennedy:
Can you say more about why you want this?

Most of the time this will be "unset", and instead we may want the "effective network mode" annotation instead.

Who is asking for this feature and why?

amit-git:
We need to inspect (and query) for containers launched with a specific network mode. I think it makes sense to also include the effective network mode if it's available with Task info.

solarkennedy (approved):

PR #1262, opened 2022-04-22
Handle better NodeLost pod state

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/PodToTaskMapper.java
```diff
@@ -241,11 +246,43 @@ private Either<TaskStatus, String> handlePodRunning() {
...
+            long deadline = lastUpdateTime + configuration.getNodeLostTimeoutMs();
+            if (deadline > now) {
+                // Keep the existing state
+                return Either.ofValue(task.getStatus());
+            }
+            // We waited long enough. Time to mark this task as failed.
+            return Either.ofValue(TaskStatus.newBuilder()
+                    .withState(Finished)
+                    .withReasonCode(REASON_TASK_LOST)
+                    .withReasonMessage("The node where task was scheduled is lost")
```
solarkennedy, line 276: `The node where task was scheduled is lost before it got a chance to start`
(or something that indicates that it was only on the Accepted state. Right?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/PodToTaskMapper.java
```diff
@@ -241,11 +246,43 @@ private Either<TaskStatus, String> handlePodRunning() {
         if (isBefore(Started, taskState)) {
             return unexpected("pod state (Running) not consistent with the task state");
         }
+
+        long now = titusRuntime.getClock().wallTime();
+        if ("NodeLost".equals(podWrapper.getReason())) {
+            // If task is still in the Accepted state, move it first, as we know it was scheduled.
+            if (taskState == TaskState.Accepted) {
+                return Either.ofValue(TaskStatus.newBuilder()
+                        .withState(Launched)
```
solarkennedy, line 255: Wait, `Launched` ?
How bad would it be to have a new state like `Disconnected` ?

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/PodToTaskMapper.java
tbak: Actually, this logic is executed when the pod is in Running state

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/PodToTaskMapper.java
solarkennedy: Hmm. I don't feel good about "lying" that we got Launched when we never heard that from the executor.

tbak: I think the launched state is the best one, as the task is assigned to a scheduler which means it moved out of Accepted state. We do not know if the task moved to StartInitiated or Running phase.

PR #1264, opened 2022-04-28
Read the capacity group data from Kubernetes

tbak:


titus-server-master/src/main/java/com/netflix/titus/master/service/management/kube/KubeApplicationSlaManagementService.java
```diff
@@ -0,0 +1,185 @@
...
+    private final Scheduler scheduler;
+
+    private volatile Map<String, ApplicationSLA> capacityGroups = Collections.emptyMap();
+
+    public KubeApplicationSlaManagementService(NamespacedKubernetesClient kubeApiClient, TitusRuntime titusRuntime) {
+        this.crdClient = kubeApiClient.resources(F8IOCapacityGroup.class).inNamespace("default");
+        this.capacityGroupInformer = new CapacityGroupInformerWrapper(kubeApiClient, titusRuntime);
+        this.scheduler = Schedulers.boundedElastic();
+
+        capacityGroupInformer.getInformer().addEventHandler(new ResourceEventHandler<F8IOCapacityGroup>() {
```
gabrielhartmann, line 68: Have you looked at this?
https://javadoc.io/static/io.kubernetes/client-java/8.0.0/io/kubernetes/client/informer/cache/Cache.html

You shouldn't have to write your own cache.

gabrielhartmann (approved):

titus-server-master/src/main/java/com/netflix/titus/master/service/management/kube/KubeApplicationSlaManagementService.java
tbak: I do not think it supports value transformers? We cache `ApplicationSla` objects to avoid constant mapping from CRDs.

PR #1266, opened 2022-04-29
Added scheduler priority and scheduler name when in mixed mode

solarkennedy:
Porting the webhook logic, this adds a new priority class for reserved/elastic workloads when in mixed mode, as well as setting the actual scheduler name.

I'm pretty sure the scheduler name may be TBD? (@joshi-keyur have you picked it out yet? How about `titus-kube-scheduler-mixed` ?)

----

For titus-kube-common constants, I currently only auto-generate annotations.
I'll make a subsequent PR to do labels and other misc constants like these scheduler names too.

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
@@ -280,6 +280,9 @@ public class KubePodUtil {
     }
 
     public static String selectScheduler(SchedulerConfiguration schedulerConfiguration, ApplicationSLA capacityGroupDescriptor, KubePodConfiguration configuration) {
+        if (configuration.isMixedSchedulingEnabled()) {
+            return configuration.getMixedSchedulingSchedulerName();
+        }
         String schedulerName;
         if (capacityGroupDescriptor != null && capacityGroupDescriptor.getTier() == Tier.Critical) {
             if (schedulerConfiguration.isCriticalServiceJobSpreadingEnabled()) {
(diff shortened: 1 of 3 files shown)


okigan:
titus-kube-scheduler-mixed

Sounds good to me.

solarkennedy:
> titus-kube-scheduler-mixed Sounds good to me.
> […](#)
> On Apr 29, 2022, at 1:10 PM, Kyle Anderson ***@***.***> wrote: titus-kube-scheduler-mixed

Looks like that is what it was from the webook too: https://github.com/Netflix/titus-control-plane/pull/1266#issuecomment-1113684240

tbak (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
```diff
@@ -292,4 +295,17 @@ public static String selectScheduler(SchedulerConfiguration schedulerConfigurati
         }
         return schedulerName;
     }
+
+    public static String selectPriorityClassName(ApplicationSLA capacityGroupDescriptor, KubePodConfiguration configuration) {
```
joshi-keyur, line 299: Why do we need this?

solarkennedy: @rostyboost said we did.
Also seems like a good idea: if we are using one kubescheduler, we don't want the possibility of a ton of "elastic" jobs skipping the line.

Right now this never happens because we have 2 different schedulers. The reserved scheduler never even sees elastic jobs. But in this new world, we want to make sure elastic jobs can flood and overwhelm the scheduler and prevent important jobs from being scheduled first.

joshi-keyur: Makes sense. 
In the previous implementation of mixed scheduling we were routing Flex tier jobs to `reserved` which made this prioritization necessary. In this iteration, we treat any Critical tier jobs that are over their reservations also as `elastic`. Maybe that finer distinction is not yet required.

joshi-keyur (approved):

PR #1269, opened 2022-05-03
Fetch a pod name from the pod metadata not the container name

tbak:


fabiokung (approved):

titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/client/DefaultDirectKubeApiServerIntegrator.java
```diff
@@ -256,51 +254,53 @@ public void onAdd(V1Pod pod) {
...
                     Stopwatch stopwatch = Stopwatch.createStarted();
+                    String taskId = newPod.getMetadata().getName();
                     try {
                         if (!KubeUtil.isOwnedByKubeScheduler(newPod)) {
                             return;
                         }
 
                         metrics.onUpdate(newPod);
 
-                        pods.put(newPod.getSpec().getContainers().get(0).getName(), newPod);
```
solarkennedy, line 273: Great!

solarkennedy (commented):
Thanks!

Looks like this is the last part of the codebase that does `getContainers().get(0)` and makes assumptions about the name.

PR #1272, opened 2022-05-05
RFC: Allow service jobs higher scheduling priority compared to batch

solarkennedy:
We are slowly dipping our toes into a new world of scheduling with the
introduction of mixed scheduling.

Soon we'll have 1 unified scheduler handling things.

In that world, I do not want users _1_ Red/Black deploy job having to wait
behind 1000 batch jobs that got queued up from a CMB run.

Therefore, I would like to introduce a new priorityclass for service
jobs to run in, that is not as high as those that have reserved
capacity, but higher than batch.

**Someday** we'll continue to expand this function to allow for more
cases, like high priority batch jobs, or preemptable stuff, etc. But for
now I would like to ease us towards the mixed scheduling world.
But I don't want mixed scheduling to be *worse* than the world we have
today, and I think a medium tier will fix that.

Google Doc: https://docs.google.com/document/d/1F0N6Fljuuxy4n2qQ-9boXtSY5YrmSU_yzNcn-hmd3-w/edit#

### DIFF
--- a/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
+++ b/titus-server-master/src/main/java/com/netflix/titus/master/kubernetes/pod/KubePodUtil.java
@@ -29,12 +29,14 @@ import com.netflix.titus.api.jobmanager.JobAttributes;
 import com.netflix.titus.api.jobmanager.TaskAttributes;
 import com.netflix.titus.api.jobmanager.model.job.Image;
 import com.netflix.titus.api.jobmanager.model.job.Job;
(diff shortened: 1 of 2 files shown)


fabiokung (approved):

gabrielhartmann (approved):

rostyboost:
LGTM. Make sure the `priorityClass` object exists before deploying.

rostyboost (approved):

tbak (approved):

joshi-keyur (approved):

PR #1279, opened 2022-05-13
Use Titus job attributies to configure PlatformSidecars

dilin-nflx:
### Description of the Change

- Added feature to configure Platform Sidecars by using container attributes. Platform Sidecar statement has priority over container attributes and will override them.

### DIFF
--- a/titus-server-master/src/test/java/com/netflix/titus/master/kubernetes/pod/KubePodUtilTest.java
+++ b/titus-server-master/src/test/java/com/netflix/titus/master/kubernetes/pod/KubePodUtilTest.java
@@ -24,9 +24,19 @@ import com.netflix.titus.api.jobmanager.model.job.ext.BatchJobExt;
 import com.netflix.titus.common.util.CollectionsExt;
 import com.netflix.titus.runtime.kubernetes.KubeConstants;
 import com.netflix.titus.testkit.model.job.JobGenerator;
+import com.netflix.titus.api.jobmanager.model.job.ext.ServiceJobExt;
+import com.netflix.titus.testkit.model.job.JobDescriptorGenerator;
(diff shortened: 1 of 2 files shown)


titus-server-master/src/test/java/com/netflix/titus/master/kubernetes/pod/KubePodUtilTest.java
```diff
@@ -55,4 +65,32 @@ public void testSanitizeVolumeName() {
         String name = "Ab9bac3e:6ea1:4bc3:a803:e0070ca434c3/";
         assertThat(KubePodUtil.sanitizeVolumeName(name)).matches("ab9bac3e-6ea1-4bc3-a803-e0070ca434c3--vol");
     }
+
+    @Test
+    public void testCreatePlatformSidecarAnnotations() {
+        List<PlatformSidecar> platformSidecars = Arrays.asList(
+                PlatformSidecar.newBuilder().withName("test_sc_v1").withChannel("dev").withArguments("{\"foo\":\"fromPS\"}").build()
```
solarkennedy, line 72: I'm confused at this test.

Can you spilt this into two tests, one called `testCreatePlatformSidecarAnnotationsPrefersAPIOverAttributes` and `testCreatePlatformSidecarAnnotationsAcceptsAttributes` ?

I think that would make it more clear what you are trying to assert

solarkennedy (commented):
I'm still conflicted on this approach.

Having a second way to configure platform sidecars like this seems like a way to build up tech debt, without having (forcing) path forward for users to migrate away from it.

Why is platform sidecars any different than any feature we have on titus that uses the API to configure it?

solarkennedy:
We are archiving this repo.
Sorry you caught us *just* as we were transitioning :(

